PR #1, opened 2016-11-30
Correlate relay logs simple

shlomi-noach:
This PR enables relay-log correlation based on relay-log coordinates only, and without need for pseudo-gtid.

correlating means: given _these_ relaylog coordinates in _this_ instance, what are the correlated relaylog coordinates in this _other_ instance?

The solution proposed in this PR only works when the two instances are siblings, ie. have the same master.

It works without GTID nor Pseudo-GTID, and does not require agents. It works as follows:

the output of `show relaylog events in ...` annoyingly results with `Pos` and `End_log_pos` per event, and the annoying part is that `End_log_pos` does not indicate the end of position in the _relay_ log but rather that of the master's _binary_ log.

It is very annoying, because given an event in the relay log, you don't know what's the position of the next event, and need to actually iterate events to get there.

Thankfully, this info helps us in syncing the binlogs. While relay logs of two replicas are uncoordinated, and names and positions may vary, the `End_log_pos` entry is always identical, because it always indicates the related position on the master, and we already clarified both instances are siblings, thus have same master.

So given `End_log_pos` in entry of one relaylog, I'm able to search for an entry with that same `End_log_pos` in another instance's relay logs.

Potentially this is still not unique enough: maybe I found a `End_log_pos` of same value, but that points to a different binary log? While unlikely this can happen. As further validation, I'm then able to compare the content of the entry: the two entries have just been copied from the master. They haven't been processed yet. They're identical. They have to be identical.

Mathematically this is still not unique enough. We _could_ hit a case where:

- our master would crash (say at binlog `mysql-bin.00025`)
- most up-to-date replica went as far as `mysql-bin.00025:12345` 
- another replica, the one we want to promote as new master, would point to a previous binary log: `mysql-bin.00024:12345` that has exact same content.

Right now I see this as highly improbable, to the point of "come on". It would mean our candidate replica is behind by a full binary log -- we'd likely not want to promote it in the first place.

Anyway there's more checks that can be made, in the future, but right now I'm leaving this as is.

This PR has nothing to do with recoveries, BTW. It's just the relay-log correlation logic.


shlomi-noach:
Merging for now. More iterations later.

PR #2, opened 2016-11-30
EXPERIMENTAL: rqlite as backend db

shlomi-noach:
Work in progress: experimenting with alternatives to MySQL as backend.

Storyline: https://github.com/github/orchestrator/issues/153

Main reason is to have an embedded database within the `orchestrator` app.


shlomi-noach:
Gah. So embedding `rqlite`/`sqlit` works well. However building the binary takes ages because the `vendor/github.com/mattn/go-sqlite3/sqlite3-binding.c` file needs to be compiled, time and again. It adds some `> 20` seconds on my box, which is unacceptable for every single run.

This isn't a distribution problem, of course, just a development annoyance.

`go install` is naturally working, but that throws the `go-sqlite3` package under `GOPATH`, and I want it to be in this repo's path. That is, I want this repo to always be self contained, such that anyone can `clone` it, and `go run ...` it.

I tried playing with `go1.7` binary-only-packages but this just doesn't work: I removed all source files created a dummy file `vendor/github.com/mattn/go-sqlite3/go-sqlite3.go` containing:
```
//go:binary-only-package

package go-sqlite3
```

but when I `go build` or `go run` I get :
```
vendor/github.com/mattn/go-sqlite3/go-sqlite3.go:3:9: expected 'IDENT', found 'go'
```

I was also hoping using `-pkgdir` into a `pkg` directory I created in the repo path would help. But any time the sources are around, they get recompiled, regardless if there's a binary package under `pkg/` or not.

---
To redefine what I eventually want to get:

I want to have the `go-sqlite3.a` binary package somewhere in my repo path, such that when I `go build` or `go run` the `c` file doesn't get recompiled. And there must be no external dependencies: the repo must be self-contained.

If anyone has a hint on how to achieve that I'm grateful.

shlomi-noach:
Not going to work with `rqlite`. Have POC with `sqlite`.

PR #3, opened 2016-12-01
Rqlite bin

shlomi-noach:
attempts with rqlite binary package, embedded.


shlomi-noach:
Not going to work with `rqlite`. Have POC with `sqlite`.

PR #15, opened 2016-12-13
Recognizing flavor, version_comment

shlomi-noach:
In this PR we read and analyze the flavor of the database: Oracle/Percona/MariaDB/MaxScale based on `version_comment` and other params.

This gets visualized in the web UI, so it's easy to distinguish between different flavors in a cluster.


shlomi-noach:
This deploys well. Ideally I would have `bootstrap`-like icons for MySQL, Percona, MariaDB so that we can indicate the flavor in smaller space than now.

PR #19, opened 2016-12-16
Add missing use case for GetNextBinlogCoordinatesToMatch when moving …

sjmudd:

See https://github.com/outbrain/orchestrator/pull/286.

The commit text incorrectly mentions only the last commit but the patch includes the full changes necessary to move a 5.7 replica around directly or indirectly under a 5.6 master when pseudo-GTID is in use and the 5.7 replica is replicating from a 5.7 intermediate master.

### Description

Issue caused by the 5.7 intermediate master generating ANONYMOUS GTID events which confuse the existing matching process.


sjmudd:
Let me try and explain the 2 use cases that this PR is trying to fix:

(a) moving a 5.7 slave (without binlogs) which is under a 5.7 intermediate master which itself is under a 5.6 master under that master or any other 5.6 intermediate master in the same topology.

Here the 5.6 binlog stream is "normal", but the binlog stream received by the 5.7 slave (and stored in the relay logs) has additional ANONYMOUS gtid events in them.  These events need to be filtered out on the 5.7 slave's relay logs.

(b) moving a 5.7 slave (which is under a 5.6 master or intermediate master) under a 5.7 intermediate master which has binlogs.

In this case the 5.7 slave has no "anonymous GTID events" as it replicates from the 5.6 master but the 5.7 intermediate master does have these events and they must be ignored in order to compare events.

In both cases "hiding" the events, as I had done previously does not work as using the binlog filter means that in the situation where replication is stoppped on the IM or the slave the position at which the stop happens _may_ be on an anonymous GTID event. If that's the case the event will be skipped and the current logic will not think that it's matched things correctly thus giving an error.

I hope that this longer description makes sense.

sjmudd:
Can I do anything more to help pull in this change?

shlomi-noach:
I'm picking up on this. Meanwhile I'm going to make a release without this.

go/inst/instance_binlog_dao.go
```diff
@@ -619,68 +619,158 @@ func getNextBinlogEventsChunk(instance *Instance, startingCoordinates BinlogCoor
 	return events, err
 }
 
-// GetNextBinlogCoordinatesToMatch is given a twin-coordinates couple for a would-be replica (instanceKey) and another
-// instance (otherKey).
+// rpad formats the binlog coordinates to a given size. If the size
+// increases this value is modified so it can be reused later. This
+// is to ensure consistent formatting in debug output.
+func rpad(coordinates BinlogCoordinates, length *int) string {
```
shlomi-noach: Can you please move this function to `binlog.go` as opposed to `binlog_dao.go`?

go/inst/instance_binlog_dao.go
```diff
@@ -619,68 +619,158 @@ func getNextBinlogEventsChunk(instance *Instance, startingCoordinates BinlogCoor
...
+		other.FlavorNameAndMajorVersion() == "MySQL-5.6" // replicating under 5.6 master
+
+	// The logic for other is a bit weird and may require us
+	// to check the instance's master.  To avoid this do some
+	// preliminary checks first to avoid the "master" access
+	// unless absolutely needed.
+
+	if instance.LogBinEnabled || // instance writes binlogs (not relay logs)
+		instance.FlavorNameAndMajorVersion() != "MySQL-5.7" || // instance NOT 5.7 replica
+		other.FlavorNameAndMajorVersion() != "MySQL-5.7" { // new master is NOT  5.7
```
shlomi-noach, line 646: I'm curious, why `MySQL-5.6` and `MySQL-5.7` as opposed to `5.6` and `5.7`? Does the problem not happen with Percona `5.6` & `5.7`?

go/inst/instance_binlog_dao.go
```diff
@@ -700,63 +790,67 @@ func GetNextBinlogCoordinatesToMatch(instance *Instance, instanceCoordinates Bin
...
 			}
 
-			instanceEventInfo = event.Info
-			eventCoordinates = event.Coordinates
-			coordinatesStr := fmt.Sprintf("%+v", event.Coordinates)
-			if len(coordinatesStr) > beautifyCoordinatesLength {
-				beautifyCoordinatesLength = len(coordinatesStr)
-			}
-			log.Debugf("> %+v %+v; %+v", rpad(coordinatesStr, beautifyCoordinatesLength), event.EventType, strings.Split(strings.TrimSpace(instanceEventInfo), "\n")[0])
+			instanceEvent = *event // make a physical copy
```
shlomi-noach, line 796: curious. Why a physical copy?

shlomi-noach (changes requested):

go/inst/instance_binlog_dao.go
sjmudd: I do not use Percona so can not add comments for something I'm not sure about.  (I guess it should work though.) Also when I added this code there was no specific percona flavours to check for so the question was not relevant at the time.

go/inst/instance_binlog_dao.go
sjmudd: That should be done now. Still having a bit of trouble figuring out the workflow here. thanks for your patience.

go/inst/instance_binlog_dao.go
sjmudd: You're comparing two events. If you write to one of them and have 2 pointers pointing to the same event both will be the same and that's not what is wanted. Hence taking a physical copy ensures that you compare 2 different events. That was my intention anyway.

shlomi-noach (approved):

shlomi-noach:
thank you!

PR #22, opened 2016-12-28
Show the count and not the list of server if the number being polled …

sjmudd:
…is over MaxOutdatedKeysToShow (default: 10)

### Issue

This change is to prevent the logging of all the outdated keys (better term would be _polled hosts_ or _polled instances_) when the number grows over a certain threshold. When a large number of servers are being polled this list just gets to big and intrusive.

I have been using a variation of this code on my own patched code for some time. It was not configurable in my case. This allows for the current behaviour by default for small environments, and changes to use my preferred output if the number of monitored servers goes over the configured limit.

### Description

Add a new config setting: MaxOutdatedKeysToShow, default 10, which if exceeded will adjust the logging to just show: `polling NNN outdated keys` rather than `outdated keys: [ host1:12345, host2:12345, ... hostn:12345 ]`

### DIFF
--- a/go/config/config.go
+++ b/go/config/config.go
@@ -345,7 +345,7 @@ func newConfiguration() *Configuration {
 		GraphiteConvertHostnameDotsToUnderscores:     true,
 		GraphitePollSeconds:                          60,
 		URLPrefix:                                    "",
-		MaxOutdatedKeysToShow:                        10,
+		MaxOutdatedKeysToShow:                        64,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


sjmudd:
Note: if you prefer a boolean setting to just show the count rather than the instance list I can change this patch to do that.

e.g. `ShowOutdatedKeysCount` (default `false`, shows all keys, as now, with `true` it would just show the count)

The disadvantage of this approach is that I have to for a non-default configuration setting on my systems and I would prefer to avoid that. I'd prefer orchestrator to adapt to the size without requiring a configuration change.

go/config/config.go
```diff
@@ -344,6 +345,7 @@ func newConfiguration() *Configuration {
 		GraphiteConvertHostnameDotsToUnderscores:     true,
 		GraphitePollSeconds:                          60,
 		URLPrefix:                                    "",
+		MaxOutdatedKeysToShow:                        10,
```
shlomi-noach: Please make the default larger (say `64`), and set your own personal config to `10`

shlomi-noach (changes requested):

go/config/config.go
sjmudd: ok. will change value as suggested.

shlomi-noach (approved):

PR #32, opened 2017-01-02
correlating relay logs

shlomi-noach:
Following up on https://github.com/github/orchestrator/pull/1, in the objective of aligning replicas at failover.

At this time this will cooperate with `orchestrator-agent` at https://github.com/github/orchestrator-agent/pull/13, though the same can be accomplished via remote SSH.

When master fails, `orchestrator` is able to use GTID/Pseudo-GTID to match replicas. However there are a few constraints: what if the most up-to-date replica doesn't have binlogging or `log-slave-updates`? What if it uses a `ROW` based replication where all others use `STATEMENT` based? What if it's of a higher MySQL version? `orchestrator` would have to lose it, even though it contained more data than others.

This PR follows on the MHA approach, which requires either remote agents on MySQL boxes, or remote SSH. The intention is to correlate relay logs between failed replicas (done, optimized speed), then copy & apply such logs from the most up-to-date replica onto a (single) candidate replica.

Why single? Because the candidate replica would have `log-slave-updates` and `orchestrator` would be able to point all other replicas under that one. It is yet to be seen whether comparing and copying relay logs onto a single replica, then applying Pseudo-GTID/GTID logic to heal the rest of the replicas, is faster or slower than comparing and copying relay logs from the most up-to-date replica onto all other replicas.

Initial commits in this PR provide heuristic search for relay log coordinates & entries, which turn relay-log correlation into a subsecond operation, within `1` minute from failure.

### DIFF
--- a/go/inst/instance_topology.go
+++ b/go/inst/instance_topology.go
@@ -1325,32 +1325,33 @@ func CorrelateBinlogCoordinates(instance *Instance, binlogCoordinates *BinlogCoo
 	return nextBinlogCoordinatesToMatch, countMatchedEvents, nil
 }
 
-func CorrelateRelaylogCoordinates(instance *Instance, relaylogCoordinates *BinlogCoordinates, otherInstance *Instance) (correlatedCoordinates, nextCoordinates *BinlogCoordinates, found bool, err error) {
+func CorrelateRelaylogCoordinates(instance *Instance, relaylogCoordinates *BinlogCoordinates, otherInstance *Instance) (instanceCoordinates, correlatedCoordinates, nextCoordinates *BinlogCoordinates, found bool, err error) {
 	// The two servers are expected to have the same master, or this doesn't work
 	if !instance.MasterKey.Equals(&otherInstance.MasterKey) {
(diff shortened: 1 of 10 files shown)


go/inst/instance_topology.go
```diff
@@ -1261,7 +1261,7 @@ func FindLastPseudoGTIDEntry(instance *Instance, recordedInstanceRelayLogCoordin
 	}
 
 	minBinlogCoordinates, minRelaylogCoordinates, err := GetHeuristiclyRecentCoordinatesForInstance(&instance.Key)
-	if instance.LogBinEnabled && instance.LogSlaveUpdatesEnabled && (expectedBinlogFormat == nil || instance.Binlog_format == *expectedBinlogFormat) {
+	if instance.LogBinEnabled && instance.LogSlaveUpdatesEnabled && !*config.RuntimeCLIFlags.SkipBinlogSearch && (expectedBinlogFormat == nil || instance.Binlog_format == *expectedBinlogFormat) {
```
shlomi-noach, line 1263: This change is actually unrelated to this PR. However the need for it came at a good time and it makes sense to include it here.

shlomi-noach:
originally this PR was meant do go the whole way to relaylog-sync on failover. 
However the functionality at this time is substantial enough, and can (and should) be tested. Once tested to work, I suggest merging this, and opening a new PR to take it from here.

shlomi-noach:
merging for now, continued work in other PRs

PR #40, opened 2017-01-05
Workaround for issue#39: set IsCandidate based on promotionRule

sjmudd:
### PromotionRules do not seem to be used as the documentation seems to imply.

See: https://github.com/github/orchestrator/issues/39

### Workaround

I have added a workaround which changes `IsCandidate` based on the `promotionRule` settings:

promotionRule | IsCandidate set to | Comments
---------------|----------------|----------
`Must` | `true` |
`Prefer` | `true` | treated the same as `Must`
`Neutral` | `false` |
`PreferNot` | `false` | treated the same as `Neutral`
`MustNot` | `false` | can't enforce this

### Caveats

Only checked to work with api calls to register candidate. I have not checked the cli code path does the right thing or if it needs modifying.

### Comments on code

- [x] already running in producction


go/inst/instance_dao.go
```diff
@@ -2163,6 +2163,22 @@ func ReadHistoryClusterInstances(clusterName string, historyTimestampPattern str
...
+	//
+	// read instance information from the orchestrator backend
+	if instance, found, err := ReadInstance(instanceKey); err == nil && found {
+		// determine what the modifed rule should be
+		newIsCandidate := (promotionRule == MustPromoteRule) || (promotionRule == PreferPromoteRule)
+
+		if instance.IsCandidate != newIsCandidate {
+			log.Debugf("RegisterCandidateInstance: %+v: Changing IsCandidate from %v to %v", instance.Key, instance.IsCandidate, newIsCandidate)
+			instance.IsCandidate = newIsCandidate
+			writeInstance(instance, false, nil)
```
shlomi-noach, line 2177: The `IsCandidate` field does not get written to the `database_instance` table. The introduced `writeInstance()` call doesn't do anything useful.

IsCandidate is inferred on https://github.com/github/orchestrator/blob/8811c420db892c49b1d0ad34d0cc443344861194/go/inst/instance_dao.go#L810-L811

I think this PR is unfortunately wrong in what it attempts to do.

shlomi-noach (changes requested):
Let's discuss further. I think I will close this PR.

sjmudd:
ok. Looks like I misunderstood the behaviour of the existing code. Let's close this then. Thanks for clarifying.

PR #42, opened 2017-01-08
Failover correlate relay logs ssh

shlomi-noach:
continued work from https://github.com/github/orchestrator/pull/32
this PR will support relaylog synching via remote SSH rather than `orchestrator-agent`.

I'm unsure at this time which of the two is the "better" way. `orchestrator-agent` requires further setup: installing and running an agent on each box. Remote SSH also has its own setup and security issues. Anyway at this time I'd like to see both running, and I'll follow up with tests.

### DIFF
--- a/go/agent/instance_topology_agent.go
+++ b/go/agent/instance_topology_agent.go
@@ -1,5 +1,5 @@
 /*
-   Copyright 2014 Outbrain Inc.
+   Copyright 2017 GitHub Inc.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


shlomi-noach:
Internal tests show this to be good and correct. Will proceed with internal automated testing setup to further validate logic.

shlomi-noach:
I'm merging this for now, and will continue to iterate in future PRs.
The names of the command line commands are subject to change (this code is not released yet and not documented yet).

shlomi-noach:
To clarify, the code in this PR is not called by the failover mechanism. That's exactly what the next iteration would do.

PR #44, opened 2017-01-09
Fix for issue#41 - use readInstancesByCondition() rather than QueryOrchestrator()

sjmudd:
See: https://github.com/github/orchestrator/issues/41

This ensures that the right code is called to pull out information of an instance.


go/inst/instance_dao.go
```diff
@@ -668,34 +668,44 @@ func ReadInstanceClusterAttributes(instance *Instance) (err error) {
...
+		readInstanceCounter.Inc(int64(len(instances)))
 
-	if err != nil {
-		return nil, err
+		for i := range instances {
+			instanceKeys = append(instanceKeys,
+				&InstanceKey{
+					Hostname: instances[i].Key.Hostname,
+					Port:     instances[i].Key.Port,
+				})
```
shlomi-noach: Can just `append(instanceKeys, &(instances[i].Key))`

shlomi-noach (changes requested):
minor change requested, otherwise good

shlomi-noach (approved):

PR #47, opened 2017-01-11
Monitor version of vendor/github.com/rcrowley/go-metrics by adding to vendor.json

sjmudd:
vendor.json includes information on the current version of the imported repo. This is good as we can see if it changes and what commit is in the local repo.

All I did was `govendor fetch github.com/rcrowley/go-metrics` and update the repo with the changes this produced.

* This change updates go-metrics and includes the latest version into orchestrator.
* If you don't use `govendor` nothing changes but later we can at least see if code changes and which changes there are from the revision that's being used.  That's not possible easily at the moment.
* I've checked it works with `./build.sh -b` but not actually checked the code doesn't change behaviour.

Probably all vendor libraries should be treated the same for a similar reason. Perhaps worth considering if you have time?


shlomi-noach (changes requested):
Please strip all `git` info from embedded libraries. Please remove `.gitignore` and otherwise irrelevant files, if any, from this PR

sjmudd:
Hi Shlomi.

I'm not sure that's the right thing to do.  The vendor directory is supposed to have pristine versions of the upstream external code.  If it's different in any way then your assumption that you're using clean external code breaks. The .gitignore and other files that get pulled in are part of the git commit that's mentioned in the vendor.json file and excluding them means that any later attempt to check for updates will trigger them coming back again.

So please reconsider accepting the PR as is for the reasons stated.

shlomi-noach:
`git` information in subdirectories has another unfortunate effect, which is to make `go get github.com/github/orchestrator` not work.

To be honest, I don't think either solution is good. Let me think about it some more.

sjmudd:
Let's forget this for now. Needs looking at but not worth worrying about too much right now.

Issue #46, opened 2017-01-10
Who is the new intermediate master?


theTibi:
Hi, 

If I have a topology like this (just an example):

```
               -> rep3 
rep1 --> rep2-|
               -> rep4

```
Rep2 is an intermediate master. If rep2 dies Orchestrator processes a `DeadIntermediateMaster` failover and reorganises the topology like (just an example):

`rep1 --> rep4 --> rep3 `

So rep4 is going to be an intermediate master now. But based on the `PostFailoverProcesses` placeholders I can not decide who is the new intermediate master.

It has the following placeholders:
{failureType}, {failureDescription}, {failedHost}, {failureCluster}, {failureClusterAlias}, {failureClusterDomain}, {failedPort}, {successorHost}, {successorPort}, {successorAlias}, {countSlaves}, {slaveHosts}, {isDowntimed}, {isSuccessful}, {lostSlaves}

I am trying to call an external script when an intermediate master dies but the script should/has to know who is the new intermediate master after failover.

Is there any solution/ideas for this? 

Thanks.

shlomi-noach:
The new intermediate master is `{successorHost}`, and more specifically it's `{successorHost}:{successorPort}`

BTW you can also get this value from the environment variable `ORC_ORCHESTRATOR_HOST` sent to your script. If you're using shell script, just read `$ORC_ORCHESTRATOR_HOST`.

theTibi:
@shlomi-noach : That was my first thought also. So in this example the `{successorHost}` should be `rep4` if I understand correctly, because `rep4` was promoted.

But in my test the `{successorHost}` was `rep1`, here are some log entries:

```
...
2017-01-10 16:37:27 INFO ChangeMasterTo: Changed master on rep4:3306 to: rep1:3306, mysql-bin.000001:62541370. GTID: false
2017-01-10 16:37:27 INFO Started slave on rep4:3306
...
2017-01-10 16:37:28 DEBUG execCmd: echo 'Recovered from DeadIntermediateMaster on rep1:3306. Failed: rep2:3306; Successor: rep1:3306' >> /tmp/recovery.log
...

```
So it promoted `rep4` but the `{successorHost}` is `rep1`.

Do I miss something here?

shlomi-noach:
Oh OK. So here's the complication: 

(though, first of all, in your case it _should_ have been `rep4`, not arguing)

[...]

PR #61, opened 2017-01-24
considering local instance as successor on RecoverDeadIntermediateMaster

shlomi-noach:
Storyline: https://github.com/github/orchestrator/issues/46#issuecomment-274732657

This PR improves on `{successorHost}` identity in the following case:

- a `DeadIntermediateMaster` recovery is executed. The server `IM` is dead.
- there is no alternate intermediate master to salvage the orphaned replicas
- one of the orphaned replicas, R, is chosen to takeover its siblings
- it manages to take over all of the siblings
- and then successfully connects to `IM`'s master (call it `M`)

In this particular case, the `{successorHost}` is considered to be R. This makes sense because it took charge of everything. Even though `M` adopted `R`, we still consider `R` the successor.

To clarify, `R` is **not** considered as a `{successorHost}` when it's unable to take over **all** of its siblings (which may be pointed onto an altogether intermediate master or even to the master itself).

cc @theTibi


shlomi-noach:
even though this doesn't satisfy #46 this is good to go.

PR #65, opened 2017-02-03
Update Dockerfile

tcaxias:
### Issue
https://github.com/github/orchestrator/issues/66

### Description

This PR reduces the size of the docker image from 3xxMB to 40MB.
Started from importing from a smaller image, plain alpine, then installed go and build-base to compile the project.
All the rest is almost the same as the original, just opted to store GOPATH under /tmp and also delete it at the end of the build.

I hope it helps. Keep up the great work!
Orchestrator is a hell of a tool!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,25 @@
-FROM golang:1.6-alpine
+FROM alpine
 
 EXPOSE  3000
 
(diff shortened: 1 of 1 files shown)


shlomi-noach:
Thank you!

PR #68, opened 2017-02-08
Made pseudo gtid event aware of server read_only

calind:
This PR makes the pseudo-gtid insertion event aware of the server's `read_only` flag. This way, the event can be loaded on co-masters without corrupting the GTID. 

With the previous version you need to manually make sure that you insert GTID binlog event only on master by running an external process. The event scheduler is disabled on replicas but in a co-master situation both servers are masters and execute events.

### DIFF
--- a/docs/pseudo-gtid.md
+++ b/docs/pseudo-gtid.md
@@ -50,9 +50,13 @@ create event if not exists
   on completion preserve
   enable
   do
-    begin
+    main: begin
       DECLARE lock_result INT;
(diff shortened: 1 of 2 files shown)


shlomi-noach (approved):
This looks good.

FWIW, at GitHub we're injecting pseudo-GTID via an external service, which connects as a "normal" user, so that it can't possibly write to a `read_only` server.

calind:
Please don't merge this yet. I think I found some problems with this approach regarding the `ORIGINATOR` filed of the event. Will update the PR.

shlomi-noach:
Thank you!

PR #69, opened 2017-02-08
Take into account replication when creating pseudo gtid event

calind:
By default, `create_pseudo_gtid_event` is replicated to the binlog, which sets the `ORIGINATOR` field to `server_id`. The event is then executed only on the server with that `server_id`. This PR disables the event creation replication and changes the behaviour  for `create_pseudo_gtid_event` in the following sense:

* in case of co-masters: the event must be loaded on both servers and the event is created and runs on co-master with the `read_only` flag disabled
* in case of single master/intermediate master: the event creation is not replicated in any way and only runs if the server has `read_only` disabled, thus giving much control on where the event is triggered

I also re-indented the code since it used a mix of 2 spaces, 4 spaces and tabs.


shlomi-noach (changes requested):
With the massive (welcomed) indentation fixes, I'm unable to clearly see the actual code changes.

However, by reading the PR's description I'm afraid the logic is incorrect.

> the event must be loaded on both servers

on co-master (active-active), injecting pseudo-GTID on both hosts is harmful and leads to p-gtid not being utilized. p-gtid expects total ordering. With active/active m1-m1 there is no total ordering. Replicas of m1 and replicas of m2 get a different sequence of events. p-gtid entries in their binary logs will differ in order, and `orchestrator` would not be able to make sense of it.
There must be a single source for p-gtid.

> the event creation is not replicated in any way

I don't follow. I also disagree that there should be a scenario where a statement is issued on the master that is note replicated (namely the `CREATE EVENT` statement.

Events created on a master are automatically set as `SLAVESIDE_DISABLED` on a replica. If that replica should ever be promoted to master, the event is still disabled. It's on you to `ALTER EVENT...ENABLE` (via hooks).
I'm afraid I do not see the problem you're describing re: "giving much control on where the event is triggered".

calind:
You are right. I got carried away by the fact that presudo gtid entries are generated only on servers with `read_only` disabled, but in fact they should run on exactly one server throughout entire topology. I'll close this one make another PR with indentation fixes.

PR #70, opened 2017-02-08
Add support for asynchronous call via /api/async-discover/:host/:port

sjmudd:
The existing api call `/api/discover/:host/:port` is synchronous which is fine. However, when doing bulk changes or potentially doing bulk discoveries (such as bootstrapping an empty orchestrator instance with a large list of servers) this process may be slow.

The new call   `/api/async-discover/:host/:port` works asynchronously which avoid such delays.


shlomi-noach (approved):
looking good

PR #73, opened 2017-02-09
Docker image fixes

calind:
It mainly fixes the complaint that it cannot find `templates/layout` and some config defaults.

BTW. do you consider publishing images on docker hub?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,11 +15,11 @@ RUN set -ex \
     && { go get -d github.com/github/orchestrator ; : ; } \
     && cd $GOPATH/src/github.com/github/orchestrator \
     && bash build.sh -b \
-    && cp ./docker/entrypoint.sh /entrypoint.sh \
     && rsync -av $(find /tmp/orchestrator-release -type d -name orchestrator -maxdepth 2)/ / \
     && rsync -av $(find /tmp/orchestrator-release -type d -name orchestrator-cli -maxdepth 2)/ / \
     && cd / \
     && apk del .build-deps \
     && rm -rf /tmp/*
 
+ADD docker/entrypoint.sh /entrypoint.sh
 CMD /entrypoint.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,5 +21,6 @@ RUN set -ex \
     && apk del .build-deps \
     && rm -rf /tmp/*
 
+WORKDIR /usr/local/orchestrator
 ADD docker/entrypoint.sh /entrypoint.sh
 CMD /entrypoint.sh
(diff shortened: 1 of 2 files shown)


docker/entrypoint.sh
```diff
@@ -4,8 +4,8 @@ cat <<EOF > /etc/orchestrator.conf.json
 {
   "Debug": true,
   "ListenAddress": ":3000",
-  "MySQLTopologyUser": "${ORC_TOPOLOGY_USER:-:orchestrator}",
-  "MySQLTopologyPassword": "${ORC_TOPOLOGY_PASSWORD:-:orchestrator}",
```
shlomi-noach, line 8: why this change?

Dockerfile
```diff
@@ -15,11 +15,12 @@ RUN set -ex \
     && { go get -d github.com/github/orchestrator ; : ; } \
     && cd $GOPATH/src/github.com/github/orchestrator \
     && bash build.sh -b \
-    && cp ./docker/entrypoint.sh /entrypoint.sh \
```
shlomi-noach, line 18: explain the removal of this line?

docker/entrypoint.sh
calind: Because it would default to 
```
"MySQLTopologyUser": ":orchestrator",
"MySQLTopologyPassword": ":orchestrator"
```
instead of
```
"MySQLTopologyPassword": "orchestrator"
```
if `ORC_TOPOLOGY_USER` or `ORC_TOPOLOGY_PASSWORD` is not set (Notice the extra `:` in the original version).

Dockerfile
calind: The entrypoint is added later in the `Dockerfile` (`ADD docker/entrypoint.sh /entrypoint.sh`) so that if it's changed `docker build` picks up the change and rebuild the the container.

docker/entrypoint.sh
shlomi-noach: Can you explain why `:orchestrator` is preferable to `orchestrator` as default username/password?

shlomi-noach:
> do you consider publishing images on docker hub?

Not really. I'm not very familiar with docker, and all docket file in this repo are community contributions. I'm unsure what it would take to publish images on docker hub.

docker/entrypoint.sh
calind: It's the other way around. The current version defaults to `:orchestrator`. This PR changes this to `orchestrator`.

shlomi-noach: Oh! You are right, of course. I'm not sure if there was an intention behind `:orchestrator`. Looking at blame/history but I don't find anything that suggests the `:` is of importance. 
Since I'm not familiar with docker, dumb question: is this perhaps a template variable syntax or similar?

calind: That's actaully shell code, which generates a sample config if there is no config file already present at `/etc/orchestrator.conf.json`. The bug was there since my initial commit, I just found it now.

shlomi-noach (approved):
thank you!

PR #74, opened 2017-02-11
Add reference in docs to DiscoveryQueueMaxStatisticsSize and async-discover api call

sjmudd:
Small patch to add some missing documentation.


sjmudd:
Sorry there was only one missing reference in the docs. Now fixed.

PR #75, opened 2017-02-13
Sqlite dependencies

shlomi-noach:
Branching off https://github.com/github/orchestrator/pull/63, this attempts to solve the `go-sqlite3` dependency while retaining the package under `/vendor`


shlomi-noach:
This was merged into #63 . Closing

PR #81, opened 2017-02-16
Make executeCheckAndRecoverFunction show unhandled analysis entries.

sjmudd:
Current code does not make this very visible and the code path should not get here anyway. However, if it does it indicates that the analysis phase is not completely correct or we are not handling the situation correctly. Both situations probably need to be identified and resolved.

### DIFF
--- a/go/logic/topology_recovery.go
+++ b/go/logic/topology_recovery.go
@@ -1209,8 +1209,10 @@ func executeCheckAndRecoverFunction(analysisEntry inst.ReplicationAnalysis, cand
 
 	if checkAndRecoverFunction == nil {
 		// Unhandled problem type
-		log.Warningf("executeCheckAndRecoverFunction: Ignoring unhandled analysisEntry: %+v; key: %+v",
-			analysisEntry.Analysis, analysisEntry.AnalyzedInstanceKey)
+		if analysisEntry.Analysis != inst.NoProblem {
+			log.Warningf("executeCheckAndRecoverFunction: Ignoring unhandled analysisEntry: %+v; key: %+v",
+				analysisEntry.Analysis, analysisEntry.AnalyzedInstanceKey)
+		}
 
 		return false, nil, nil
 	}
@@ -1278,9 +1280,9 @@ func CheckAndRecover(specificInstance *inst.InstanceKey, candidateInstanceKey *i
 				promotedReplicaKey = topologyRecovery.SuccessorKey
 			}
 		} else if recoveryDisabledGlobally {
-			log.Infof("CheckAndRecover: InstanceKey: %+v, candidateInstanceKey: %+v, "+
+			log.Infof("CheckAndRecover: Analysis: %+v, InstanceKey: %+v, candidateInstanceKey: %+v, "+
 				"skipProcesses: %v: NOT Recovering host (disabled globally)",
-				analysisEntry.AnalyzedInstanceKey, candidateInstanceKey, skipProcesses)
+				analysisEntry.Analysis, analysisEntry.AnalyzedInstanceKey, candidateInstanceKey, skipProcesses)
 		} else {
 			go executeCheckAndRecoverFunction(analysisEntry, candidateInstanceKey, false, skipProcesses)
 		}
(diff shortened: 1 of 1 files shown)


go/logic/topology_recovery.go
```diff
@@ -1201,14 +1201,15 @@ func executeCheckAndRecoverFunction(analysisEntry inst.ReplicationAnalysis, cand
...
-	if checkAndRecoverFunction == nil {
+		// Right now this is mostly causing noise with no clear action.
+		// Will revisit this in the future.
+		// case inst.AllMasterSlavesStale:
+		// 	checkAndRecoverFunction = checkAndRecoverGenericProblem
+	default:
 		// Unhandled problem type
+		log.Warningf("executeCheckAndRecoverFunction: Ignoring unhandled analysisEntry: %+v; key: %+v",
+			analysisEntry.Analysis, analysisEntry.AnalyzedInstanceKey)
+
```
shlomi-noach: Incorrect change of logic. Consider the behavior of `case inst.UnreachableMaster:`: it doesn't assign a `checkAndRecoverFunction` but instead issues an immediate `go emergentlyReadTopologyInstanceReplicas(...)`.

It should then `return false, nil, nil`. But in your code change it _does not_ return, but instead proceeds to execute the recovery function. Such function does not exist and proceeding is wrong.

shlomi-noach (changes requested):

go/logic/topology_recovery.go
sjmudd: Ok. Will handle this properly shortly. Thanks for the heads up.

sjmudd:
Hi Shlomi. thanks for brining up the problem with the change I proposed. it does look as if I'd missed the meaning of what was going on here. Bottom line, the original code has a reference to unhandled exceptions and I wanted to see if there were any and why.  I can change the code therefore to do the check properly by using the 

```
if checkAndRecoverFunction == nil
```

and adding a check to log if `if analysisEntry.Analysis != inst.NoProblem`.

Having said that I hadn't realised but in the same function `executeCheckAndRecoverFunction` you call in some cases `go emergentlyXXXX` which seems to be designed to trigger an instant discovery of one or more servers, but currently that's handled outside of the concurrency queueing mechanism.  For discovering a single server this may not be an issue but `emergentlyReadTopologyInstanceReplicas` bypasses the discovery queue which is intended to limit the number of concurrent concurrency attempts.  For a large replication chain you may trigger an extra X number of discoveries which might not be good, and might perhaps trigger later writes to the orchestrator backend to hit `max_connections`.

If you agree that analysis is correct then I'd like to modify these functions to "queue" their discoveries, but rather than doing this "at the end of the queue" to push the discovery request to the beginning so they get serviced first. That seems like a reasonable compromise etc.

This change in behaviour would seem to be fine, unless `executeCheckAndRecoverFunction` repeatedly triggers these `emergentlyXXX` functions, something which seems to be something I'm seeing as basically it bypasses the current FIFO discovery queue.

So maybe this is something to consider?

sjmudd:
I've pushed a further change to log only if the analysis is not `NoProblem` and is not "handled" which I think should resolve your concerns.

sjmudd:
The current patch, if you pull the latest changes in the branch, will only log that something has been missed if the problem is not NoProblem and if `if checkAndRecoverFunction == nil` is called. So no changes in behaviour: just some extra logging which hopefully will be useful. I can then follow up on that if I see (as I am) repeated occurrences of this being triggered.

So would you accept the PR as it is now so we can make progress?

Thanks for your thoughts on this.

go/logic/topology_recovery.go
```diff
@@ -1209,6 +1209,11 @@ func executeCheckAndRecoverFunction(analysisEntry inst.ReplicationAnalysis, cand
 
 	if checkAndRecoverFunction == nil {
 		// Unhandled problem type
+		if analysisEntry.Analysis != inst.NoProblem {
+			log.Warningf("executeCheckAndRecoverFunction: Ignoring unhandled analysisEntry: %+v; key: %+v",
+				analysisEntry.Analysis, analysisEntry.AnalyzedInstanceKey)
+		}
+
```
shlomi-noach, line 1216: :+1:

shlomi-noach (approved):

PR #86, opened 2017-02-24
NodeHealthExpiry: if false ensure the expiry happens on the active node

sjmudd:
All nodes, "cli clients" or active "http" nodes, insert into the node_health table, and keep it up to date while they are running. By default they also try to clean up old entries in this table. If
a large number of http or cli nodes are running then this clean up may trigger unwanted contention on the table, so it has been possible for some time to disable the expiry.  However, this requires a clean up, and the best node to do this is the active orchestrator node.

This patch thus checks if NodeHealthExpiry is false and if so the active node will carry out this task.

### DIFF
--- a/go/process/health_dao.go
+++ b/go/process/health_dao.go
@@ -54,11 +54,7 @@ const (
 	OrchestratorExecutionHttpMode                           = "HttpMode"
 )
 
-var (
-	continuousRegistrationInitiated bool       = false
-	expiryMutex                     sync.Mutex // lock for ExpireAvailableNodes
-	expiryRunning                   bool       // is ExpireAvailableNodes running
-)
+var continuousRegistrationInitiated bool = false
 
 // RegisterNode writes down this node in the node_health table
 func RegisterNode(extraInfo string, command string, firstTime bool) (sql.Result, error) {
@@ -149,16 +145,6 @@ func ContinuousRegistration(extraInfo string, command string) {
 // ExpireAvailableNodes is an aggressive purging method to remove
 // node entries who have skipped their keepalive for two times.
 func ExpireAvailableNodes() {
-	// Ensure this is only running once in each process
-	expiryMutex.Lock()
-	if expiryRunning {
-		expiryMutex.Unlock()
-		log.Errorf("ExpireAvailableNodes: still running")
-		return
-	}
-	expiryRunning = true
-	expiryMutex.Unlock()
-
 	_, err := db.ExecOrchestrator(`
 			delete
 				from node_health
@@ -170,11 +156,6 @@ func ExpireAvailableNodes() {
 	if err != nil {
 		log.Errorf("ExpireAvailableNodes: failed to remove old entries: %+v", err)
 	}
-
-	// show we're done
-	expiryMutex.Lock()
-	expiryRunning = false
-	expiryMutex.Unlock()
 }
 
 // ExpireNodesHistory cleans up the nodes history and is run by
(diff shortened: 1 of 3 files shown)


shlomi-noach:
Let's make it simpler. I'm in favor that only the leader service will run the expiry. Cancel the configuration, cancel most everything. Just call `expireAvailableNodes()` from https://github.com/github/orchestrator/blob/d36867fc60cf6ec980d6b56ef191d7b2c02fd171/go/logic/orchestrator.go#L348, and remove https://github.com/github/orchestrator/blob/d36867fc60cf6ec980d6b56ef191d7b2c02fd171/go/process/health_dao.go#L124

So add row, delete row, capitalize function name.

go/process/health_dao.go
```diff
@@ -120,25 +127,38 @@ func ContinuousRegistration(extraInfo string, command string) {
...
 	go func() {
 		registrationTick := time.Tick(time.Duration(registrationPollSeconds) * time.Second)
 		for range registrationTick {
-			go tickOperation(false)
+			// We already run inside a go-routine so
+			// do not do this asynchronously.  If we
+			// get stuck then we don't want to fill up
+			// the backend pool with connections running
+			// this maintenance operation.
+			tickOperation(false)
```
shlomi-noach, line 139: :+1:

go/process/health_dao.go
```diff
@@ -147,10 +167,18 @@ func expireAvailableNodes() error {
...
 	)
-	return log.Errore(err)
+	if err != nil {
+		log.Errorf("ExpireAvailableNodes: failed to remove old entries: %+v", err)
+	}
+
+	// show we're done
+	expiryMutex.Lock()
+	expiryRunning = false
+	expiryMutex.Unlock()
```
shlomi-noach: While the lock/expiryRunning mechanism is correct, I'm curious why your are so concerned with this? With this PR, this code will only run once per minute. A couple dozen other similar "cleanup" functions similarly run once per minute, and do not have this protection.

If this function blocks for the duration of one minute, there are greater problems, and skipping the next run is _nice_ but wont be the thing to solve those greater problems. I'd suggest removing this protection code.

shlomi-noach (changes requested):

go/process/health_dao.go
sjmudd: To be honest I've seen pile ups and things in the past and am just paranoid. Overly paranoid? maybe but I think that most of this type of event should probably only run once and if it doesn't then we should see there's an issue. I don't have time to check/fix existing code but if I bump into new code then I'd prefer to catch this.  If you're concerned this is too much then I guess this code can be removed but I don't see any harm in keeping it.

shlomi-noach: There is no harm per se, just the complication of the code. We would need to create 30 such mutexes and duplicate this code all over the place.

For this, I don't see the need for paranoia. Again, if you hit this, then your problems are greater.

The solution you offer is slightly complicated in terms of coding. You can achieve the same with two lines of code via https://golang.org/pkg/sync/atomic/#CompareAndSwapInt64, or consolidate the `Lock/Unlock()` code, or refactor it into a class. As it is, it is _correct_, but something that I wouldn't want to see in this function. 

We can take this offline to discuss how the same can be solved in other ways, but meanwhile let's please remove it.

go/process/health_dao.go
sjmudd: Cleaned up further as requested. Hopefully good now.

shlomi-noach (approved):

PR #87, opened 2017-02-24
bump_release_version_and_tag: allow remote not to be 'origin'

sjmudd:
Tiny change to allow bumping the version even if you don't want to push to remote 'origin'.


shlomi-noach (changes requested):
please call the variable something different than `$origin` ; call it `$remote_host` so as not to confuse.

shlomi-noach (approved):

shlomi-noach:
I just pushed those changes I requested. All good now.

PR #93, opened 2017-03-06
graceful-master-takeover attempts setting replication credentials

shlomi-noach:
Storyline: https://github.com/github/orchestrator/issues/78#issuecomment-279930018

On `graceful-master-takeover`, the demoted master typically has no replication credentials (because it has no master of its own).

With this PR:
- shortly before being promoted
  - and assuming replication state is in system tables (`slave_master_info`)
  - and assuming `orchestrator` user can read from that table
  - we read replication credentials from the designated replica (would-be new master).
- immediately following takeover
  - we apply credentials to the old master (now turned replica)

cc @ecortestws @sjmudd


shlomi-noach:
tested to work well

PR #94, opened 2017-03-06
Handle SIGHUP better: now reacts to changes in InstancePollSeconds correctly

sjmudd:
If you send SIGHUP to the active orchestrator process it would not correctly react to InstancePollSeconds changes as caching that's used by this remained unchanged. This patch simply ensures the caching TTL values are adjusted to use the new period.


go/logic/orchestrator.go
```diff
@@ -150,7 +155,10 @@ func discoverInstance(instanceKey inst.InstanceKey) {
 		return
 	}
 
-	if existsInCacheError := recentDiscoveryOperationKeys.Add(instanceKey.DisplayString(), true, cache.DefaultExpiration); existsInCacheError != nil {
+	// Calculate the expiry period each time as InstancePollSeconds
+	// _may_ change during the run of the process (via SIGHUP) and
+	// it is not possible to change the cache's default expiry..
+	if existsInCacheError := recentDiscoveryOperationKeys.Add(instanceKey.DisplayString(), true, instancePollSecondsDuration()); existsInCacheError != nil {
```
shlomi-noach, line 161: May I suggest here using `instancePollSecondsDuration()*2` ? 
the real interval between two pools for an instance may be slightly greater than `instancePollSecondsDuration()` because life. It would be a pity to expire it before the next run.

shlomi-noach (changes requested):

go/logic/orchestrator.go
sjmudd: Why? Your original code uses this to set the default instance expiry:

```
        recentDiscoveryOperationKeys = cache.New(time.Duration(config.Config.InstancePollSeconds)*time.Second, time.Second)
```

So this code change does nothing other than mimic that behaviour.  I can not change the cache's default expiry so I simply set this explicitly when calling `.Add()` when adding a new entry to the cache.

shlomi-noach: Oh, right... I confused this for something else.

shlomi-noach (approved):

PR #95, opened 2017-03-08
Validate config keys

orkestrov:
Related issue: https://github.com/github/orchestrator/issues/82

### Description

Simple config validation. Parsing JSON twice: firstly into a hash map for validation, secondly into a Config struct. Currently only parameter names are checked.


shlomi-noach:
Are you able to add unit tests for:
- confirming valid configuration is not failing the code
- confirming `_xyz` configuration is silently ignored
- confirming invalid configuration is failing the code?

The `log.Fatalf()` makes it hard to test. I suggest that on error `validate()` and `read()` should just return the error, and that the handling code should do the `Fatal` part.

sjmudd:
I also think that given I have seen invalid configurations in the past that by default invalid settings should not break things as this may cause issues for people who are not aware of the sudden changes.
* log all the invalid settings in a single line rather than try to be fatal on the first one you find. It's probably better to log all issues rather than the first one that is found.
* Maybe add a setting `ConfigValidation = [ WARN | FATAL ]`
* Check what happens if the configuration is reread e.g. due to receiving SIGHUP as we probably do not want to kill an orchestrator process that's already running (?)
* Maybe even add a orchestrator -c check_config to validate the configuration without actually running. Might be useful?

So let's look to see if we can get these changes in. The validation is useful, let's not make that cause extra trouble for people but allow us to catch things if we want to.

orkestrov:
I pushed a couple of config tests and also following @sjmudd's advice made sure that orchestrator does not die on SIGHUP when config is erroneous.

shlomi-noach:
@orkestrov haha this is so meta!

So I've just added TravisCI which will run public CI for PRs. As you can see, your PR fails right now.

But why is it failing? It's because the tests use a variable called `AutoRecoveryGloballyEnabled` which is not in the Config.

So your PR just tested itself. Please remove the variable from `tests/integration/orchestrator.conf.json` and let's see if your PR builds.

orkestrov:
@shlomi-noach I used to test with `go test ./go/...` which works. I probably missed the moment we started running integration tests. Is there an intro/requirements for this? When I run `./tests/integration/test.sh` I get this:

`ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/lib/mysql/mysql.sock' (2)
`

shlomi-noach:
@orkestrov like I said, "So I've just added TravisCI which will run public CI for PRs." 

We were having private-only CI and as of now we have public CI.

> Is there an intro/requirements for this?

CI must pass in order to merge a PR.

did you do anything wrong?

No, the CI tests actually _validate_ your PR. And you have the honor of fixing the config file.

orkestrov:
@shlomi-noach hmm I don't mind fixing the config as you suggest. What I'm missing is: how can I test it or get Travis logs? I just removed the `AutoRecoveryGloballyEnabled` as you suggested but it does not seem to have fixed the problem.

shlomi-noach:
> What I'm missing is: how can I test it or get Travis logs?

If you like to test this locally, I'll walk you through this. However you don't need to, aince TravisCI will do it for you.
Are you not able to see a "Details" link to the right of the failing test? And does it not show you the build/test log?

shlomi-noach:
screen shot 2017-04-03 at 14 56 26

shlomi-noach:
@orkestrov regardless, to be able to run the tests locally, you need a mysql server running on your box, with no credentials.

orkestrov:
@shlomi-noach I don't have the "Details" link, but I just figured out that I can click the red cross next to the build to reach the logs. Got it now, thanks for the prompt replies.

shlomi-noach:
@orkestrov interesting. For me the red cross is not clickable and I get a "details" link.

shlomi-noach:
@orkestrov `OAuthClientID` should be `OAuthClientId`

orkestrov:
@shlomi-noach probably we get different results due to some GH frontend experiment:
image

shlomi-noach:
@orkestrov `OAuthClientID` should be `OAuthClientId` -- it's just capitalization

shlomi-noach:
woohoo!

orkestrov:
> @orkestrov OAuthClientID should be OAuthClientId -- it's just capitalisation

Thanks for spotting, added it back now.

orkestrov:
Cool! it worked!

Nice to have the CI btw. This branch has been very meta indeed :-)

PR #97, opened 2017-03-12
Sqlite support -- without sqlite

shlomi-noach:
This is one-step-down from https://github.com/github/orchestrator/pull/63 ; it's all the same: full support for `sqlite` backend, along with schema changes, query changes. However in this PR `sqlutils.go` doesn't actually `import _ "github.com/mattn/go-sqlite3`.

~~(Do note `sqlite3` is included in `vendor/` in this PR)~~ `sqlite3` removed from `vendor/` until I figure out what's the best way to incorporate it (maybe just binary? Future PR to tell)

Reason is still ongoing thoughts about the compilation process. `sqlite3` is heavyweight to compile. While `go build -i` does the job reasonably well, there is a once-in-a-while long compilation time. I'll be looking into using `sqlite3` as `.so` shared file via `go1.7`.

So in this PR everything _supports_ sqlite, but `sqlite` is not strictly included and is effectively disabled.


shlomi-noach:
This branch, and its predecessor `sqlite`, have been running in production for almost a month. I'm going to merge this.

This PR includes rehaul of indexes on all tables. Users of `orchestrator`: first time you will run `orchestrator` where this branch is merged, you will notice a single long-db-deployment operation.

PR #103, opened 2017-03-20
Analysis: dead intermediate master and replicas

shlomi-noach:
Surprisingly, `orchestrator` didn't analyze the case where an intermediate master and all of its replicas are dead.

There is no operation to take on such case (they're all dead, what can you do?)

But detection-wise this mustn't go under the radar. This PR adds this missing diagnostic.

- Added tests for expected behavior


shlomi-noach:
This deploys and runs well and as expected.

shlomi-noach:
detection does the job -- and there's no failover process because there's nothing to be done here.

shlomi-noach:
it's working, all right.

Issue #83, opened 2017-02-20
Slaves lagging by couple of hours are elected as master by orchestrator


adkhare:
We are seeing instances where slaves with couple of hours of lag are elected as masters. Is there any configuration for that not to happen?

@shlomi-noach

sougou:
I believe the root cause was that the master reached its connection limit, and orchestrator couldn't connect to it. The master was otherwise healthy.
At the same time, the replicas were heavily lagged.

I'm thinking we could look at special-casing the connection limit error also.

shlomi-noach:
> I believe the root cause was that the master reached its connection limit, and orchestrator couldn't connect to it

That shouldn't trigger a failover. `orchestrator` would failover only when the master is inaccessible by o`rchestrator` AND all of its replicas.

> We are seeing instances where slaves with couple of hours of lag are elected as masters

Assuming there was a justified failover scenario in the first place, were there any replicas that were not lagging (or lagging by less than "coupld of hours")? 
- If so, what was the configuration for such replicas? `log-bin`? `log-slave-updates`?
- If not, then promoting an hours-lagging replica isn't a good solution, but the least-evil one, and `orchestrator` did the right thing, by current design. We can certainly consider a configuration such as `DoNotFailoverIfAllReplicasAreLaggingByMoreThanSeconds`.

adkhare:
@shlomi-noach we do have following config for bin logs:
log-bin = /opt/data/vitess-data/vt_0102548541/bin-logs/vt-0102548541-bin
log-slave-updates

We have 2 replicas configured and we did not have any of those replicas being caught up completely.

I would love to have a config to avoid electing lagging slaves as master because that would definitely cause issues and eventually cause a data loss.

@sougou please feel free to pitch in.

adkhare:
Other option could be that do a failover and elect the nearest slave to master but dont allow any writes to flow in until everything from relay logs is caught up

sougou:
If connection limit couldn't have triggered the failover, we need to find out what did. Is there some log we can check to see what caused it?

I don't think a replica would be reparented until relay logs are caught up. This raises another question: I assume you have semi-sync turned on. This means that at least one of the replicas must have had all the transactions from the master.

The other question is: could the master have been network-partitioned? If so, commits should have failed or at least timed out during that time. Did you see anything like that?

shlomi-noach:
> we need to find out what did. Is there some log we can check to see what caused it?

[...]

PR #104, opened 2017-03-20
Supporting FailMasterPromotionIfSQLThreadNotUpToDate config

shlomi-noach:
Storyline:

- https://github.com/github/orchestrator/issues/88
- https://github.com/github/orchestrator/issues/83

This PR provides a simple solution to #83 by introducing the `FailMasterPromotionIfSQLThreadNotUpToDate` boolean config.

When this config is `"true"`, and master failover takes place:

- after the topology is healed
- and candidate replica is put on top
- and just before considering this a success
- and before running hooks
- We verify that SQL thread is up to date with IO thread, effectively meaning all relay logs are consumed.
   - and if that is _not the case_, i.e. if the SQL thread is not up do date, i.e. some relay log entries are still pending, then we _fail_ the failover and return with error.

To clarify:
- detection+hooks still take place
- topology still healed
- no post-hooks are executed
- failover returns with error.

cc @sougou @adkhare @bbeaudreault @sjmudd


shlomi-noach:
@sougou @adkhare @bbeaudreault any thoughts on your side?

sougou:
@shlomi-noach The change looks good to me.

PR #111, opened 2017-03-23
Add CountDistinctOkInstanceKeys to discovery aggregate metrics.

sjmudd:
We had a count of the different number of hosts seen in the period,
and the different number of discovery failures but given that in the monitoring
period successful and failed discoveries may happen for the same
host we can't infer the missing metric. This patch adjusts that by
providing the 3 distinct values:

* CountDistinctInstanceKeys
* CountDistinctOkInstanceKeys (new)
* CountDistinctFailedInstanceKeys


shlomi-noach (approved):

go/discovery/aggregated.go
```diff
@@ -99,12 +100,18 @@ func aggregate(results []collection.Metric) AggregatedDiscoveryMetrics {
...
+		// different names
 		x := names["InstanceKeys"]
 		x[v.InstanceKey.String()] = 1 // Value doesn't matter
 		names["InstanceKeys"] = x
-		// failed names
-		if v.Err != nil {
+
+		if v.Err == nil {
+			// ok names
+			x := names["OkInstanceKeys"]
```
shlomi-noach: Can you please make all these strings a type? See for example [AnalysisCode](https://github.com/github/orchestrator/blob/0333dfa6c37bd5c14fd4e9f13cb0a841d1299745/go/inst/analysis.go#L23-L54)

shlomi-noach (changes requested):

sjmudd:
You asked for the strings to be made to types and I've done that now.

shlomi-noach (approved):

PR #125, opened 2017-03-30
end-downtime does not produce error when instance not downtimed

shlomi-noach:
It makes sense to silently say "ok, instance is not downtimed anymore".


shlomi-noach:
Same applies for `end-maintenance`

PR #128, opened 2017-04-02
supporting TravisCI

shlomi-noach:
This will allow external contributors to see CI status


shlomi-noach:
cc @sjmudd

shlomi-noach:
Also adding badges: https://github.com/github/orchestrator/blob/travisci/README.md

PR #129, opened 2017-04-02
fixing Config-dependant constructs

shlomi-noach:
Some constructs were depending on `config.Config` variables even before these would be populated. Such constructs were just assuming `var someCache = ...` within general scope.

With this PR it is possible to listen on a global `ConfigurationLoaded` channel, which produces endless "I have loaded" tokens right after loading.


shlomi-noach:
I don't really like this solution but merging it nonetheless, this will be debt. Will iterate sometime in the future on this.

PR #138, opened 2017-04-05
Fix for Issue#135 (fix target)

sjmudd:
target was hard-coded by default as being linux and thus broke on OSX. Fixed by using `uname` to check the OS and adjust accordingly.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -31,10 +31,9 @@ usage() {
 }
 
 function precheck() {
-  local target build_only
+  local target="$1"
+  local build_only="$2"
   local ok=0 # return err. so shell exit code
-  target="$1"
-  build_only="$2"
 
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -214,8 +213,17 @@ while getopts a:t:p:s:v:dbhr flag; do
...
+if [ -z "$target" ]; then
+	uname=$(uname)
+	case $uname in
+	Linux)	target=linux;;
+	Darwin)	target=darwin;;
+	*)	echo "Unexpected OS from uname: $uname. Exiting"
+		exit 1
+	esac
+fi
+arch=${arch:-"amd64"} # default for arch is amd64 but should take from environment
```
shlomi-noach, line 226: nice!

shlomi-noach (approved):

PR #139, opened 2017-04-05
Monitor mysql hosts which require TLS if we can.

sjmudd:
* existing `MySQLTopologyUseMutualTLS` behaviour will not be modified
* however extend the TLS setup to work for hosts which require TLS if we provide a `MySQLTopologySSLCAFile` configuration.
* use the `MySQLTopologySSLCertFile` and `MySQLTopologySSLPrivateKeyFile` settings if they are provided.
* if no such configuration is provided we can not access such hosts so provide a periodic warning to indicate the issue.


sjmudd:
This is intended to address https://github.com/github/orchestrator/issues/127.

sjmudd:
Note: I moved out the TLS related to code to `go/db/tls.go` to keep things simpler. Hope this makes sense.
Also note the code had to work around a bug (one TLS configuration per server) due to a bug in the go driver which would not allow a shared configuration.  Hence my solution is a tiny bit more complex than I would have liked.

shlomi-noach:
Wow this is a bit over the top for me to review right now, since I'm not using code affected by this change and given the scope of changes I'm not sure whether this will break something or not.
Perhaps @grierj or @rlowe can chime in?

dveeden:
This probably should have been split in two commits to make it easier to review.

1. Move out code to `go/db/tls.go`
2. Add the Upgrade TLS connections code.

This could be 2 commits in one PR or two PR's.

sjmudd:
Shlomi, I understand your concern and Daniel's suggestion seems reasonable. Please close this PR and I'll look to provide you 2 separate PRs to keep things simpler.

sjmudd:
So https://github.com/github/orchestrator/pull/154 is one of these changes. Will pull out the other ones too in chunks that are easier to digest.

grierj:
It would also be nice to see some tests on the the tls.go break out. There's some test certs in the https://github.com/github/orchestrator/blob/master/go/ssl/ssl_test.go that worked well enough for that test setup.

sjmudd:
To be honest the tests for me fail. I'm not sure why. Perhaps there's an expectation on my setup that my config is different. (that is `$ go test ./go/...` fails for the TLS tests, but `$ cd go/db && go test` works fine. I'm not sure of the original design of how this was supposed to be used so I've adapted the previous setup to work with my requirements without touching the previous setup.

If you have thoughts or suggestions on improvements I'm most interested to hear it.

sjmudd:
Closing this as https://github.com/github/orchestrator/pull/242 implements a more complete solution that works on both the cli and orchestratorapp usage.

PR #141, opened 2017-04-06
make the configuration race detector safe.

sjmudd:
The go race detector notices that we may overwrite the configuration
while reading it as there are no locks to prevent this.  This is
most likely to happen when you send SIGHUP to the running process
as it is quite busy.

So replace config.Config with config.Config() and return a copy of
the pointer to the configuration.

Load() Reload() and ForceReload() have been adjusted to use locking
and to also create a new copy of the configuration rather than
modify the existing one. The new copy then replaces the original
configuration and this process is surrounded by locking.  Configuration
modifications by the application are now done via calls which also
lock.

### DIFF
--- a/go/agent/agent_dao.go
+++ b/go/agent/agent_dao.go
@@ -51,12 +51,12 @@ func InitHttpClient() {
 		return
 	}
 
-	httpTimeout := time.Duration(time.Duration(config.Config.HttpTimeoutSeconds) * time.Second)
+	httpTimeout := time.Duration(time.Duration(config.Config().HttpTimeoutSeconds) * time.Second)
 	dialTimeout := func(network, addr string) (net.Conn, error) {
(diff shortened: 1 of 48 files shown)


shlomi-noach (approved):
This looks good. Have you ran this in prod under `-race`?

shlomi-noach:
You need to apply same changes in `*_test.go` files.

shlomi-noach (changes requested):
apply same changes in `*_test.go` files

sjmudd:
Please hold off on applying this. I want to test more first. It's probably ok but also with the other PRs you have pending (the use TLS if necessary and some missing patches of mine) it probably will need  to be adjusted.

sjmudd:
Going to close this. I'll come back to it again later.

PR #144, opened 2017-04-07
Add API calls to provide write buffer metrics

sjmudd:
Somehow I forgot to expose the api calls in a previous PR. This fixes this.

Add the following api calls
* /api/instance-buffered-write-metrics-raw/:seconds
* /api/instance-buffered-write-metrics-aggregated/:seconds


sjmudd:
I can't figure out the right way to push a missing commit I pushed to my tree to get it here so going to open a new pull. Need to figure out this workflow better.

PR #146, opened 2017-04-07
Add API calls to provide write buffer metrics (2nd attempt)

sjmudd:
Somehow I forgot to expose the api calls in a previous PR. This fixes this.

Add the following api calls
* /api/instance-buffered-write-metrics-raw/:seconds
* /api/instance-buffered-write-metrics-aggregated/:seconds


shlomi-noach (approved):

go/inst/write_buffer.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	if s, err := stats.Percentile(stats.Float64Data(writeItems), 95); err == nil {
+		wba.P95InstancesWritten = s
+	}
+	if s, err := stats.Percentile(stats.Float64Data(writeItems), 99); err == nil {
+		wba.P99InstancesWritten = s
+	}
+
+	return wba
+}
```
shlomi-noach, line 150: I'm pretty sure `go-metrics`, which we used under `vendor`, automatically does the `50th`, `75th` etc. percentiles. Can you not use that?

sjmudd: It doesn't work on non-integer values. I guess I could ensure I'm using the time.Duration values and work on that and then convert any final results into the ..Seconds values I'm using but that's not how I implemented this.

If you'd accept the PR as is I'll make an issue to look to remove the use of this extra package if that can be done.

sjmudd: This has been running on my systems since April so it would be good to get these extra metrics into orchestrator.

PR #152, opened 2017-04-18
Some quicker operations

shlomi-noach:
Adding some quick wins for reducing times,

- Iterative `sleep()` following `START SLAVE` (see https://github.com/github/orchestrator/pull/150/commits/4b2bedc97015c392f5947b41c606476bacf07a4b from where this was extracted), only sleeps as much as needed (well, _heuristically_) before checking on `SHOW SLAVE STATUS`.

- `UpdateClusterAliases()` to run on instance-poll intervals as opposed to `1min` intervals. This is useful for first discoveries and for failovers.


shlomi-noach:
When regrouping replicas, master's DC is taken into account, such that two replicas of equal coordinates will be sorted by master's DC (prefer to promote local DC server)

shlomi-noach:
This runs well and I'm going to merge.

PR #155, opened 2017-04-19
TLS functions in go/db/db.go to go/db/tls.go

sjmudd:
This is patch 2 of https://github.com/github/orchestrator/pull/139. which moves some of the files out of db.go to a separate file. No functional change, this just moves things around a bit.

- [x] code is tested via `go test ./go/...`   # well not really. They fail when run against the github master branch (TLS tests fail). Maybe we should figure out how to make them pass or my environment is not as expected. Does this work for you? `$ cd go/db && go test` does work for me.


shlomi-noach:
closing based on @sjmudd 's comment: https://github.com/github/orchestrator/pull/242#issuecomment-317665011

PR #158, opened 2017-04-23
Add reference to the Google groups mysql-orchestrator forum

sjmudd:
This is just a README documentation update.


shlomi-noach:
Honestly I'm not sure what the correct discussion forum should be.

sjmudd:
Maybe but right now there is no place to have a discussion other than opening an issue in github. That doesn't strike me as being a good place where different people can contribute to the same thread.

PR #159, opened 2017-04-23
Lost in recovery: don't end downtime until alive

shlomi-noach:
Storyline: https://fosdem.org/2017/schedule/event/autopsy_mysql/ and https://www.percona.com/live/17/sessions/autopsy-automation-disaster

This PR tackles the following case:

- A cluster is split in two because of a failover scenario
- "old" master has lost replicas, which could not be salvaged
- the "old", now redundant half-cluster comes back alive (e.g. recovering from a network partitioning)
- a failure is detected on the old half-cluster
- the recovery mechanism for the old half-cluster corrupts the good half-cluster by revoking VIPs
  - since both old/new half-clusters think they're the rightful owners of the same cluster.

With this PR:
- a failed master goes into a long downtime
  - no longer configurable
- the dead master's downtime gets continuously renewed and extended. It does not expire
- the downtime is ended when:
  - [x] said old (dead) master is found to be happily replicating (its role has changed, it has been provisioned as a replica)
  - [ ] other conditions... (TBD)

cc @sjmudd


shlomi-noach:
Re: "other conditions...", I'm thinking of:

- [x] the old, dead master is found to be the master of a healthy topology, and no other topology claims to belong to same cluster.

shlomi-noach:
Both conditions now implemented

shlomi-noach:
runs well in prod

PR #162, opened 2017-04-30
The inactive node should always indicate the active node (helps debugging)

sjmudd:
### Description

I normally use orchestrator in debug mode as this provides more output but think that it should be run without this. Therefore important information should be logged. One of these things is indicating on each node which is the active node. This avoids you having to poke around to find out.  This tiny patch changes the log level of these messages from Debug to Info.

       No. This fails but then it fails on github_master for me? See: https://github.com/github/orchestrator/issues/163.


shlomi-noach:
While we're looking into it, as long as CI tests pass this is good to go. We should fix your local tests failure but for now OK to ignore them.

shlomi-noach (approved):

PR #165, opened 2017-05-05
Add govendor support for github.com/martini-contrib/render

sjmudd:
While debugging something I wanted to check if this was the most
up to date code and currently that info was not available. Hence
added the info.  I ran:

```
$ govendor fetch "github.com/martini-contrib/render"
```

which pulled in the latest changes (minimal in this case so no
change) but at least now vendor.json includes the commit that was
used from this vendor software.

The actual render package has no code changes, just an extra test which wasn't there before. Having the info on which commit is in the orchestrator repo makes it easier to notice if it has changed.


vendor/github.com/martini-contrib/render/render.go
```diff
@@ -41,7 +41,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"html/template"
-	"io"
+  "io"
```
shlomi-noach, line 44: Just this one strange indentation problem

shlomi-noach (changes requested):

PR #171, opened 2017-05-11
web: marking promotion-tule: must_not

shlomi-noach:
This PR will show a "ban" icon on an instance whose promotion rule is `must_not`.


shlomi-noach:
works as expected

PR #179, opened 2017-05-18
Add a list of users of Orchestrator

sjmudd:
People may want to know who uses Orchestrator so this list so I am adding booking.com as we've publicly said in the past we use Orchestrator. Hopefully others can add their names too so that people will see that this is a product that is really being used by quite a lot of people.

### DIFF
--- a/docs/toc.md
+++ b/docs/toc.md
@@ -21,6 +21,7 @@
 - [Supported topologies and versions](supported-topologies-and-versions.md)
 - [Bugs](bugs.md)
 - [Contributions](contributions.md)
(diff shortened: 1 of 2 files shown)


shlomi-noach (changes requested):
I think we can drop "used in production" since that will be implied?

sjmudd:
So removed the _used in production_ for booking as it's clear that we are not evaluating this but using on all systems. Also fixed the merge conflict related to the PR referencing presentations.

shlomi-noach (approved):

PR #180, opened 2017-05-18
Add some presentations (more need adding later)

sjmudd:
Add a list of references to presentations about Orchestrator.
I've added one I did yesterday and your previous one at Percona Live but know you've done a lot more. Perhaps they can be added later?

Either way I think it's good for users to see this list and hopefully it will give more context to people looking at Orchestrator for the first time.

### DIFF
--- a/docs/presentations.md
+++ b/docs/presentations.md
@@ -1,0 +1,14 @@
+
+# Orchestrator Presentations
+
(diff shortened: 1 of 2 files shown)


shlomi-noach:
Excellent! I'll add more later. FYI, all my presentations are on https://speakerdeck.com/shlominoach

PR #182, opened 2017-05-21
concurrent reads on ReadTopologyInstanceBufferable

shlomi-noach:
Some reads on `ReadTopologyInstanceBufferable()` are safe to run concurrently. Such are reads that subsequent actions do not rely upon.

Running such reads concurrently reduces instance read latency.


shlomi-noach:
On some remote DC, high latency network, this reduced instance detection time from `~560ms` to `~250ms`. This is a substantial reduction.

For local DC or low latency networks I see little difference (and this is expected).

shlomi-noach:
https://github.com/github/orchestrator/pull/182/commits/1cb9736d1b80413a97323f7a4a8d9a02d2d33fcc further reduces to ~`190ms`

shlomi-noach:
This deploys well.

PR #183, opened 2017-05-22
WIP: orchestrator/raft[/sqlite]

shlomi-noach:
Following up on the design as presented in https://github.com/github/orchestrator/issues/175:

This (Work in Progress) PR introduces the [hashicorp/raft](http://github.com/hashicorp/raft) library for raft consensus.

Notheworthy that `orchestrator` implements a new raft `LogStore` & `StableStore` on top of its relational backend (be it MySQL or SQLite). The relational backend is a classic and simple implementation for that, and saves yet-another-datastore utilizing the existing relational backend.

Many iterations to follow.


shlomi-noach:
Will implement a relational-db snapshot store, to match the raft store and raft log both implemented in relational.

shlomi-noach:
I will shortly merge this PR. This doesn't mean existing `orchestrator` setups suddenly begin speaking `raft`. Existing setups shouldn't see a change (other than the many many side-improvements in this PR).

This PR has been long in the open ; it's now tested in production.

bbeaudreault:
@shlomi-noach does this PR include the sqlite changes too, or will orchestrator still depend on a separate mysql cluster with this as it stands now

shlomi-noach:
@bbeaudreault this PR includes `sqlite`. Assuming you choose to setup a raft cluster, you will be able to choose whether you want each node to run on its own, independent MySQL backend, or on its own `sqlite` backend.

shlomi-noach:
@bbeaudreault still missing for this PR to be merged is documentation for the `raft` setup, where I will clarify all options.

bbeaudreault:
much appreciated @shlomi-noach and very exciting :)

shlomi-noach:
Disabled remote snapshot shipping via raft. We're using empty data snapshots -- the snapshot is the entire database.
However hashicorp/raft attempts to ship the snapshot over transport; consider, for example, an empty node joining the cluster. raft attempts to see it with snapshot + log. We actually don't want that. We reject snapshots we couldn't have created ourselves. 

For a completely newly bootstrapped orchestratpr/raft node joining an existing cluster we expect the user to dump/load or backup/restore reasonably recent data into the new node's DB.

shlomi-noach:
Would anyone care to review these documents?

- https://github.com/github/orchestrator/blob/raft-sqlite/docs/raft.md
- https://github.com/github/orchestrator/blob/raft-sqlite/docs/raft-vs-sync-repl.md
- https://github.com/github/orchestrator/blob/raft-sqlite/docs/high-availability.md
- https://github.com/github/orchestrator/blob/raft-sqlite/docs/using-orchestrator-client.md

🙇

docs/raft.md
```diff
@@ -0,0 +1,137 @@
...
+  "SQLite3DataFile": "/var/lib/orchestrator/orchestrator.db",
+```
+
+`orchestrator` is bundled with `sqlite`, there is no need to install an external dependency.
+
+#### Proxy
+
+You should only send requests to the leader node. To make that happen set up a `HTTP` proxy (e.g HAProxy) on top of the `orchestrator` services.
+
+- Use `/api/leader-check` as health check. At any given time at most one `orchestrator` node will reply with `HTTP 200/OK` to this check; the others will respond with `HTTP 500/Internal server error`.
```
dbussink: How about a 503 when it's not the leader? That has "Service Unavailable" as the meaning and is a bit more meaningful than a general 500. Also means a separate bug in the API that would trigger a 500 is distinguishable from the "not a leader state".

dbussink: Also does this mean that nodes that are not the master also will throw an error if http requests are incorrectly sent to them? Or would it still accept the http requests and cause potential issues? As a future development, would it make sense a node forwards requests to the leader automatically if it's not the leader at the http level?

shlomi-noach: > How about a 503 when it's not the leader?

That's actually a documentation mistake. The returned status is `404 Not Found` (will update doc). I'm open to other responses, though `404` is useful with `HAProxy` with `http-check disable-on-404`.

> Also does this mean that nodes that are not the master also will throw an error if http requests are incorrectly sent to them?

A non-leader node will allow read-only requests and reject any write request. This is reasonably safe. However possibly in a future development a node that is not in the quorum should reject any data-related request.

> would it make sense a node forwards requests to the leader automatically

Interesting idea; let me thing about it for a while and see if I see flaws. Will you please open an issue? (this is applicable also to non-raft setups).

shlomi-noach: also, it's not a problem creating a `/api/leader-check/:statusCodeToReturnForNonLeaders`, so you'd call `/api/status-check/503` and get `503` as response.

dbussink: `404` also makes sense to me and I'll open an issue for the request forwarding bit.

shlomi-noach:
Merging this branch. Further `raft` work to take place on future PRs.
This will be a Pre-Release.

PR #199, opened 2017-06-07
support for force-master-failover command

shlomi-noach:
This PR introduces the `-c force-master-failover` command.

This tells `orchestrator` to assume the master is dead, and that it should kick a full failover, by whatever means it finds necessary to promote a replica.

Usage:

```
orchestrator -c force-master-failover -alias mycluster
```
or
```
orchestrator -c force-master-failover -i some.instance.in.a.cluster
```

`orchestrator` will figure out what the master is, shut its eyes and make pretend it's `dead`, then kick a failover.

Contrast with `force-master-takeover` where the user specifies the replica which must get promoted (more control but potentially more complicated for the user).

Note: this PR extends https://github.com/github/orchestrator/pull/196

### DIFF
--- a/go/app/cli.go
+++ b/go/app/cli.go
@@ -1384,6 +1384,15 @@ func Cli(command string, strict bool, instance string, destination string, owner
 				fmt.Println(promotedInstanceKey.DisplayString())
 			}
 		}
+	case registerCliCommand("force-master-failover", "Recovery", `Forcibly discard master and initiate a failover, even if orchestrator doesn't see a problem. This command lets orchestrator choose the replacement master`):
+		{
+			clusterName := getClusterName(clusterAlias, instanceKey)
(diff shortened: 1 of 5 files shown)


shlomi-noach:
Web interface will look like this:

screen shot 2017-06-08 at 08 28 41

jonahberquist:
Nice! That's a good interface update, too.

shlomi-noach:
Web interface simplified. For master, the "recover" options look like:

screen shot 2017-06-09 at 21 49 36


For intermediate master:

screen shot 2017-06-09 at 21 50 03

I completely removed the `recover-lite` options, as well as various "regroup" options. I'm htinking they're not really interesting, and are mostly confusing at time of failover.

jonahberquist:
Even better 👍

PR #202, opened 2017-06-10
Replace InstanceKey.StringCode() and DisplayString() with String()

sjmudd:
The 3 functions generated the same output and were interchangeable so the code is cleaner and easier to understand if just String() is used.  This suggested patch removes the redundant functions.

It's confusing otherwise and you need to check the code to see what the difference is between the 3 functions when actually there's no difference at all.  There are no functional changes to any of the code.

### DIFF
--- a/go/app/cli.go
+++ b/go/app/cli.go
@@ -262,7 +262,7 @@ func Cli(command string, strict bool, instance string, destination string, owner
 			if err != nil {
 				log.Fatale(err)
 			}
-			fmt.Println(fmt.Sprintf("%s<%s", instanceKey.DisplayString(), destinationKey.DisplayString()))
+			fmt.Println(fmt.Sprintf("%s<%s", instanceKey.String(), destinationKey.String()))
 		}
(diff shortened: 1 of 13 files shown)


sjmudd:
Shlomi, I saw the travis build failed and fixed the failed tests, and have pushed these fixes but don't see a way to do an "update branch".   The workflow here is not clear to me.

shlomi-noach:
@sjmudd I updated the branch for you. It's strange that you don't see the button, since you're a committer.

shlomi-noach:
@sjmudd the reason for the different functions:

- `String()` is by convention a visual string representation of an object; something that is made for humans and debug output.
- As such, I do not wish to rely on `String()` on anything that is logic. So never use it as a key to a map.
- Which is what `StringCode()` is there for.
- Granted, both currently return the same value.
- I'm happy to remove `DisplayString()` though.

In summary, I will not replace `StringCode()` with `String()` but am happy to replace `DisplayString()` with `String()`.

sjmudd:
@shlomi-noach Thanks for clarifying. Perhaps this could be clearer in the code comments?

It's easier to close this PR and create a new one just changing DisplayString() to String() and I'll add a comment along the lines of what you indicate so this is clear.

PR #203, opened 2017-06-11
simplify: InstanceKey.DisplayString() -> String()

sjmudd:
See: https://github.com/github/orchestrator/pull/202.

Simplify by using `InstanceKey.String()` rather than having both `DisplayString()` and `String()`.
A similar function `InstanceKey.StringCode()` has not been changed as requested in the referenced PR.


shlomi-noach:
@sjmudd build fails, still missing replacement on 
```
go/http/api.go:2211: topologyRecovery.SuccessorKey.DisplayString undefined (type *inst.InstanceKey has no field or method DisplayString)
```

PR #209, opened 2017-06-12
make backend read/write concurrency limits configurable.

sjmudd:
Add two new settings with their original defaults:

* BackendDBReadConcurrency:  20
* BackendDBWriteConcurrency: 20

These values may be fine for most people but under very high backend load may become a bottleneck. This patch makes them configurable.  Look at increasing them if discovery performance seems unstable and you know you have a large (> 1000?) number of nodes being monitored.

The previously defined constant (common for both cases) is buried in the code and this bottleneck is not very obvious.  While most people may not need to touch this some of us do. I made a similar patch a long time ago but merging with the upstream (outbrain) tree at some point left me with the original unconfigurable code and I did not notice.


shlomi-noach (commented):
So, I am trying really hard to remove configuration variables and use some reasonable magic values instead. Do these two values _have_ to be in configuration?

sjmudd:
I understand you are trying to remove these magic values and indeed I understand that. The problem seems to be that the only way to figure out what values to use requires making them configurable. For most people the default values you have used may be fine but given some performance Issues I've seen I'd like to make these settings configurable so I can identify the best values to use and should those settings require change I'll be able to modify them as needed.   If it's clear that any settings I use may be better than the default values provided in orchestrator then feedback will be provided suggesting a change to the defaults and the reason why that change may be good.

PR #210, opened 2017-06-13
Adjust db.SetMaxIdleConns for backend orchestrator to a more dynamic value

sjmudd:
Recent issues seen talking to the backend can give a: cannot assign requested address error. The number of open tcp sockets to the backend went over the limit.  This was triggered by the hard-coded value of `db.SetMaxIdleConns(10)` intended to prevent too many backend database
connections being left open triggering the go driver to close connections and shortly afterwards open them again triggering the issue described.

This patch now adjusts the value of `SetMaxIdleConns` to 25% of `MySQLOrchestratorMaxPoolConnections` if this value is larger and so will trigger less reconnections and help aleviate this point of stress.

I chose not to make the setting configurable (yet) as this may confuse people. If you make MySQLOrchestratorMaxPoolConnections higher then the number of idle connections is allowed to grow.

### DIFF
--- a/go/db/db.go
+++ b/go/db/db.go
@@ -1346,7 +1346,23 @@ func OpenOrchestrator() (db *sql.DB, err error) {
 		if !config.Config.SkipOrchestratorDatabaseUpdate {
 			initOrchestratorDB(db)
 		}
-		db.SetMaxIdleConns(10)
+		// A low value here will trigger reconnects which could
+		// make the number of backend connections hit the tcp
+		// limit. That's bad.  I could make this setting dynamic
+		// but then people need to know which value to use. For now
+		// allow up to 25% of MySQLOrchestratorMaxPoolConnections
+		// to be idle.  That should provide a good number which
+		// does not keep the maximum number of connections open but
+		// at the same time does not trigger disconnections and
+		// reconnections too frequently.
+		maxIdleConns := int(config.Config.MySQLOrchestratorMaxPoolConnections * 25 / 100)
+		if maxIdleConns < 10 {
+			maxIdleConns = 10
+		}
+		log.Infof("Connecting to backend: maxConnections: %d, maxIdleConns: %d",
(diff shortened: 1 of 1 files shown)


shlomi-noach:
~~I would go farther and hard code:~~

~~- `MySQLOrchestratorMaxPoolConnections`  to `3`~~
~~- `SetMaxIdleConns` to `3`~~

this will match the changes in https://github.com/github/orchestrator/pull/150/commits/9e1cd18ad708e6eb4c9185d26673b62d851b4d3f

shlomi-noach:
I must have been sleep deprived writing my previous comment. Ignore :)

sjmudd:
Oh good. (not that you were sleep deprived but that you weren't being serious ...)
For context my current setting is:

```
"MySQLOrchestratorMaxPoolConnections”: 1000
```

And with spiky backend traffic the go driver was closing and re-opening backend connections too frequently. The proposed change tries to remedy that.

shlomi-noach:
How about allowing `SetMaxIdleConns(config.Config.MySQLOrchestratorMaxPoolConnections)` ? If we're happy to allocate `MySQLOrchestratorMaxPoolConnections` connections, they may as well be idle in the `go` pool, with `mysql` itself potentially closing them on `wait_timeout` (or not, depending on your config)

sjmudd:
Hi Shlomi,

Basically seeing hundreds or thousands of idle connections which normally shouldn’t be there is not ideal.

If you don’t set this setting then you’ll get the same affect.

I’d rather the number of connections used is as small as reasonable and every time I’ve seen thousands of connections etc it means you have to pay more attention to memory consumption and a number of bad queries all sent through at once can take you down.

The MySQLOrchestratorMaxPoolConnections is to provide a limit but I’d prefer to keep away from that limit if possible.

shlomi-noach (approved):

PR #214, opened 2017-06-24
orchestrator client

shlomi-noach:
Working to create an orchestrator-cli compatible HTTP API based client.
This work will include cleanup of the API, refactoring of command line arguments, help messages and more.

It will become useful in particular for pure-HTTP based setups, (see https://github.com/github/orchestrator/pull/183) where direct backend DB access is not supported.


sjmudd:
One question: why use a separate cli for this ? That feels more complicated that just having a simpler `/path/to/orchestrator.conf.json` file with less options, one of those being something like `"CliUsesHttp": true` which later you could change to the default behaviour.

shlomi-noach:
> why use a separate cli for this ?

@sjmudd perhaps my explanation wasn't clear. This PR adds a _shell script_ that behaves much like the _cli_, but instead invokes HTTP requests on the `orchestrator` API.

This means you don't need an `orchestrator` installed on your "normal" servers and only need `orchestrator` installed on your `orchestrator`-service servers.

The new script is merely a convenience wrapper for curl HTTP requets, with `jq` parsing the correct output for your convenience.

PR #215, opened 2017-06-26
Add PowerAuthGroups: users in these unix groups get admin access

sjmudd:
This simplifies the configuration and avoiding the need to have to explicitly name all PowerAuthUsers. If the authenticated user is in the given list of groups then they will be given "admin" / write access.

So you can change what may be a large list of 
```
  "PowerAuthUsers": ["usera", "userb", "userc", "...", "userz", "api_user1", "api_user2" ],
```

with something like
```
  "PowerAuthGroups": [ "orchestrator_users_group", "orchestrator_api_users_group" ],
```

That is put the user in the right unix group and orchestrator will be happy.  Note: this does assume that you are using something like nginx to provide the authentication service etc.

### DIFF
--- a/go/config/config.go
+++ b/go/config/config.go
@@ -147,6 +147,7 @@ type Configuration struct {
 	HTTPAuthPassword                           string            // Password for HTTP Basic authentication
 	AuthUserHeader                             string            // HTTP header indicating auth user, when AuthenticationMethod is "proxy"
 	PowerAuthUsers                             []string          // On AuthenticationMethod == "proxy", list of users that can make changes. All others are read-only.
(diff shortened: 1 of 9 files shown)


shlomi-noach:
Thank you -- this looks real good. Would you mind moving this away from the `http` package and into the `os` package?

sjmudd:
ok, will do.

shlomi-noach (approved):

sjmudd:
So moved the files as requested to os, and fixed a small issue noted with go vet / golint.

sjmudd:
Thanks for the fast response.

PR #216, opened 2017-06-26
Added MySQL backend to /web/status.

maurosr:
In order to not have to manually check which mysql backend (host:port) orchestrator is configured to talk to but rather see it on the GUI under /web/status.
my_sql_backend

### DIFF
--- a/go/db/db.go
+++ b/go/db/db.go
@@ -1249,8 +1249,7 @@ var generateSQLPatches = []string{
 	`,
 	`
 		ALTER TABLE node_health
-			ADD COLUMN mysql_backend_hostname varchar(128) CHARACTER SET ascii,
-			ADD COLUMN mysql_backend_port smallint UNSIGNED
+			ADD COLUMN db_backend varchar(255) CHARACTER SET ascii NOT NULL DEFAULT ""
 	`,
 }
 
--- a/go/db/db.go
+++ b/go/db/db.go
@@ -1247,6 +1247,11 @@ var generateSQLPatches = []string{
 	`
 		CREATE UNIQUE INDEX host_port_active_recoverable_uidx_topology_failure_detection ON topology_failure_detection (hostname, port, in_active_period, end_active_period_unixtime, is_actionable)
 	`,
+	`
+		ALTER TABLE node_health
+			ADD COLUMN mysql_backend_hostname varchar(128) CHARACTER SET ascii,
+			ADD COLUMN mysql_backend_port smallint UNSIGNED
+	`,
 }
 
 // Track if a TLS has already been configured for topology
--- a/go/process/health_dao.go
+++ b/go/process/health_dao.go
@@ -93,12 +93,12 @@ func RegisterNode(extraInfo string, command string, firstTime bool) (healthy boo
 		}
 	}
 	{
-		db_backend := ""
+		dbBackend := ""
 		if config.Config.IsSQLite() {
-			db_backend = config.Config.SQLite3DataFile
+			dbBackend = config.Config.SQLite3DataFile
 		} else {
-			db_backend = config.Config.MySQLOrchestratorHost + ":" +
-				fmt.Sprintf("%d", config.Config.MySQLOrchestratorPort)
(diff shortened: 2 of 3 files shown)


go/db/db.go
```diff
@@ -1247,6 +1247,11 @@ var generateSQLPatches = []string{
 	`
 		CREATE UNIQUE INDEX host_port_active_recoverable_uidx_topology_failure_detection ON topology_failure_detection (hostname, port, in_active_period, end_active_period_unixtime, is_actionable)
 	`,
+	`
+		ALTER TABLE node_health
+			ADD COLUMN mysql_backend_hostname varchar(128) CHARACTER SET ascii,
+			ADD COLUMN mysql_backend_port smallint UNSIGNED
```
shlomi-noach: columns are `NULL`-able

go/process/health_dao.go
```diff
@@ -215,11 +221,13 @@ func ReadAvailableNodes(onlyHttpNodes bool) (nodes [](*NodeHealth), err error) {
...
-			Token:           m.GetString("token"),
-			AppVersion:      m.GetString("app_version"),
-			FirstSeenActive: m.GetString("first_seen_active"),
-			LastSeenActive:  m.GetString("last_seen_active"),
+			Hostname:             m.GetString("hostname"),
+			Token:                m.GetString("token"),
+			AppVersion:           m.GetString("app_version"),
+			FirstSeenActive:      m.GetString("first_seen_active"),
+			LastSeenActive:       m.GetString("last_seen_active"),
+			MySQLBackendHostname: m.GetString("mysql_backend_hostname"),
```
shlomi-noach: what if column was `NULL`?

go/process/health_dao.go
```diff
@@ -215,11 +221,13 @@ func ReadAvailableNodes(onlyHttpNodes bool) (nodes [](*NodeHealth), err error) {
...
+			MySQLBackendPort:     m.GetInt("mysql_backend_port"),
```
shlomi-noach: same

shlomi-noach (changes requested):
This needs to adapt to support for `sqlite`, in which case there is no `OrchestratorMySQLHost` nor `OrchestratorMySQLPort`.
I'd like to please properly visualize the case where  [`config.Config.IsSQLite`](https://github.com/github/orchestrator/blob/685e502e876e3d6fda4033e212e5651cfbc0e224/go/db/db.go#L1314)

Regardless, you've created the columns `NULL`-able -- how does the app react when the values are `NULL`? 
I'd advise `NOT NULL` columns, as the `golang` handling of `NULL` values is so dirty. An empty value for hostname is a great hint for "no value".

go/db/db.go
maurosr: What do you think of using NOT NULL DEFAULT "" and NOT NULL DEFAULT 0 respectively?

shlomi-noach: That was my suggestion in the comment.

maurosr: Cool. I will then test this and also add support for sqlite

go/process/health_dao.go
```diff
@@ -93,17 +93,22 @@ func RegisterNode(extraInfo string, command string, firstTime bool) (healthy boo
 		}
 	}
 	{
+		db_backend := ""
+		if config.Config.IsSQLite() {
+			db_backend = config.Config.SQLite3DataFile
+		} else {
+			db_backend = config.Config.MySQLOrchestratorHost + ":" +
+				fmt.Sprintf("%d", config.Config.MySQLOrchestratorPort)
```
shlomi-noach: nitpick, but the `golang` way, since you've already chosen `fmt`, is:
```
db_backend = fmt.Sprintf("%s:%d", config.Config.MySQLOrchestratorHost, config.Config.MySQLOrchestratorPort)
```

while at it, please change `db_backend` to `dbBackend`. `golang`'s naming conventions use camelBack

shlomi-noach (changes requested):
see comments inline

shlomi-noach (approved):

PR #219, opened 2017-06-28
fixing concurrency execution of ReplicationLagQuery

shlomi-noach:
https://github.com/github/orchestrator/pull/182 split instance detection and probing queries to run concurrently. The `ReplicationLagQuery` was one of the queries to run concurrently and to be executed straight away.

However that was a mistake, since `ReplicationLagQuery` competes with another query, `SHOW SLAVE STATUS` for the replication lag.

`SHOW SLAVE STATUS` must run first, and `ReplicationLagQuery` must run later, so that it can override, if needed, the `ReplicationLagSeconds` value.

There is no problem for `ReplicationLagQuery` to run concurrently, as long as it gets executed _after_ `SHOW SLAVE STATUS`.

cc @jonahberquist

### DIFF
--- a/go/inst/instance_dao.go
+++ b/go/inst/instance_dao.go
@@ -357,22 +357,6 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
 			}()
 		}
 
-		if config.Config.SlaveLagQuery != "" {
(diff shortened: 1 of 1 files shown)


jonahberquist:
Nice 👍

PR #226, opened 2017-07-04
Hide configuration wait implementation from the user.

sjmudd:
While the change is cosmetic this change hides the implementation details and makes it much clearer exactly what's being done.
It just saves time for people looking at the code and wondering what is going on.

So no functional change but hide the channel used to co-ordinate waiting for the signal and sending the signal the config has been loaded.

This does the following:
* hide the channel name `config.ConfigurationLoaded`
* replace it with a function `config.WaitForConfigurationToBeLoaded()`


shlomi-noach (approved):

shlomi-noach:
I hate everything about the configuration-loaded solution in the first place. We need to find an alternative.
This is just BTW; your PR is good and improved on the current scheme.

PR #228, opened 2017-07-06
Allow for optional post-processing after starting orchestrator

sjmudd:
See: https://github.com/github/orchestrator/issues/227

A solution would be to add the following to /etc/orchestrator_profile

--- snip ---
post_start_daemon_hook () {
        sleep 1
}
--- snip ---

No go changes, only the suggested init script.


sjmudd:
Hm. Please don't push this. The commit message makes a reference to another project. I'll make a new PR to fix this.

PR #230, opened 2017-07-06
Allow for optional post-processing after starting orchestrator

sjmudd:
This is a corrected version of https://github.com/github/orchestrator/pull/228. The commit message referenced a different project.

This is because of https://github.com/github/orchestrator/issues/227

A solution would be to add the following to /etc/orchestrator_profile

--- snip ---
post_start_daemon_hook () {
	sleep 1
}
--- snip ---

No go code changes, just the init script.

### DIFF
--- a/etc/init.d/orchestrator.bash
+++ b/etc/init.d/orchestrator.bash
@@ -19,6 +19,23 @@ SCRIPTNAME=/etc/init.d/$NAME
 
 ulimit -n 16384
 
+# initially noop but can adjust according by modifying orchestrator_profile
+# - see https://github.com/github/orchestrator/issues/227 for more details.
+post_start_daemon_hook () {
+	# by default do nothing
+	:
+}
+
+# Start the orchestrator daemon in the background
+start_daemon () {
+	# start up daemon in the background
+	$DAEMON_PATH/$DAEMON $DAEMONOPTS >> /var/log/${NAME}.log 2>&1 &
+	# collect and print PID of started process
+	echo $!
+	# space for optional processing after starting orchestrator
+	post_start_daemon_hook
+}
+
 # The file /etc/orchestrator_profile can be used to inject pre-service execution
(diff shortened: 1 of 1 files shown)


etc/init.d/orchestrator.bash
```diff
@@ -26,6 +26,23 @@ SCRIPTNAME=/etc/init.d/$NAME
...
+}
+
+# Start the orchestrator daemon in the background
+start_daemon () {
+	# start up daemon in the background
+	$DAEMON_PATH/$DAEMON $DAEMONOPTS >> /var/log/${NAME}.log 2>&1 &
+	# collect and print PID of started process
+	echo $!
+	# space for optional processing after starting orchestrator
+	post_start_daemon_hook
```
shlomi-noach: you must make sure that `post_start_daemon_hook` doesn't generate output, since you catch the output later on -- it's the PID

shlomi-noach (changes requested):
make sure to hide `post_start_daemon_hook`'s output, if there is one.

etc/init.d/orchestrator.bash
sjmudd: Is it not sufficient to add a comment into the existing function?
If not I can make another function call post_start_daemon_hook and redirect output there. It's a bit more convoluted but can be done if you want that.

sjmudd: In my case the fix may work outside of the start_daemon function in which case the problem does not arrive. I'll check and make an approprpriate change.

shlomi-noach: I believe you can `(post_start_daemon_hook 1>&2)`. That should do the trick.

sjmudd: Well go with that then.

sjmudd:
So made change as requested redirecting stdout to stderr.

shlomi-noach (approved):

PR #234, opened 2017-07-09
Add handling for SIGINT (^C)

dveeden:
Related issue: https://github.com/github/orchestrator/issues/233

This tries to restart replication if ^C (SIGINT) is received.
The implementation is probably very naive:
- Assumes cli with one action, might not work with http
- Only adds to the slice of RollbackActions (discard on end of process)
- Testing is absent


go/inst/rollback.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+var rollbackActions []rollbackAction
+
+func AddRollbackAction(action string, argument string) {
+	act := rollbackAction{action, argument}
+	rollbackActions = append(rollbackActions, act)
+}
+
+func Rollback() {
```
shlomi-noach: please make this function return an `error`. I realize it is unlikely we would check the return value calling this function. Nonetheless perhaps in some future testing scenario we would.

go/inst/rollback.go
```diff
@@ -0,0 +1,47 @@
...
+		switch ra.action {
+		case "print_message":
+			log.Infof("%s", ra.argument)
+		case "start_slave":
+			instanceKey, err := NewRawInstanceKey(ra.argument)
+			if err == nil {
+				log.Infof("Rollback: Starting replica %s", ra.argument)
+				StartSlave(instanceKey)
+			} else {
+				log.Debugf("Failed to get instanceKey for %s", ra.argument)
```
shlomi-noach: return `log.Errorf(...)`

go/inst/rollback.go
```diff
@@ -0,0 +1,47 @@
...
+			}
+		default:
+			log.Errorf("Unknown rollback action: %s", ra.action)
```
shlomi-noach: return ...

shlomi-noach: There is a discussion to make here, actually. Do we wish to abort on first sign of trouble or do we wish to brute force anything that works?

shlomi-noach (changes requested):
Overall good idea; however consider the following:

- The `Rollback` operation could itself block. In your specific scenario, it _will_ block. So the `Ctrl^C` would not serve its purpose.
  If anything, make each operation run in its own `goroutine`. But even then I'm unsure if this is right.

go/inst/rollback.go
dveeden: Fixed

go/inst/rollback.go
dveeden: Fixed

go/inst/rollback.go
dveeden: I think doing a best effort here would be best. Try to restart replication where possible and log an error where this doesn't work. Then the admin or automation only has to fix a few machines instead of all machines.

dveeden: Another option would be to just echo the commands needed to fix this.. but with automation that isn't very nice.

shlomi-noach:
Revisiting. @dveeden , in `StopSlave` there's a `AddRollbackAction("start_slave", instanceKey.StringCode())`.

What happens if the function is successful, al ends well, and `5` hours later `Ctrl+C` is hit? Seems to me like this rollback would still apply, and this would be a wrong thing to do (replica shouldn't be started `5` hours later because who knows).

Thoughts on this?

Perhaps `AddRollbackAction` should return a unique token, to the effect of:
```
rollbackToken := AddRollbackAction("start_slave", instanceKey.StringCode())
defer CancelRollbackAction(rollbackToken)
```

dveeden:
Yes that sounds like a better way.

shlomi-noach:
Further thinking about this: does the `^C` handling cause the function to terminate, thus running the `defer` anyway?

dveeden:
I don't have time to work on this in the next month(s). But @sjmudd said that he or @maurosr might want to continue with it.

shlomi-noach:
@dveeden thanks for letting me know; the issue is good and interesting, and I foresee a couple complications along the way.

PR #239, opened 2017-07-13
Updated github.com/go-sql-driver/mysql to the master HEAD.

maurosr:
Fixes bug https://github.com/go-sql-driver/mysql/issues/536 so now the same TLS config does not need to be registered for each host and with different name (I am working on the ability to have mixed TLS and non-TLS connections to different nodes of the topology).

Also, an interesting effect noticed: after the upgrade of the driver there was a 20% decrease of median poll time to orchestrator backend.


shlomi-noach:
> Also, an interesting effect noticed: after the upgrade of the driver there was a 20% decrease of median poll time to orchestrator backend.

very interesting indeed.

shlomi-noach:
@maurosr can you kindly ping me after this had been deployed at your company for a few days in production? I like nothing better than knowing this runs successfully on X,000 servers.

maurosr:
Will do (3 hours running in production by now and is looking good).

maurosr:
@shlomi-noach everything looks fine in production with the driver upgrade.

shlomi-noach:
Ok cool. I'll merge shortly

maurosr:
Sorry, forgot to make a new branch for TLS. I'll revert and create another PR for that.

PR #242, opened 2017-07-21
Added ability to mix TLS and non-TLS connections to nodes.

maurosr:
Encapsulated all TLS logic in db/tls.go.

Created new DB table database_instance_tls to record TLS requirement per host/port. Tested both in MySQL and sqlite3.

Set a configurable TLS info cache TTL (TLSCacheTTLFactor * InstancePollSeconds).

New metrics exposed: "instance_tls.read", "instance_tls.write", "instance_tls.read_cache", "instance_tls.write_cache",

Note that for this to work the go-sql-driver needs to be updated ( #239 ) otherwise having a single TLS config for more than one host is not possible due to issue go-sql-driver/mysql#536 .

### DIFF
--- a/go/config/config.go
+++ b/go/config/config.go
@@ -260,7 +260,7 @@ func newConfiguration() *Configuration {
 		MySQLOrchestratorMaxPoolConnections:        128, // limit concurrent conns to backend DB
 		MySQLOrchestratorPort:                      3306,
 		MySQLTopologyUseMutualTLS:                  false,
-		MySQLTopologyUseMixedTLS:                   false,
+		MySQLTopologyUseMixedTLS:                   true,
 		MySQLOrchestratorUseMutualTLS:              false,
 		MySQLConnectTimeoutSeconds:                 2,
 		MySQLOrchestratorReadTimeoutSeconds:        30,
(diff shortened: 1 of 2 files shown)


shlomi-noach:
Is this PR based on another PR suggested earlier? I seem to remember similar changes.

shlomi-noach:
Seems to be based  on https://github.com/github/orchestrator/pull/155

maurosr:
It is related to the same issue, but that PR solves the problem by having an in-memory map of hosts that require TLS, hence it only works for orchestrator service, not for commands run as CLI. This PR solves that by writing this map in the database also.
Another difference is that it pings the host when setting up the connection instead of catching the TLS error at first connection attempt some time after, which seems to me a bit cleaner while not impacting overall performance.

sjmudd:
Basically the previous PR which I should close would only work on discovery with an orchestratorapp. That was "ok" but not ideal.  The first connection would cache a flag (in RAM) "this needs tls" and use it for the next discovery.

The work that @maurosr has done now remembers this behaviour (stored in a table per host:port) and thus it works for both cli and orchestratorapp usage.  As mentioned the new code also includes a cache to minimise required accesses to this table.

I'll close the old PR as obsolete and reference this one.

shlomi-noach (approved):

PR #254, opened 2017-08-08
raft hostnames resolving to IP

shlomi-noach:
Addresses https://github.com/github/orchestrator/issues/253

with this PR, `orchestrator` will resolve raft hostnames to IP addresses before handing over the details to the raft library.

This supports `orchestrator` configuration to specify hostnames rather than IP addresses.

cc @bbeaudreault


bbeaudreault:
does not solve the problem completely but will help configuration. LGTM

PR #255, opened 2017-08-10
Updated pem test certs in go/ssl_test.go to fix failing tests.

maurosr:
Tests in go/ssl_test.go were failing due to to a change in crypto/x509 library. More specifically in https://github.com/golang/go/blob/master/src/crypto/x509/x509.go#L951

@dveeden aided in the investigation and has more knowledge about the specific issue.


shlomi-noach:
Will look into it -- thank you so much!

shlomi-noach:
@maurosr thank you very much!

maurosr:
np :)

PR #265, opened 2017-08-17
handling nil replicas

shlomi-noach:
Fixes https://github.com/github/orchestrator/issues/263

This filters out `nil` replicas, which can happen on network error.


sjmudd:
+1

PR #279, opened 2017-08-28
Improve selection of server chosen to replace a failed IM

sjmudd:
This PR tries to resolve an issue of a development server configured with a promotion rule as `must_not` being promoted when an intermediate master died.

As far as I can see in `GetCandidateSiblingOfIntermediateMaster` there are 2 mistakes:
* the candidate siblings are not always chosen first (if such a sibling exists)
* any sibling configured with a `promotion_rule` of `must_not` is not excluded

This PR attempts therefore to address both issues, so the order of checking in the existing function is modified so that candidates are chosen first, and also servers configured not to be promoted are excluded, marked [*].

The logic becomes:

* search for IsCandidate in same dc & env
* search for IsCandidate somewhere else
* search for sibling in the same DC & ENV [*]
* search for any remaining valid sibling [*]

The first match is returned to the caller.

One remaining issue I think may exist after looking at this code is that if a IM fails then a sibling of the IM is chosen as a replacement IM. This therefore requires a sibling to exist. However, if no siblings exist recovery is still possible if you point the slaves at the master of the failed IM. In theory that should always work too. It's not clear if current orchestrator code handles this and if it does not then I'd expect a new issue should be created (but I'm not going to do that as I haven't checked that code path).

Code not actually checked as I don't have a good enough test environment to trigger such a scenario at the moment. Hopefully the change in logic is correct and this change will thus be acceptable.

### DIFF
--- a/go/logic/topology_recovery.go
+++ b/go/logic/topology_recovery.go
@@ -929,10 +929,10 @@ func GetBestMasterReplacementFromAmongItsReplicas(topologyRecovery *TopologyReco
 	return nil, fmt.Errorf("GetBestMasterReplacementFromAmongItsReplicas: cannot find replacement")
 }
 
-// GetCandidateSiblingOfIntermediateMaster chooses the best sibling of a dead intermediate master
-// to whom the IM's replicas can be moved.
+// GetCandidateSiblingOfIntermediateMaster chooses the best sibling
+// of a dead intermediate master under which the IM's replicas can
+// be moved.
 func GetCandidateSiblingOfIntermediateMaster(topologyRecovery *TopologyRecovery, intermediateMasterInstance *inst.Instance) (*inst.Instance, error) {
-
 	siblings, err := inst.ReadReplicaInstances(&intermediateMasterInstance.MasterKey)
 	if err != nil {
 		return nil, err
@@ -943,11 +943,15 @@ func GetCandidateSiblingOfIntermediateMaster(topologyRecovery *TopologyRecovery,
 
 	sort.Sort(sort.Reverse(InstancesByCountReplicas(siblings)))
 
-	// In the next series of steps we attempt to return a good replacement.
-	// None of the below attempts is sure to pick a winning server. Perhaps picked server is not enough up-todate -- but
-	// this has small likelihood in the general case, and, well, it's an attempt. It's a Plan A, but we have Plan B & C if this fails.
+	// Attempt to return a good replacement.
+	//
+	// None of the below attempts is sure to pick a winning
+	// server. Perhaps picked server is not enough up to date,
+	// but this has small likelihood in the general case and,
+	// well, it's an attempt. It's a Plan A, but we have Plan B &
+	// C if this fails.
 
-	// At first, we try to return an "is_candidate" server in same dc & env
+	// 1. search for IsCandidate in same dc & env
 	AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("searching for the best candidate sibling of dead intermediate master %+v", intermediateMasterInstance.Key))
 	for _, sibling := range siblings {
 		sibling := sibling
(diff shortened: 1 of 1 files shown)


shlomi-noach:
> One remaining issue I think may exist after looking at this code is that if a IM fails then a sibling of the IM is chosen as a replacement IM. This therefore requires a sibling to exist. However, if no siblings exist recovery is still possible if you point the slaves at the master of the failed IM. In theory that should always work too. It's not clear if current orchestrator code handles this and if it does not then I'd expect a new issue should be created (but I'm not going to do that as I haven't checked that code path).

https://github.com/github/orchestrator/blob/8665b6a544914fe63e3e02714c40f313b6427090/go/logic/topology_recovery.go#L1035

https://github.com/github/orchestrator/blob/8665b6a544914fe63e3e02714c40f313b6427090/go/logic/topology_recovery.go#L1041

https://github.com/github/orchestrator/blob/8665b6a544914fe63e3e02714c40f313b6427090/go/logic/topology_recovery.go#L1055

https://github.com/github/orchestrator/blob/8665b6a544914fe63e3e02714c40f313b6427090/go/logic/topology_recovery.go#L1067-L1068

shlomi-noach:
I think some code has diverged from comments. Let me look into this again.

shlomi-noach:
No it hasn't. Let me point you to a different PR.

shlomi-noach:
https://github.com/github/orchestrator/pull/281 fixes the issue in a different way.

shlomi-noach:
> As far as I can see in GetCandidateSiblingOfIntermediateMaster there are 2 mistakes:
>    the candidate siblings are not always chosen first (if such a sibling exists)

not so. Candidates are searched first:

https://github.com/github/orchestrator/blob/8665b6a544914fe63e3e02714c40f313b6427090/go/logic/topology_recovery.go#L952-L961
^ first searches for a candidate in same DC

https://github.com/github/orchestrator/blob/8665b6a544914fe63e3e02714c40f313b6427090/go/logic/topology_recovery.go#L962-L971
^ then prefers a non-candidate, but in same DC
(wrongly admits a `must_not`, see later on)

https://github.com/github/orchestrator/blob/8665b6a544914fe63e3e02714c40f313b6427090/go/logic/topology_recovery.go#L962-L971
^ then looks for a candidate _anywhere_ (ie other DC)

https://github.com/github/orchestrator/blob/8665b6a544914fe63e3e02714c40f313b6427090/go/logic/topology_recovery.go#L980-L987
^ then finds _any valid server_ anywhere
(wrongly admits a `must_not`, see later on)


>    any sibling configured with a promotion_rule of must_not is not excluded

correct. Fixed in #281

shlomi-noach (changes requested):
I'm suggesting #281 instead

sjmudd:
So I see that https://github.com/github/orchestrator/pull/281 seems better. However, I disagree on the second check which chooses a server in the same datacentre rather than a candidate in a different data centre. I think that if I choose X candidate servers explicitly then I prefer to use these servers rather than another server that's not a candidate even if it's in the same DC.

Maybe "it depends" but my thinking is that this should be preferable. Note: https://github.com/github/orchestrator/pull/281 does not change the original order and therefore is not I think completely correct.

shlomi-noach:
"it depends" is the right wording. Do you prefer your 20 replicas to have an intermediate master from another DC? Leading to 20 cross-DC connections?
A "candidate" server could conceptually be associated with a data center. "these are my candidates for this data center".

In the case of intermediate master the choice of server is less critical than in the case of master promotion. Can you please illustrate a case for picking a candidate server from a remote DC as opposed to picking a `neutral` (not `candidate`, not `must_not`) server in a local DC?

sjmudd:
The original problem of promoting a server configured as "must not" has been solved by https://github.com/github/orchestrator/pull/281 so perhaps therefore I should close this PR and open an "issue" about the correct behaviour of choosing a replacement intermediate master and the best solutions to this problem.

So going to do that now.

PR #281, opened 2017-08-29
Intermediate master recovery: exclude must_not servers

shlomi-noach:
A different, simpler take on https://github.com/github/orchestrator/pull/279

This PR makes sure to exclude servers which have `must_not` promotion rule from being promoted as result of `RecoverDeadIntermediateMaster`.

cc @sjmudd

### DIFF
--- a/go/logic/topology_recovery.go
+++ b/go/logic/topology_recovery.go
@@ -826,6 +826,9 @@ func isValidAsCandidateSiblingOfIntermediateMaster(intermediateMasterInstance *i
 	if !isGeneralyValidAsCandidateSiblingOfIntermediateMaster(sibling) {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


sjmudd:
Hi Shlomi,

This looks good to me and indeed you put the "must not" check where other validity checks are already made where a server is checked if it's a potential candidate. Thanks for checking this and coming up with a cleaner solution.

However, the ordering of the checks in https://github.com/github/orchestrator/pull/279 was changed deliberately as I feel that explicitly labelled candidates should have preference over other boxes even if they're in a different data centre and that's not currently the case.

shlomi-noach:
OK I'm going to test this for a while and merge, and discuss intermediate master priorities separately.

sjmudd:
Makes sense (to apply this).

PR #282, opened 2017-08-29
using interpolateParams for both backend and topology

shlomi-noach:
This PR sets `interpolateParams=true` in the DB connection, so that the driver doesn't use prepared statements on the MySQL side but rather on client side.

Hopefully this can reduce some probe time.

### DIFF
--- a/go/db/db.go
+++ b/go/db/db.go
@@ -1313,7 +1313,7 @@ func OpenTopology(host string, port int) (*sql.DB, error) {
 }
 
 func openTopology(host string, port int, readTimeout int) (*sql.DB, error) {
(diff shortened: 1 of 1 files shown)


shlomi-noach:
I don't see savings in probe time, having deployed this in production and observing inter-DC and cross-DC probing.

sjmudd:
This was mainly used for talking to the MySQL backend and of course depends on the number of servers being polled. The issue iirc I saw was not necessarily the actual time for this but the extra concurrency and doubling of the number of queries on the backend MySQL server. At least in my case the change helped performancewise and is still the default setting in my environment so it's good to not have to remember to configure this explicitly.

shlomi-noach:
To clarify this now applies to both backend servers as well as topology servers.

PR #291, opened 2017-09-10
Topology ascii orchestrator client

shlomi-noach:
Apparently the JSONifying of the ASCII topology compresses spaces within text strings. I'm not sure why that is and spent far too long trying to fix this.

This PR just gives up and prints non-breakable spaces instead of spaces, resulting with nice formatting.

cc @tomkrouper


tomkrouper (approved):
yay technology!

PR #299, opened 2017-09-17
Sanpshot data persist and restore: auto-join nodes

shlomi-noach:
This PR introduces snapshot data. This was intentionally excluded from the `orchestrator/raft` implementation and now is the time to pay the debt.

Previously, it was assumed the snapshot is the relational data. Which is true: it _is_ the de-facto data.

However this prevents an important operation: the joining of a newly provisioned node that has no data. The `hashicorp/raft` library fully supports provisioning of new nodes via snapshot+delta ; however it is on the app to implement snapshot data dump+restore. Well, here it is.


shlomi-noach:
Ugh, I accidentally incorporated changes from https://github.com/github/orchestrator/pull/298. Will close this, cherry pick my changes, open a new PR.

PR #316, opened 2017-10-17
Use raft advertise

samiahlroos:
Related issue: https://github.com/github/orchestrator/issues/312

### Description

This PR adds possibility to set `RaftAdvertise` in orchestrator config, and if set, uses that when setting up Raft communications.

Previously orchestrator would always set the `advertise` address to whatever RaftBind was set to, which fails if nodes need to communicate via NAT gateways.

If RaftAdvertise is not set, RaftBind is used which is same behavior as before this change.

### DIFF
--- a/docs/configuration-raft.md
+++ b/docs/configuration-raft.md
@@ -47,6 +47,8 @@ as well as this:
   ],
 ```
 
+If your orchestrator/raft nodes need to communicate via NAT gateways, you can additionally set "RaftAdvertise" to IP or hostname which other nodes should contact. Otherwise other nodes would try to talk to the "RaftBind" address and fail.
+
 ### Backend DB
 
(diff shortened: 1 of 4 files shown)


docs/configuration-raft.md
```diff
@@ -47,6 +47,8 @@ as well as this:
   ],
 ```
 
+If your orchestrator/raft nodes need to communicate via NAT gateways, you can additionally set "RaftAdvertise" to IP or hostname which other nodes should contact. Otherwise other nodes would try to talk to the "RaftBind" address and fail.
```
shlomi-noach, line 50: Thank you for documenting this.

shlomi-noach (approved):

tyagi-kuldeep (commented):
sdgdfgdf

shlomi-noach:
Thank you!

PR #320, opened 2017-10-19
Native support for consul

shlomi-noach:
Beginning work to natively support [Consul](https://github.com/hashicorp/consul).

`orchestrator` will use Consul's KV store to:

- populate known masters (key: alias, value: master)
- update upon master failure


shlomi-noach:
It is insane to add `800` files just to run a `consul` client that can be re-implemented with simple HTTP client. Just saying.

shlomi-noach:
No. Really, it's insane. I just need the equivalent of

```shell
curl -v "http://localhost:8500/v1/kv/path/to/key" 2> /dev/null | jq -r ".[0].Value" | base64 --decode
```

It seems like I'm still short of a few dependencies.

I'm switching to https://github.com/armon/consul-kv, which is unsupported and DEPRECATED, but just runs a simple client that gets the job done.

shlomi-noach:
Replaced by https://github.com/github/orchestrator/pull/341

PR #326, opened 2017-10-24
Cluster page UI changes

shlomi-noach:
- Better alignment
- Removal of background tile
- compact alerts


shlomi-noach:
Changes look like this:

screen shot 2017-10-24 at 14 41 50

PR #338, opened 2017-11-02
Sparser binlog coordinates history; some cleanup

shlomi-noach:
With this PR the recorded history of `database_instance_coordinates_history` becomes sparser: once per minute, as opposed to once per poll-seconds.

`database_instance_coordinates_history` is _mostly_ used when relocating servers via pseudo-GTID ; recorded coordinates act as a heuristic optimization, where the search for pseudo-GTID can heuristically begin not from the top of the binary log, but from a point close to its end. This saves a lot of binlog search time.

However, the high frequency for recording such coordinates is not required. The magic number for searching pseudo-GTID entries stands at "two minutes ago" ; which means resolution of, say, once-per`5sec` is over the top. A resolution of once per minute is fine, and suggests "two minutes ago" _could_ imply "almost three minutes ago". That's fine.

### DIFF
--- a/go/db/generate_patches.go
+++ b/go/db/generate_patches.go
@@ -486,4 +486,8 @@ var generateSQLPatches = []string{
 		ALTER TABLE node_health
 			ADD COLUMN db_backend varchar(255) CHARACTER SET ascii NOT NULL DEFAULT ""
 	`,
+	`
+		ALTER TABLE node_health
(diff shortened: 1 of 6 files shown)


shlomi-noach:
Also this PR fixed health check interval, as well as uses `1sec` caching for `HealthTest()` results (so that load balancers / monitoring tools don't generate database write load by merely asking for health)

PR #341, opened 2017-11-05
KV store: internal key/value + Consul + ZooKeeper support

shlomi-noach:
Supersedes https://github.com/github/orchestrator/pull/320, this PR introduces support for key-value stores:

- internal
- Consul
- ZooKeeper

the immediate objective is to write cluster's master identities to KV, as well as update such entires upon failover.

Users may then use functionalities such as `consul-template` to update/populate their proxy configs or client configs with identities of masters.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -25,6 +25,7 @@
 
 #### Failure detection & recovery
 - [Topology recovery](topology-recovery.md): failure detection and recovery
+- [Key-Value stores](kv.md): master discovery for your apps
 
 #### Various
 - [Docker](docker.md)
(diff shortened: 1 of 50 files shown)


shlomi-noach:
ZooKeeper remains unsupported, but with an existing template.

shlomi-noach:
I'm going to merge this without actually coding the ZooKeepr client. 

@samiahlroos, @sroysen, I believe you have an interest in Zk? Would you like to implement this missing piece in a future PR?

sroysen:
Thanks for the heads up @shlomi-noach. 
We might give it a shot, hopefully, in a not very distant future.

PR #345, opened 2017-11-14
Removed status simple check; normal check caches response

shlomi-noach:
Solves https://github.com/github/orchestrator/issues/145

`StatusSimpleHealth` is deprecated and has no impact whatever the value is. Health checks are always complete and full. However, they cache response for one second making call to `/api/status` more lightweight.

### DIFF
--- a/go/config/config.go
+++ b/go/config/config.go
@@ -190,7 +190,6 @@ type Configuration struct {
 	SSLCAFile                                  string            // Name of the Certificate Authority file, applies only when UseSSL = true
 	SSLValidOUs                                []string          // Valid organizational units when using mutual TLS
 	StatusEndpoint                             string            // Override the status endpoint.  Defaults to '/api/status'
-	StatusSimpleHealth                         bool              // If true, calling the status endpoint will use the simplified health check
 	StatusOUVerify                             bool              // If true, try to verify OUs when Mutual TLS is on.  Defaults to false
 	AgentPollMinutes                           uint              // Minutes between agent polling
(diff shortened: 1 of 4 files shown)


sjmudd:
Shlomi, just curious: "StatusSimpleHealth is deprecated and has no impact whatever the value is."
Is this logged? While it may be people don't look at the logs it might make people pay more attention and if you want to deprecate something then making this more visible helps (a bit).

I personally don't care but if you want to change stuff and deprecate old/useless behaviour it seems helpful to add some logging to indicate that. Looking at the code there's a sort of comment about maybe removing it but int he call that's made atm `process.SimpleHealthTest` does not make it clear that the test should go away.  Just curious.

shlomi-noach:
@sjmudd This behavior was originally added by a Square contribution, where they didn't want load-balancers or other monitoring services, which continuously check health, to cause database writes as result of said checks.

Their solution was to create a SimpleHealthCheck that doesn't actually write to database per check.

About a week ago I changes the behavior of the "normal" check to cache result for one second. This makes the SimpleHealthCheck hack irrelevant, which means wveryone can run the same checks, and the configuration variable becomes obsolete. You will get similar behavior whether or not you've configured `StatusSimpleHealth` variable.

I am not removing API endpoints and people will keep calling same API and get good results.

As for removing the variable itself, I still need to apply @orkesstrov's patch that compares config variables with existing variables.

sjmudd:
Thanks for clarifying.

PR #298, opened 2017-09-17
 Replaced stopwatch with a shared-stopwatch suitable for sharing between go routines

shlomi-noach:
resubmission of https://github.com/github/orchestrator/pull/273 by @maurosr , which I merged into this branch for testing.

PR #346, opened 2017-11-14
Add DiscoveryIgnoreHostnameFilters and deprecate DiscoveryIgnoreReplicaHostnameFilters

sjmudd:
It may be necessary to filter out accessing the master as it may not be reachable (e.g. if in an external network) so this change changes the use of `DiscoveryIgnoreReplicaHostnameFilters` to the more general name `DiscoveryIgnoreHostnameFilters`.

Use of the old variable name will still work but will generate a deprecation warning.

Additional extra checks are added to ensure that a host that we want to ignore really is ignored (the config may change on a running orchestrator)


sjmudd:
Note: requires https://github.com/github/orchestrator/pull/298 or at least there are some issues seen if this patch is not applied first.

chrisboulton:
@shlomi-noach We're super interested in this change for a couple of things in our environment.. Is there anything we can do to help you test this, and what sounds like the requirement of #298 (I've not connected the requirement of that one to this, yet) to get them brought down into an upcoming release?

shlomi-noach:
This PR makes sense to me, but I'm having trouble with https://github.com/github/orchestrator/pull/298. @sjmudd looking deeper, I'm unsure how this PR depends on https://github.com/github/orchestrator/pull/298 ; I think it can be applied without https://github.com/github/orchestrator/pull/298

dougfales:
@sjmudd  we are interested in using this setting, or a similar one, to also ignore hosts on specific ports (see https://github.com/github/orchestrator/issues/923). I was wondering, would you consider allowing this new setting to match against the full host:port instead of just the hostname? Also, is there something blocking this PR? It seems to improve the existing functionality of this setting quite a bit, so I'd love to see it get merged.

PR #363, opened 2018-01-02
Fixing duplicate detection

shlomi-noach:
Work in progress: we've seen a recurring case where same master failure was detected twice, the second detection immediately follows up on the recovery.
This is explainable to some part by the fact the recovery acknowledges the detection. However, the recovery also downtimes the dead master, and so it should be skipped on `CheckAndRecover`.

This PR looks into the issue.


shlomi-noach:
The problem was caused due to internal scripting.

PR #364, opened 2018-01-04
recovery poll is more aggressive, and non-reentrant for safety

shlomi-noach:
In further reducing response time to failures, in this PR:
- `RecoveryPollSeconds` config variable is removed (and at this time silently ignored)
- It is implicitly set to `1sec`
- the recovery logic is thus invoked every `1sec`
- the recovery logic is non-reentrant, so that if it takes more than 1sec to execute, the next iteration will be skipped.
- The non-reentrance logic only applies to automated checks/failovers; a manual command will not block

### DIFF
--- a/conf/orchestrator-sample.conf.json
+++ b/conf/orchestrator-sample.conf.json
@@ -93,7 +93,6 @@
   "SkipBinlogEventsContaining": [],
   "ReduceReplicationAnalysisCount": true,
   "FailureDetectionPeriodBlockMinutes": 60,
-  "RecoveryPollSeconds": 10,
   "RecoveryPeriodBlockSeconds": 3600,
   "RecoveryIgnoreHostnameFilters": [],
   "RecoverMasterClusterFilters": [
--- a/conf/orchestrator-simple.conf.json
+++ b/conf/orchestrator-simple.conf.json
@@ -51,7 +51,6 @@
   "#": "Recovery related config:",
   "#": "-----",
   "FailureDetectionPeriodBlockMinutes": 60,
-  "RecoveryPollSeconds": 10,
   "RecoveryPeriodBlockSeconds": 3600,
   "RecoveryIgnoreHostnameFilters": [],
   "RecoverMasterClusterFilters": [
--- a/docs/configuration-failure-detection.md
+++ b/docs/configuration-failure-detection.md
@@ -6,12 +6,11 @@ Recovery is discussed in [configuration: recovery](configuration-recovery.md)
 
 ```json
 {
-  "RecoveryPollSeconds": 5,
   "FailureDetectionPeriodBlockMinutes": 60,
 }
 ```
 
-In the above `orchestrator` will run detection every `5` seconds.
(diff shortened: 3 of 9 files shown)


shlomi-noach:
The reasoning for this change:
- Remove a configuration variable.
- Be aggressive about failure checks.

Setups with X,000s of servers will benefit from this, since they will not need to do guesswork for the time it takes for `RecoveryPollSeconds` to complete.
Setups with smaller (X00s or X0s) of servers will benefit from this, since failure checks will run every second.

serdarsahin:
Hi Shlomi,

It is a known problem on the cloud services like AWS that you might experience intermittent network issues. So, we'd like to wait for 15-20 seconds before failing over the slave database to make sure it is not something intermittent. How can we control that behavior now?

shlomi-noach:
@serdarsahin if your method was to use sparser checks then that wasn't reliable, either. You would statistically get 50% of what you want 50% of the time...

So I'd like to both clarify and ask: `orchestrator` will not claim a master dead until both `orchestrator` cannot see it, as well as all of the replicas can't see it. This probing in itself has nothing to do with the recovery poll.

What's more interesting is how soon do replicas say "oh hey the master is dead". Can you please let me know:

- What, in your opinion, is a failover scenario?
- Are you comfortable with master being unreachable:
  - by orchestrator
  - and by _all of its replicas_
  and still prefer failover didn't take place?

FWIW on our system replicas take ~4-5 seconds to realize the master is dead. This is somewhat controllable with replication heartbeat.

serdarsahin:
Hi Shlomi,

We were running a number of tests so I didn't want to get back to you without finalising them.

> What, in your opinion, is a failover scenario?
I think Orchestrator has the best failure detection systems developed for MySQL environments so far. We really appreciate that. We also want to make sure that the failover won't be trigger by some intermittent issue on the AWS side. We know from the experience that it is possible to experience short network connection issues on the AWS side, so we would like to wait for 5/10 seconds before failing over to the slave.
> Are you comfortable with master being unreachable:
> by orchestrator
> and by all of its replicas
> and still prefer failover didn't take place?

If this happens only for 5 seconds for example, we would rather not to failover to the slave. 

> FWIW on our system replicas take ~4-5 seconds to realize the master is dead. This is somewhat controllable with replication heartbeat.

How can we control this with that parameter? Could you shortly guide us on that?

shlomi-noach:
> If this happens only for 5 seconds for example, we would rather not to failover to the slave.

I'll give this a thought but I can expect some problems with a "wait 20 seconds before taking action" design. I do not expect to implement this in the near future.

> How can we control this with that parameter? Could you shortly guide us on that?

```
set global slave_net_timeout = 4;
```
potentially followed by
```
stop slave;change master to master_heartbeat_period=2;start slave;
```
on replicas.

https://dev.mysql.com/doc/refman/5.7/en/replication-options-slave.html#sysvar_slave_net_timeout

serdarsahin:
Thanks Shlomi. We'll try this one out!

PR #372, opened 2018-01-15
raft: properly resolving recovery

shlomi-noach:
This PR fixes an issue where in a raft setup, a recovery done by the leader is propagated to all nodes, but is only marked as "successful" on the leader, while remains "failed" on the other nodes.


shlomi-noach:
Deployed and tested in prod (from before the PR was created)

PR #374, opened 2018-01-15
Stricter detection acknowledgement on recovery

shlomi-noach:
This PR tackles the case where a recovery acknowledges a detection, leading to an immediate detection followup.


shlomi-noach:
The problem this PR solves was introduced via https://github.com/github/orchestrator/commit/0419721dd2ec4f5b9e58708d7586baf64c979fc6

PR #383, opened 2018-01-21
Recovery: auditing write to KV

shlomi-noach:
Auditing: noting down KV write, including KV pair content.


shlomi-noach:
Works as expected.

PR #396, opened 2018-02-02
Auto Pseudo-GTID: detect via performance_schema

shlomi-noach:
Continued work on [auto Pseudo-GTID](https://github.com/github/orchestrator/blob/master/docs/configuration-discovery-pseudo-gtid.md#automated-pseudo-gtid-injection) (followup to https://github.com/github/orchestrator/pull/384).

This PR improves the heuristic by which `orchestrator` detects existence of Pseudo-GTID on a server. For MySQL >= `5.6`, `orchestrator` is now able to query `performance_schema.events_statements_current`, as in the following query:

```sql
select
  ifnull(
    min(
      timestampdiff(
        second, from_unixtime(
          conv(
            substring_index(
              substring_index(sql_text,':',-3),
              ':',1),
            16,10
          )
        ),
        now()
      )
    ) <= 14400,
    0
  ) as pseudo_gtid_is_found
from
  performance_schema.events_statements_current
where
  sql_text like 'drop view if exists `_pseudo_gtid_`%'
;
```

If possible, `orchestrator` runs this query. If query is successful and confirms the existence of a Pseudo-GTID injection, we can safely conclude that the instance does indeed run with Pseudo-GTID.
If found, the result is cached for `1hr` to avoid needless querying of `performance_schema`.

### DIFF
--- a/docs/configuration-discovery-pseudo-gtid.md
+++ b/docs/configuration-discovery-pseudo-gtid.md
@@ -33,6 +33,18 @@ Those statements will do nothing but will serve as magic markers in the binary l
 REVOKE DROP ON _pseudo_gtid_.* FROM 'orchestrator'@'orch_host';
 ```
 
+#### In versions > 3.0.6:
+
+It is also advisable that you grant:
+```sql
+GRANT SELECT ON `performance_schema`.`events_statements_current` TO 'orchestrator'@'orch_host';
+```
+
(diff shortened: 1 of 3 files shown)


shlomi-noach:
In production the query does not always identify the pseudo-GTID injection. In fact, it rarely does.

shlomi-noach:
Not going to pursue this.

PR #398, opened 2018-02-06
Raft snapshots: storing minimal instance

shlomi-noach:
Raft snapshots don't write down entire `Instance` contents. Reason: too much content, which is likely outdated after restart. Instead, they just give minimal context for `orchestrator` to pick up.

The instances context is relevant when the `orchestrator` node starting up has lost its data.

Up till now, the only context would be the instance key. `orchestrator` would then discover the instances, each by its key, and repopulate its backend.

However, this is a problem when an instance is out for the moment. It cannot be discovered (which is legit, things happen); but orchestrator also lost the identity of its master. Not good, because `orchestrator` needs the replication graph to better analyze errors.

This PR adds the `MasterKey` to the snapshot data. Instances could still be missing, but `orchestrator` is able to still maintain the correct graph.


shlomi-noach:
we also add `ClusterName`

PR #401, opened 2018-02-09
Support SSL replication in graceful master takeover

stankevich:
As summarized in https://github.com/github/orchestrator/issues/380, during a graceful master takeover (or co-master pairing), Orchestrator needs to issue a `CHANGE MASTER TO MASTER_SSL = 1` in addition to other `CHANGE MASTER TO ...` statements on the former master (now a replica) in order for it to start replicating from the new master.

This change allows Ochestrator detect whether `CHANGE MASTER TO MASTER_SSL = 1` is required based on `AllowTLS` value of the former replica (now master).

Based on [`ChangeMasterCredentials`](https://github.com/github/orchestrator/blob/638f67c188cae28fbce46a13a8d05df42fdf65d2/go/inst/instance_topology_dao.go#L547).
 
---


shlomi-noach:
This looks good. I'll send to testing and report.

shlomi-noach:
sorry for the delay, this will take a few more days.

Issue #403, opened 2018-02-09
Enhancement: Allow custom rendering or other hooks for the web interface


berlincount:
In some constellations, you might want to be able to customise the look and feel of the web interface.

It would be helpful to have some dedicated injection points for custom CSS and JavaScript.

shlomi-noach:
While I'm not opposed to the general idea of customizing JS & CSS, the following must be made clear:

- How will I support users who have set up custom CSS and JS? (I don't wish to)
- What will be my indication that they are indeed running custom CSS and JS? (To be able to tell them "unsupported")
- Documentation must be updated to reflect these changes.

Thank you!

berlincount:
Thanks for your consideration :)

- I'd make clear that customisations are optional and unsupported, but that endpoints / hooks can be added as needed, but that the API and data structures are subject to change at your leisure, and also largely undocumented. Anyone capable of using the Web Developer features of any current browser should have no trouble inspecting and adjusting as needed
- I'd make the corresponding files "config files" in the Debian package (so `dpkg` handles them correctly on updates) and use some XHR-style fetching in the JavaScript to verify original checksums (gathered during the build process) against the version currently running? If the checksum of the current `css/custom.css`/`js/custom.js` mismatched with the shipped one, add `-customized` to the version string .. and ask to reproduce any issues on a non-customised version when supported.
- I'd leave the documentation sparse and mention that there are some hooks and that (unstable) data structures can be used .. if needed.

Hope that helps?

shlomi-noach:
Thank you! a few more clarifications:

> I'd make the corresponding files "config files" in the Debian package (so dpkg handles them correctly on updates) and use some XHR-style fetching in the JavaScript to verify original checksums (gathered during the build process) against the version currently running?

  Do you know how to do this? Are you able to add this to the PR?

> I'd leave the documentation sparse and mention that there are some hooks and that (unstable) data structures can be used .. if needed.

  Heh, it was a polite way of asking you to add said documentation 😆

berlincount:
Heh, I was trying to sidestep it ;) I guess I might be able to do so, yes. Need to find some time under somewhere under the sofa or so ...

berlincount:
I'm considering rewriting it slightly to use https://learn.jquery.com/events/introduction-to-custom-events/ with event names like `orchestrator:preRenderCluster` and `orchestrator:postRenderCluster`, so we got a more standardised approach here (and don't need empty hook functions, plus a grep for available hooks becomes easy).

[...]

PR #404, opened 2018-02-09
Add custom CSS and JS Rendering Hook

berlincount:
Related issue: https://github.com/github/orchestrator/issues/403

This PR adds HTML code to load custom CSS and JavaScript code into the website.

It also introduce a single new JavaScript hook which is called after the Cluster was rendered, so further transmutation on it can be done (there are none in the supplied empty hook function).

The naming of the hook function might not be ideal, but it's clear what it's supposed to do.


shlomi-noach:
Please see comments on #403

berlincount:
Updated per comments :)

shlomi-noach (commented):
Looking very good! will send to testing env.

PR #408, opened 2018-02-13
Raft reverse proxy API calls to leader

shlomi-noach:
Closes https://github.com/github/orchestrator/issues/245

This PR introduces a reverse proxy for API calls such that:

- when running Raft
- when calling **selected** `/api/...` calls on a non-leader
- when the node is a valid follower (can see the leader)
- the node will reverse proxy the request to the leader

- excluded from reverse-proxy: queries like `status`, `raft-health`, `raft-state`, `lb-check` etc., which are node-specific.

This will allow querying any valid raft node as well as making changes, refactoring, etc.

Introduced: `/api/raft-health`: a health/readiness check to confirm the node is a healthy part of the raft group.


shlomi-noach:
- [x] documentation

shlomi-noach:
This works well.

PR #420, opened 2018-02-25
Maintainance lock cycle

shlomi-noach:
Solves https://github.com/github/orchestrator/issues/407

This PR introduces:

- `in-maintenance` command (and backing function) which tells if an instance is currently under maintenance.
- `match-below` command (Pseudo-GTID relocate) requires that the target instance is not already under maintenance. This breaks the loop depicted in https://github.com/github/orchestrator/issues/407 without imposing new locks.

cc @sjmudd


shlomi-noach:
Works well.

PR #436, opened 2018-03-08
Read topology instance: partial success, improve analysis

shlomi-noach:
The following analysis:

- `UnreachableMaster`
- `UnreachableIntermediateMaster`
- `UnreachableCoMaster`

will not fire if the server has been at least partially reachable (at the very least, a `select @@global.hostname ...` has succeeded).

These three analysis fire, today, whenever a single probing query fails. Sometimes queries fail, and detection gets noisy (though, fun to note, there is no actioanble detection noise, hence no false positive failovers caused).

Since the three above analysis are non-actionable, there is no risk in reducing their fire rate. They can tolerate a situation where the instance is partially detected.

An instance with only partial probing successful is still considered "invalid", appears _black_ on web UI etc., it's just that an `UnreachableMaster` detection (for example) will not fire.


shlomi-noach:
Works as expected

PR #437, opened 2018-03-11
Fixing infinite auto-reload mayhem

shlomi-noach:
A few pages on the web interface auto-reload after `60sec`.

If the browser is not quick enough to initiate the reload, the code introduces a new reload request, and a new one, and a new one... Either until the browser/network is quick enough or until the compueter's vent is screaming. 

This fixes the infinite reload loop by satisfying with a single reload request. If it fails, it fails. No further attempt is made if the first attempt is failed.


shlomi-noach:
Cherry picked into #436

shlomi-noach:
Works as expected

PR #442, opened 2018-03-15
Failure analysis contains command hint, advertised to hooks

### LINKED ISSUE #441: Expose commands as hook arguments
It would be great to be able to see if a promotion was the result of a graceful-master-takeover as opposed to an unplanned promotion, taking different actions in the hooks depending, like paging or no

Something like {{command}}?

shlomi-noach:
Closes #441 

This PR adds, where possible, the failover command as part of the analysis structure, thereafter passed to external hooks as `{command}` or `$ORC_COMMAND`.

These commands can be:

- `graceful-master-takeover`
- `force-master-failover`
- `force-master-takeover`
- or empty, when failover was not initiated by a specific command.


choadrocker:
This looks great! Thanks!

PR #443, opened 2018-03-15
graceful-takeover to attempt relocate-replicas

shlomi-noach:
`graceful-master-takover` takes a more permissive approach and now allows failing over when then master has multiple replicas, given that:

- The user specifies the particular rdesignated eplica they want to failover to
- `orchestrator` is able to replocate all other replicas below designated replica.


shlomi-noach:
TODO/check:

- [ ] API
- [ ] CLI
- [ ] `orchestrator-client`
- [ ] documentation

shlomi-noach:
Heh, this wasn't actually merged in the true sense of code. Gonna re-work on this in a different PR.

PR #447, opened 2018-03-19
Support for 'ack-all-recoveries'

shlomi-noach:
Supporting a way to acknowledge all recoveries, thereby releasing all blocks from a pending recovery (aborting anti-flapping block time).

New command is supported via CLI, API, and `orchestrator-client`


sanjida3:
Best

rudy-gao:
@shlomi-noach  Hi, when I use the following command, I find nothing output
```
orchestrator help ack-all-recoveries
ack-all-recoveries:

```
this may be in command_help.go not including CommandHelp.

PR #458, opened 2018-04-04
Fix go version reported too low, issue: 457

sjmudd:
See: https://github.com/github/orchestrator/issues/457

This fixes the reported problem with go1.10.


shlomi-noach (changes requested):
Unfortunately this won't catch a `go1.4.1` version.

PR #460, opened 2018-04-04
Fix behaviour related to connections with/without TLS

sjmudd:
### Description

See: https://github.com/github/orchestrator/issues/456

This fixes the issue where orchestrator might think the connection to a server needed TLS when it did not, usually caused by a server reconfiguration related to TLS usage.

inst.ForgetInstance() will now clean out `database_instance_tls` removing any reference to the instance if found.

The checks in `db.requiresTLS()` now identify if the stored TLS configuration requirement matches what is needed to connect to the server and will adjust it if wrong.

- [x] code is built via `./build.sh` - note: there's a pending patch to support go1.10
- [x] code is tested - a similar patch is running on my production system.


go/db/tls.go
```diff
@@ -54,42 +58,100 @@ func init() {
...
+		_, err = ExecOrchestrator(insertOrUpdateQuery, host, port, requiredWhenConnecting)
+		if err != nil {
+			log.Errore(err)
+			return requiredWhenConnecting != 0
+		}
+
+		writeInstanceTLSCounter.Inc(1)
+	}
+
+	return requiredWhenConnecting != 0
```
shlomi-noach, line 146: What's the reason for `requiredWhenConnecting` being an `int` rather than `bool`?

shlomi-noach (commented):
Thank you for the PR. One question inline.

Also, I assume this is running on your production setup?

go/db/tls.go
sjmudd: The column in the side-table `orchestrator.database_instance_tls` which records the TLS setting is a `tinyint unsigned` (setup that way previously) so I'm not bothering to change it but to use the value retrieved etc.

```
root@my-orchestrator-backend [orchestrator]> show create table database_instance_tls\G
*************************** 1. row ***************************
       Table: database_instance_tls
Create Table: CREATE TABLE `database_instance_tls` (
  `hostname` varchar(128) NOT NULL,
  `port` smallint(5) unsigned NOT NULL,
  `required` tinyint(3) unsigned NOT NULL DEFAULT '0',
  PRIMARY KEY (`hostname`,`port`)
) ENGINE=InnoDB DEFAULT CHARSET=ascii
1 row in set (0.00 sec)
```

shlomi-noach: You can have `requiredWhenConnecting` as a `bool` and it will transparently translate to/from an `INT` in the database.

shlomi-noach: @sjmudd would you like to change to `bool`?

PR #465, opened 2018-04-12
limit more precisely when special binlog filtering is used

sjmudd:
The routines for doing special binlog filtering are required when the 2 servers run a mix of 5.6 and 5.7 due to some events which are in the 5.7 binlogs but not the 5.6 binlogs.

This patch tightens up the checks by doing the following:
* the code would call `GetInstanceMaster()` which tries to talk to the instance's master but this code may be called when the intermediate master is dead so will fail. Instead call `ReadInstance()` which uses the orchestrator backend to get version information of the instance's master.

- [x] code running on my own system


shlomi-noach (approved):
Makes perfect sense

PR #470, opened 2018-04-15
graceful takeover: support multiple replicas and explicit candidate

shlomi-noach:
Mostly a resubmission of https://github.com/github/orchestrator/pull/443

`graceful-master-takeover` supports multiple master direct replicas, when the designated new master is explicitly indicated.

TODO/check:

- [x] API
- [x] CLI
- [x] `orchestrator-client`
- [x] documentation


shlomi-noach:
Woot! Looking real good! Now working on docs.

shlomi-noach:
- [x] ok to ignore downtimed servers in case of error

PR #486, opened 2018-05-06
Fix detection of Go 1.10.x

dveeden:
Fedora 28 ships with go 1.10.x which build.sh detects as too old, while it isn't too old.
This PR fixes that by a more restrictive regex.

```
$ go version
go version go1.10.1 linux/amd64
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -56,7 +56,7 @@ function precheck() {
     ok=1
   fi
 
-  if [[ $(go version | egrep "go1[.][01234]") ]]; then
+  if [[ $(go version | egrep "go1[.][01234][.]") ]]; then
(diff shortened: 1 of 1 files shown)


dveeden:
Looks like this is a duplicate of #458

shlomi-noach (changes requested):
We would need a better method, because the version could be `1.8` or `1.9` (no trailing `.`).

I have something else in mind, let me push a change.

shlomi-noach (approved):

shlomi-noach:
I pushed a change to your PR, and it should be good now: explicitly allowing `go1.8`, `go1.9`, `go1.10`, and anything else is disallowed.

PR #489, opened 2018-05-07
support for event_scheduler

shlomi-noach:
Closes https://github.com/github/orchestrator/issues/478

This PR will:

- [x] support probing & auditing of `@@global.event_scheduler`.

- [x] Upon master recovery, and if the dead master had `event_scheduler=ON`, also turn `event_scheduler=ON` on promoted server.


shlomi-noach:
Copying comment by @joegrasse: https://github.com/github/orchestrator/issues/478#issuecomment-387386390

@shlomi-noach I see where you are turning on and off the event scheduler, but I don't see where you are actually doing anything with the events that are actually being ran.

Usually you would have some events that are 'ENABLED' on the current master and 'SLAVESIDE_DISABLED' on the slaves. When the master is changed, the events on the new master need to be enabled.

shlomi-noach:
@joegrasse ah yes. This is quite unfortunate. It means `orchestrator` would need to have the `EVENT` privilege over the entire server (or for all possible schemas). I'm not enthusiast about making this change, since I like to keep `orchestrator`'s privileges limited, and not have access to production data (it already has some access to binlogs).

joegrasse:
Yes, however, this is opt-in. They would only need to give ```orchestrator``` permission if they are using the event scheduler and they want ```orchestrator``` to manage it on failover.

shlomi-noach:
@joegrasse that is doable, but incomplete. What if they only grant `orchestrator` privileges on some schemas and not on all? I can provide with some visibility into what `orchestrator` does upon failover, but I fear I'll get the backfire on this issue.

I'll look into it though.

shlomi-noach:
@joegrasse that is doable, but incomplete. What if they only grant `orchestrator` privileges on some schemas and not on all? I can provide with some visibility into what `orchestrator` does upon failover, but I fear I'll get the backfire on this issue.

I'll look into it though.

PR #498, opened 2018-05-11
Ensure that KV Stores are always initialized before they're accessed

TylerLubeck:

Related issue: https://github.com/github/orchestrator/issues/497

### Description

Ensure that KV Stores are initialized before they're accessed

- [] code is tested via `go test ./go/...`

/cc @choadrocker


TylerLubeck:
Fixed in #499

PR #512, opened 2018-05-20
fixed typo in command help for "clusters"

fipar:
Related issue: https://github.com/github/orchestrator/issues/511

### Description

This PR fixes a typo in the example for the 'clusters' command help. 

- [ x] contributed code is using same conventions as original code

I did not check the last three checkboxes but the change is just swapping the order of two letters in 'orchestrator' on a help string, so the format is the same, and this shouldn't impact the build process or break any tests.


shlomi-noach (approved):
thank you!

PR #521, opened 2018-05-31
Supporting `can-replicate-from` command

shlomi-noach:
Example: 
```shell
$ orchestrator-client -c can-replicate-from -i some.instance:3306 -d from.other.instance:3306
true
```

Prints out `true|false` if 1st instance can replicate from 2nd instance according to replication rules (binlog format, versions,binary logs exist etc.), same way `orchestrator` validates internally.
Either `true` or `false`, exit code is `0`. Non-zero code is reserved to errors (e.g. cannot figure out which server is involved).

### DIFF
--- a/go/app/cli.go
+++ b/go/app/cli.go
@@ -749,26 +749,25 @@ func Cli(command string, strict bool, instance string, destination string, owner
 			if instanceKey == nil {
 				log.Fatalf("Unresolved instance")
 			}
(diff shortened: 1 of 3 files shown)


shlomi-noach:
Change of output:

prints destination ("2nd"/"parent") instance key if true, empty if false.

shlomi-noach:
Added `is-replicating` command

PR #542, opened 2018-06-28
Test for "Access denied for user" to trigger TLS

### LINKED ISSUE #541: Accounts requiring an encrypted connection fail to be discovered
When `MySQLTopologyUseMutualTLS` is `false`, accounts that require an encrypted connection fail during discovery.

```
$ orchestrator-client -c all-instances | wc -l
0

$ mysql -h 192.168.56.20 -Bse "select user, ssl_type from mysql.user where user = 'orchestrator'"
orchestrator    ANY

$ egrep MySQLTopologyUse\(Mutual\|Mixed\)TLS /etc/orchestrator.conf.json
  "MySQLTopologyUseMutualTLS": false,

$ orchestrator-client -c discover -i 192.168.56.20:3306
Error 1045: Access denied for user orchestrator@monitor (using password: YES)
```

Checking for `Access denied for user` in `requiresTLS` triggers the switch to TLS mode
and discovery is then possible.

cezmunsta:
To fix https://github.com/github/orchestrator/issues/541
```
$ ln -f /usr/local/orchestrator/orchestrator.requireTLSfix /usr/local/orchestrator/orchestrator; systemctl restart orchestrator.service

$ orchestrator-client -c discover -i 192.168.56.20:3306
192.168.56.20:3306

$ orchestrator-client -c all-instances | wc -l
1

$ orchestrator-client -c all-instances | wc -l
2

$ orchestrator-client -c topology -a db1
192.168.56.20:3306   [0s,ok,5.7.21-21-log,rw,ROW,>>]
+ 192.168.56.21:3306 [0s,ok,5.7.21-21-log,ro,ROW,>>]
```

### DIFF
--- a/go/db/tls.go
+++ b/go/db/tls.go
@@ -33,6 +33,7 @@ import (
 )
 
 const Error3159 = "Error 3159:"
+const Error1045 = "Access denied for user"
 
 // Track if a TLS has already been configured for topology
 var topologyTLSConfigured bool = false
@@ -64,7 +65,7 @@ func requiresTLS(host string, port int, mysql_uri string) bool {
 
 	required := false
 	db, _, _ := sqlutils.GetDB(mysql_uri)
-	if err := db.Ping(); err != nil && strings.Contains(err.Error(), Error3159) {
+	if err := db.Ping(); err != nil && (strings.Contains(err.Error(), Error3159) || strings.Contains(err.Error(), Error1045)) {
 		required = true
 	}
 


shlomi-noach (approved):
Thank you!

shlomi-noach (commented):
Question: does `Access denied for user` necessarily mean a TLS setup? Will the same message not fire for a "normal" problem with credentials?

cezmunsta:
@shlomi-noach the "access denied for user" could indeed appear for normal failure of credentials. However, because `requiresTLS` is only called [if `MySQLTopologyUseMixedTLS` is enabled](https://github.com/github/orchestrator/blob/master/go/db/db.go#L91) this seemed to be an acceptable tradeoff as TLS is "expected". If you feel that there is a better way to resolve this issue then please apply that instead.

shlomi-noach:
@cezmunsta thank you. This seems reasonable.

PR #545, opened 2018-07-01
Reduce container image size by using multi-stage builds

### LINKED ISSUE #66: Smaller Docker Image
Hi all,

I've just created a [PR](https://github.com/github/orchestrator/pull/65) to update the Dockerfile to produce a smaller base image.
Sorry for not creating an issue beforehand, I've just read that's how it should be done.
Anyway, hope it's useful to you.

Thanks for the amazing tool that Orchestrator is!

rhoml:
fixes: #66 

By using multi stage builds we can reduce the image size by 50%.

Regular build 
```
<none>              <none>              cec3efb55b6d        8 minutes ago       60.6MB
```

After multi stage build
```
<none>              <none>              4ab2bafd0fc9        8 minutes ago       30.1MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 FROM alpine
 
-EXPOSE 3000
-
 ENV GOPATH=/tmp/go
 
 RUN set -ex \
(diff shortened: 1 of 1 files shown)


shlomi-noach (approved):
Nice!
I've never seen the use of `FROM` twice in the same Dockerfile.

I see this is new in https://docs.docker.com/v17.09/engine/userguide/eng-image/multistage-build/#use-multi-stage-builds. Perhaps we should add comment specifying you'd need a `17.09` or later?

Unrelated, the existing `Dockerfile` is about to get a facelift. It is plain wrong: it pulls `orchestrator` from `origin/master` as opposed to running the very branch you're on. TODO.

PR #553, opened 2018-07-04
New variable: disable AutoPseudoGTID where GTID is on

shlomi-noach:
A new config variable, `AutoPseudoGTIDDisabledOnGTID`, introduces a new behavior:

When `AutoPseudoGTIDDisabledOnGTID` is `true`, and a master has `gtid_mode = ON`, then `orchestrator` avoids injecting Pseudo GTID entries onto that master.

There is subtlety here: the fact that a master has `gtid_mode=ON` _does not mean_ the replicas are using GTID. They could use `master_auto_position: 0` and just use file:pos. `orchestrator` does not check all of the master's replicas (at this time, maybe this will change) hence the need to a new variable.


shlomi-noach:
On hold. I don't think I will merge this.

shlomi-noach:
Closing this PR.

PR #555, opened 2018-07-08
reduce logs

shlomi-noach:
Logs are overwhelming. This PR aggressively reduces log count by cutting down frequent and noisy log entries.


shlomi-noach:
cc @sjmudd

PR #570, opened 2018-07-31
Stop graceful master takeover immediately when attempting to promote an ignored host

dougfales:
Related issue: https://github.com/github/orchestrator/issues/569

### Description

During a graceful master takeover, this simply checks the `IsBannedFromBeingCandidateReplica` function for the newly designated master before doing anything else. By checking this early and failing early, we avoid making any changes to the topology for a promotion we know will fail at the outset. 

(Thanks @zmoazeni for the idea!)

### DIFF
--- a/go/logic/topology_recovery.go
+++ b/go/logic/topology_recovery.go
@@ -1683,10 +1683,11 @@ func GracefulMasterTakeover(clusterName string, designatedKey *inst.InstanceKey)
 		if designatedInstance == nil {
 			return nil, nil, fmt.Errorf("GracefulMasterTakeover: indicated designated instance %+v must be directly replicating from the master %+v", *designatedKey, clusterMaster.Key)
 		}
-		if inst.IsBannedFromBeingCandidateReplica(designatedInstance) {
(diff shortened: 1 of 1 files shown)


shlomi-noach (changes requested):
Thank you, looking good! The same check should also apply for this case: https://github.com/dougfales/orchestrator/blob/4b1833b73f8c6c110464a07fb585a875707d8e85/go/logic/topology_recovery.go#L1670

i.e. the check should be made whetehr the designated instance is explicitly indicated (the case you handle) or implied (master has a single direct replica, hence that's the designated instance).

dougfales:
Great point! I pulled the check outside of the `if designatedKey == nil` conditional so that it will be done with the `designatedInstance`, regardless of whether that instance was indicated by the user or selected by orchestrator. 

I set up a local test env with three replicas, banning one of them via `PromotionIgnoreHostnameFilters`, and I verified that the error is reported immediately in both cases:
1. When I pass the banned replica explicitly.
2. When I make the banned replica the only descendant of the current master and therefore orchestrator tries to choose it as the `designatedInstance`.

Please let me know if there's anything else I should add or clean up here. Thanks!

shlomi-noach (approved):

dougfales:
Thanks Shlomi!

PR #577, opened 2018-08-15
Add MySQLConnectionLifetime

dbnski:

Related issue: https://github.com/github/orchestrator/issues/559

### Description

This PR adds a new configuration variable MySQLConnectionLifetime to define the maximum lifetime for persistent database connections.


go/db/db.go
```diff
@@ -113,7 +117,13 @@ func openOrchestratorMySQLGeneric() (db *sql.DB, fromCache bool, err error) {
 	if config.Config.MySQLOrchestratorUseMutualTLS {
 		uri, _ = SetupMySQLOrchestratorTLS(uri)
 	}
-	return sqlutils.GetDB(uri)
+	if db, fromCache, err = sqlutils.GetDB(uri); err != nil {
+		return nil, fromCache, err
+	}
+	if config.Config.MySQLConnectionLifetime > 0 {
+		db.SetConnMaxLifetime(time.Duration(config.Config.MySQLConnectionLifetime) * time.Second)
+	}
```
shlomi-noach: This is not needed. If the `db` is returned from cache then `SetConnMaxLifetime` has already been applied.

go/config/config.go
```diff
@@ -129,6 +129,7 @@ type Configuration struct {
 	MySQLOrchestratorReadTimeoutSeconds        int      // Number of seconds before backend mysql read operation is aborted (driver-side)
 	MySQLDiscoveryReadTimeoutSeconds           int      // Number of seconds before topology mysql read operation is aborted (driver-side). Used for discovery queries.
 	MySQLTopologyReadTimeoutSeconds            int      // Number of seconds before topology mysql read operation is aborted (driver-side). Used for all but discovery queries.
+	MySQLConnectionLifetime                    int      // Number of seconds the mysql driver will keep database connection alive before recycling it
```
shlomi-noach: Can you please rename the variable to `MySQLConnectionLifetimeSeconds`?

shlomi-noach (changes requested):
Thank you! Please see comments.

go/config/config.go
dbnski: Done.

go/db/db.go
almeida-pythian: I'm not sure if this was intended @maciej-dobrzanski and @shlomi-noach but I noticed when trying to compile these changes I ran into an error due to a missing function. I eventually found this section on db.go file, these last few lines are missing:

```console
// ReadTimeNow reads and returns the current timestamp as string. This is an unfortunate workaround
// to support both MySQL and SQLite in all possible timezones. SQLite only speaks UTC where MySQL has
// timezone support. By reading the time as string we get the database's de-facto notion of the time,
// which we can then feed back to it.
func ReadTimeNow() (timeNow string, err error) {
	err = QueryOrchestrator(`select now() as time_now`, nil, func(m sqlutils.RowMap) error {
		timeNow = m.GetString("time_now")
		return nil
	})
	return timeNow, err
}
```

This is the error I received when building: (just the snippet below)

```console
+ go build -i -o /root/orchestrator/bin/orchestrator -ldflags '-X main.AppVersion=eb7a3b642f6e0aa83a4257ae62441571cc14a292 -X main.BuildDescribe=v3.0.12-12-geb7a3b64-dirty' ./go/cmd/orchestrator/main.go
# github.com/github/orchestrator/go/inst
go/inst/candidate_database_instance.go:42:31: undefined: db.ReadTimeNow
```


On a positive note, once I added the missing function to the file, I was able to compile these changes without any issues. I then added the following config parameters to my orchestrator config file and I am able to verify (so far), the "invalid connection" error has disappeared (we were experiencing this quite often, before we applied this latest suggested changes, thank you for finding this):

```console
"MySQLConnectTimeoutSeconds": 10, 
"MySQLOrchestratorReadTimeoutSeconds": 10, 
"MySQLDiscoveryReadTimeoutSeconds": 10, 
"MySQLTopologyReadTimeoutSeconds": 10, 
"MySQLConnectionLifetimeSeconds": 600, 
```

shlomi-noach: `db.ReadTimeNow` is a functional that was added recently. I suspect the branch was not updated properly. I just updated it.

shlomi-noach: Please remove as per comment.

almeida-pythian: Thanks @shlomi-noach :1st_place_medal:

shlomi-noach:
Thank you all for your patience while I look into the changes.

go/db/db.go
dbnski: Reverted to the original code here, as per your request

shlomi-noach (approved):
Thank you

PR #581, opened 2018-08-19
orchestrator-client basic auth fix

shlomi-noach:
Fixes https://github.com/github/orchestrator/issues/580

Correct basic auth flags passed to `curl` (with default `:` user:password).


shlomi-noach:
@timhughes @ cloudufull would you like to test this? (looks good on my side)

PR #587, opened 2018-08-21
Initialize CandidateDatabaseInstance LastUpdated field to UTC time

nickdelnano:
Related issue: https://github.com/github/orchestrator/issues/585

### Description

Initializes CandidateDatabaseInstance LastUpdated field to time.Now().UTC() instead of time.Now(). 

time.Now() outputs the system time with an offset relative to UTC -- on a system set in PDT, sample output is `2018-08-20 20:09:12.395052589-07:00`. SQLite only reads timestamps in the format YYYY-MM-DD HH:MM:SS.SSS", and therefore comparisons against this field are incorrect, namely when determining whether a promotion rule is up for expiration. 

The SQLite datetime('now') function is used for generating a value to compare against the last_updated column of the candidate_database_instance table, and this function implicitly uses UTC. I propose initializing the CandidateDatabaseInstance LastUpdated field to time.Now().UTC() in order for the comparison in the [ExpireCandidateInstances](https://github.com/github/orchestrator/blob/master/go/inst/instance_dao.go#L2572) function to be consistent despite the system time setting.

> In case this PR introduced Go code changes:

- [x] contributed code is using same conventions as original code
- [x] code is formatted via `gofmt` (please avoid `goimports`)
- [x] code is built via `./build.sh`
- [x] code is tested via `go test ./go/...`

Issue #585, opened 2018-08-21
Promotion rule expiration is dependent on system time


nickdelnano:
Orchestrator [uses the system time when setting the LastUpdated field of a new CandidateDatabaseInstance](https://github.com/github/orchestrator/blob/master/go/inst/candidate_database_instance.go#L37). 
This stores the system time with an offset to UTC: (I've set the system time to PDT)
```
/usr/local/orchestrator # sqlite3 orchestrator.sqlite3 "select * from candidate_database_instance;"
8fa4a00f0156|3306|2018-08-20 20:09:12.395052589-07:00|1|prefer
```

When [ExpireCandidateInstances](https://github.com/github/orchestrator/blob/master/go/inst/instance_dao.go#L2572) looks to expire entries, it uses the query:
```
delete from candidate_database_instance where last_suggested < NOW() - INTERVAL ? MINUTE
```
where the NOW() function is converted to datetime('now") for sqlite, which implicitly uses UTC. Sqlite does not parse the offset listed in the last_suggested column, only the text in the format "YYYY-MM-DD HH:MM:SS.SSS" (see section 2.2 of sqlite types manual https://www.sqlite.org/datatype3.html).

This causes promotion rules to be forgotten in near instant time on a machine whose time is set 1 hour or more behind UTC. I believe this also effects MySQL users the same.

I've reproduced this with the project's docker setup; the reproduced output and the steps to reproduce are listed in https://github.com/github/orchestrator/pull/584.

shlomi-noach:
Thank you for this analysis. Before looking into it, do you happen to have a suggestion for remediation?

shlomi-noach:
Closed by #591

PR #591, opened 2018-08-26
 Candidate databse instance: database native time

shlomi-noach:
This PR resubmits https://github.com/github/orchestrator/pull/587 with only the actual code change. See discussion on #587 and #585 

cc @ndelnano

### DIFF
--- a/go/app/cli.go
+++ b/go/app/cli.go
@@ -1456,7 +1456,7 @@ func Cli(command string, strict bool, instance string, destination string, owner
 			if err != nil {
 				log.Fatale(err)
 			}
-			err = inst.RegisterCandidateInstance(inst.NewCandidateDatabaseInstance(instanceKey, promotionRule))
+			err = inst.RegisterCandidateInstance(inst.NewCandidateDatabaseInstance(instanceKey, promotionRule).WithCurrentTime())
 			if err != nil {
 				log.Fatale(err)
 			}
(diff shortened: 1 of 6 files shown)


nickdelnano:
thank you @shlomi-noach !

nickdelnano:
Is a date for the next release planned? The dates of the last several releases does not look like there is a fixed schedule.

shlomi-noach:
You are correct, there is no fixed schedule. But a kind reminder helps!

shlomi-noach:
While this may work well for `sqlite`, this breaks `MySQL` for servers with non-UTC timezone. Looking into.

shlomi-noach:
We need to juggle between three different scenarios:

- MySQL backend (has timezone notion)
- Sqlite (does not have timezone notion)
- `raft` messages (`register-candidate` can apply by reading historical `raft` event, the value of `now()` becomes meaningless)

shlomi-noach:
@ndelnano I ended up solving the problem in a different way: reading the timestamp as a string from the database, then feeding it back.
Would you care to test this PR works for you? Works for me, on `sqlite` and `MySQL` on different timezone boxes.

nickdelnano:
@shlomi-noach I have tested and it does indeed work for me. Thanks again!

shlomi-noach:
OK good, this has been deployed to staging and in production for a while. I'll let our daily tests run and then merge.

PR #603, opened 2018-09-04
Supporting ReplicationCredentialsQuery config

### LINKED ISSUE #323: missing replication username/password error when changing from master-slave to master-comaster

shlomi-noach:
Fixes #323 

This PR introduces the new configuration variable `ReplicationCredentialsQuery`: 

`
// custom query to get replication credentials. Must return a single row, with two text columns: 1st is username, 2nd is password. This is optional, and can be used by orchestrator to configure replication after master takeover or setup of co-masters. You need to ensure the orchestrator user has the privileges to run this query
`

The query can allow `orchestrator` to get replication username + password from some metadata table in an instance. If the user wishes to use such query, the user must generate the data on the instances, as well as make sure `orchestrator` has the credentials to run the query.

`orchestrator` would use the credentials in these operations:

- Setting up co-masters
- Graceful master takeover

This PR replaces #582 

cc @cezmunsta @igroene


cezmunsta:
LGTM :+1:

PR #607, opened 2018-09-06
GTID errant transaction detection

shlomi-noach:
In this PR `orchestrator` auto-analyzes errant transactions for GTID-based replicas.

At this time it does nothing about it, merely presents it in the web UI. 

### Heuristic for avoiding false positives

`orchestrator` polls different servers at different times. It may be that it first polled the master and then its replica. In such case, the replica will _naturally_ be seen to have more entries than the master.
If the replica also happens to have real errant transactions, how can we tell apart the two?

`orchestrator` ignores the GTID set that relates to the replica's immediate master UUID. At this time `orchestrator` does not further ignore upstream UUIDs, and this is TBD.

This PR is to be considered as POC.

screen shot 2018-09-06 at 15 32 50

In future PRs `orchestrator` will allow the user to fix errant transactions across the cluster, in a click of a button, or via command line. To be carefully thought of.


shlomi-noach:
Right now build fails on `go fmt`, unrelated to anything, other that my upgrading to `go1.11` on my local box. Will follow up on that.

shlomi-noach:
Errant GTIDs now also show up:

- As a structure warning in `replication-analysis`
  - But not if `promotion-rule` is _downtimed_
- Corollary as a structure warning in Web interface cluster page and cluster analysis
- A `problem` in `/api/problems`, and corollary in problems sidebar
- A red Globe-icon indicator, see screenshot

screen shot 2018-09-08 at 13 43 57

PR #618, opened 2018-09-12
Fail instead of Error if unable to setup DB

bbuchalter:
Closes https://github.com/github/orchestrator/issues/609

### DIFF
--- a/go/db/db.go
+++ b/go/db/db.go
@@ -141,7 +141,7 @@ func OpenOrchestrator() (db *sql.DB, err error) {
 			// first time ever we talk to MySQL
 			query := fmt.Sprintf("create database if not exists %s", config.Config.MySQLOrchestratorDatabase)
 			if _, err := db.Exec(query); err != nil {
-				return db, log.Errore(err)
+				return db, log.Fatale(err)
 			}
 		}
 		db, fromCache, err = sqlutils.GetDB(getMySQLURI())
--- a/tests/integration/test.sh
+++ b/tests/integration/test.sh
@@ -149,6 +149,7 @@ deploy_internal_db() {
   bash $exec_command_file 1> $test_outfile 2> $test_logfile
   if [ $? -ne 0 ] ; then
     echo "ERROR deploy internal db failed"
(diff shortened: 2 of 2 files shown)


shlomi-noach (approved):

shlomi-noach:
Seems like CI consistently fails at the same spot. Possibly due to this very change.

bbuchalter:
I've added a `cat $test_logfile` that's found [elsewhere in `tests/integration/test.sh`](https://github.com/github/orchestrator/blob/master/tests/integration/test.sh#L93). Let's see what CI tells us.

bbuchalter:
Hmmm, well, I think we need the change I added, but the fail is not coming from the integration tests, it's the units: `2018-09-13 17:33:58 FATAL Error 1045: Access denied for user ''@'localhost' (using password: NO)` 🤔

bbuchalter:
I've found at least one unit test that is trying to connect to the database:

https://github.com/github/orchestrator/blob/f7533848fcd6cae6506ed8fc2d79fa7c751389c3/go/inst/cluster_test.go#L59

https://github.com/github/orchestrator/blob/622e979722bf433232ae267727933ba055cdbfed/go/inst/cluster.go#L58

https://github.com/github/orchestrator/blob/41d02cdd021875582ca5c7867f14d895686ea58d/go/inst/resolve_dao.go#L371

https://github.com/github/orchestrator/blob/c2cfa3e2d3cd5853b2a5ce01ce49a591190f905c/go/db/db.go#L366

Which hits this change: https://github.com/github/orchestrator/blob/c2cfa3e2d3cd5853b2a5ce01ce49a591190f905c/go/db/db.go#L148

Not sure what the right course of action with this is...

bbuchalter:
Upon commenting out the offending call here:
https://github.com/github/orchestrator/blob/622e979722bf433232ae267727933ba055cdbfed/go/inst/cluster.go#L58

I was able to get the remainder of the unit tests to pass successfully: this is the only place where the database must be used for the unit tests.

PR #621, opened 2018-09-13
checks to zk key and updates or creates depending on if the key exists

timhughes:

Related issue: https://github.com/github/orchestrator/issues/619

### Description
Fixes an issue where zk keys are not updated because they already exist


timhughes:
fixed by https://github.com/github/orchestrator/pull/620

shlomi-noach (commented):
:+1: Thank you for submitting a fix. Do drop me a note if you intend to fix something yourself so that we avoid working on the same thing. Here, of course, the fix was near immediate and nearly no time spent.

PR #623, opened 2018-09-16
Formalizing InstanceKey, much refactoring

shlomi-noach:
Much around `InstanceKey` has been simplified, revised and refactored.

There should be no behavioral changes.

There is initial support for `ipv6` but the way is still long.


shlomi-noach:
Fixes #658

PR #624, opened 2018-09-17
Fix non-zero exit code for redeploy-internal-db

bbuchalter:
Closes https://github.com/github/orchestrator/issues/609

This ensures we get the expected non-zero exit code and print the
integration test log files which should contain the error messages.


shlomi-noach (approved):
Good solution!

shlomi-noach:
deploys well

PR #627, opened 2018-09-18
Decouple integration test runs from Travis

### LINKED ISSUE #625: Decouple integration test suite from Travis and a single orchestrator configuration
## Issue
There is currently a significant coupling to the Travis CI environment and the ability to run integration tests. This inhibits new developers who want to contribute new code and tests (https://github.com/github/orchestrator/issues/611). 

Additionally, in it's current state, the integration test suite can only be run against a single Orchestrator configuration, [specifically configured](https://github.com/github/orchestrator/blob/master/tests/integration/test.sh#L161) for [Travis CI](https://github.com/github/orchestrator/blob/622e979722bf433232ae267727933ba055cdbfed/.travis.yml#L10). This prevents new developers from adding new integration tests for different configurations or topologies.

## Proposed Solution
A naive approach to this issue would be to support multiple directories of integration tests. For example:
```
tests
└── integration
    ├── mysql_56_backend_comaster
    ├── mysql_57_backend_comaster
    └── sqlite_backend_comaster
```

An integration test runner could enumerate over these directories calling `test.sh` on each, assuming it will:
* do all the work of setting up a backend and topology (perhaps using [dbdeployer](https://github.com/github/orchestrator/issues/601)), 
* running the tests
* returning a non-zero exit code upon failure

This approach has several drawbacks:
* A directory name does not convey a meaningful description of the configuration/topology under test.
* Integration test suite times may balloon as more are added.
* The existing integration test suite would be duplicated across the SQLite and MySQL configurations.

This approach has one key benefits: **it's incredibly easy to add and run a new integration test suite**:
* Copy an existing integration test directory
* Modify `orchestrator.conf.json`
* Modify the `test.sh` to bootstrap the desired environment (different backends, topologies, etc...)
* Add/remove/modify the test assertions as needed

## Discussion and Next Steps
The configuration of Orchestrator, the topology under test and its expected behavior can be closely coupled. Supporting integration tests for more than one configuration and topology is important for users of Orchestrator to have high confidence that their use cases are supported upstream.

> A directory name does not convey a meaningful description of the configuration/topology under test.

This can be mitigated by having a README in each integration test directory discussing the particular configuration/topologies/tests.

 >  Integration test suite times may balloon as more are added.

The isolation between each integration test directory makes parallelism an easily solution to this problem (should it arise).

> * The existing integration test suite would be duplicated across the SQLite and MySQL configurations.

This is the most problematic concern for me, and one in which I don't see an obvious solution to. **I'd like to try and implement a proof of concept of my proposal and then see what steps can be taken to mitigate this central issue.**

bbuchalter:
Closes https://github.com/github/orchestrator/issues/625

By using dbdeployer, we can allow the MySQL backend version of the integration
test suite to be run anywhere.

You can view the improved version of the developer docs here: https://github.com/bbuchalter/orchestrator/blob/integration_tests_anywhere/docs/developers.md#running-tests

The `dbdeployer/setup.sh` script is flexible enough to also support creating a sandbox for local development as well. That will happen in a future PR, which should allow https://github.com/github/orchestrator/issues/611 to be closed.


shlomi-noach:
@bbuchalter I'm guessing you closed this due to inactivity on my side. I'm sorry for the long stall, to be honest this got buried under multiple other things. I'd still like to work on this.

PR #632, opened 2018-09-23
Fix to analysis GROUP BY

shlomi-noach:
Context: https://github.com/github/orchestrator/issues/631

This PR fixes the analysis query, that has an invalid `GROUP BY` term. While the fix was easy, I'm puzzled at how this went under the radar with CI, and in particular with SQLite. I would have assumed it would have been caught earlier. I'll look into that.

cc @sharad-jha


shlomi-noach:
Turns out SQLite, like MySQL without `ONLY_FULL_GROUP_BY`, allows invalid `GROUP BY` statements.
Work in progress to run `ONLY_FULL_GROUP_BY` `sql_mode` on CI MySQL.

PR #635, opened 2018-09-27
Include stdout/stderr in CommandRun's returned error

bbuchalter:
Closes https://github.com/github/orchestrator/issues/634

This change:
* Combines the output of STDOUT and STDERR in the CommandRun log messages. I did not see much benefit of being able to identify the source of the output.
* Includes the combined output in the returned error which improves the reporting back to the audit recovery log.
* Adds test coverage for the returned error when using environment variables.

cc: @kirinrastogi @insom @fjordan


shlomi-noach (approved):
Thank you for adding a test.

PR #639, opened 2018-10-01
Add support for go1.11

sjmudd:
Current code does not support the latest go1.11. This patch allows orchestrator to run with go 1.11.

`build.sh` now runs to completion happily on go 1.11.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -56,8 +56,8 @@ function precheck() {
     ok=1
   fi
 
-  if ! go version | egrep -q 'go(1\.(8|9|1[0]))' ; then
-    echo "go version must be between 1.8 and 1.10"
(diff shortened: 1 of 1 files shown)


shlomi-noach:
There is a single issue with `1.11` which is that `go fmt` produces slightly different output (different tabulation).
This breaks PRs submitted with `go < 1.11`, as our CI tests verify `go fmt` is applied.

Eventually this will be resolved when all users have upgraded to `1.11`, but in the meantime this is a breaking change. What do you think?

bbuchalter:
FWIW, it looks like golang is having problems on macos right now: https://github.com/golang/go/issues/25908. The fix is also having trouble being backported to 1.9.

New members of my team cannot use `script/build` with the state of things as they are right now.

shlomi-noach:
redundant; we're on go1.12 now

PR #646, opened 2018-10-05
Add global recoveries button

mateusduboli:
Related issue: https://github.com/github/orchestrator/issues/645

### Description
Add a button on the interface to show and change the global-recovery status:

Example

- [X] code is formatted via `gofmt` (please avoid `goimports`)
- [X] code is built via `./build.sh`
- [X] code is tested via `go test ./go/...`

### DIFF
--- a/resources/public/js/orchestrator.js
+++ b/resources/public/js/orchestrator.js
@@ -967,6 +967,35 @@ function getParameterByName(name) {
 }
 
 
+function renderGlobalRecoveriesButton(isGlobalRecoveriesEnabled) {
+  var iconContainer = $("#global-recoveries-icon > span");
+  if (isGlobalRecoveriesEnabled) {
+    iconContainer
+      .prop("title", "Global Recoveries Enabled")
+      .addClass("glyphicon-heart")
+      .removeClass("hidden")
+      .click(function(event) {
+        bootbox.confirm("<h3>Global Recoveries</h3>Are you sure you want to <strong>disable</strong> global recoveries?", function(confirm) {
+          if (confirm) {
+            apiCommand("/api/disable-global-recoveries");
+          }
+        })
+      });
+  } else {
+    iconContainer
+      .prop("title", "Global Recoveries Disabled")
+      .addClass("glyphicon-heart-empty")
+      .removeClass("hidden")
+      .click(function(event) {
+        bootbox.confirm("<h3>Global Recoveries</h3>Are you sure you want to enable global recoveries?", function(confirm) {
+          if (confirm) {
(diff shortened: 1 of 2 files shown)


shlomi-noach:
Thank you for this feature!
I find that the position of the button on the navigation bar consumes too much space. Would you be open to relocate it under, say, the context menu, as other toggles seen in this screenshot?

screen shot 2018-10-07 at 06 15 38

sjmudd:
Mateus, I agree with Shlomi on the size of the button/display.

I'd suggest:
* remove text "global" from the text as you're looking at a global dashboard. So just "recoveries enabled" or "recoveries disabled"
* change the "recoveries disabled" to red as if disabled we want this to be very visible.

Shlomi I'd rather see a smaller icon on the top bar. If recoveries are disabled you really want this to be visible in *all* views.

My preferred location would be sort of between the orchestrator text and the home button but to have an icon which is small and red (disabled) or green (enabled). The exact type of icon or spare probably doesn't matter that much,I'd favour a single light traffic light icon, and a mouse over would give the full text "Recoveries disabled globally" or "Recoveries enabled globally".

mateusduboli:
Hello thanks for the feedback!

@sjmudd I'm adding "global" to the name as this button is shown in all contexts, it's not aware if either the page is looking at a cluster or looking at all of them.

@shlomi-noach, I was thinking the directions of the "read-only" icon when someone is using orchestrator without modification privileges.

I could think of using a more clear a icon or maybe use a mouse-over text to explain what it means.
(Mouse over the elements)

<div title="Global recoveries enabled">
  <spam >:heart:</spam>
</div>
<div title="Global recoveries disabled">
 <spam >:broken_heart:</spam>
</div>

mateusduboli:
I did a quick example here, see if it's good

(The mouse hover shows the correct text, but my recording it's now showing it for some reason)

screencast-localhost-3000-2018 10 08-16-16-23

sjmudd:
The [broken] heart icon is I think confusing but icon size and placement is better.

shlomi-noach:
Heart (full/empty) icon is definitely much better, and consistent with sidebar icons. I'm OK to go with that.

mateusduboli:
I've pushed the code with the heart icon now, and added a alert when the user clicks the button to add some security, since this can be a very dangerous action.

shlomi-noach:
In the future, and to reduce friction, please submit a PR not from your `master` but from a branch. It just makes the testing more trivial. Thanks!

shlomi-noach:
#654 resubmits this PR and is merged

Issue #229, opened 2017-07-06
No systemd init script for orchestrator


sjmudd:
Just a reminder. If I get time I´ll try and write one but I use systems which use systemd for starting processes. The BSD init scripts provided in the current code work but a systemd version would be better.

This issue is to just reflect that this is missing. Usage would depend on the OS you are using.

igroene:
Here is a simple one I wrote:
```
[Unit]
Description="orchestrator: MySQL replication management and visualization"
After=syslog.target network.target

[Service]
Type=simple
ExecStart=/usr/local/orchestrator/orchestrator --verbose http
PIDFile=/var/run/orchestrator.pid
WorkingDirectory=/usr/local/orchestrator

[Install]
WantedBy=multi-user.target
```
Place in file /etc/systemd/system/orchestrator.service

laurent-indermuehle:
Thanks @igroene.
I rely on a mysql backend so I modified the "After" parameter, to avoid Orchestrator starting befor MySQL is ready:

```
[Unit]
Description="orchestrator: MySQL replication management and visualization"
After=syslog.target network.target mysqld.service

[Service]
Type=simple
ExecStart=/usr/local/orchestrator/orchestrator --verbose http
PIDFile=/var/run/orchestrator.pid
WorkingDirectory=/usr/local/orchestrator

[Install]
WantedBy=multi-user.target
```

mcrauwel:
This issue has bee mostly fixed by openark/orchestrator#915

However @Honiix suggestion is still valid, I ran into the same issue today.

laurent-indermuehle:
@mcrauwel This has been addressed here: https://github.com/openark/orchestrator/pull/1012 and release in 3.1.4.

Are you using an older version?

mcrauwel:
@Honiix yeah that documentation piece describes how you can solve it, however it's safe to add the `After=` entry in the actual service-file. After does not *force* the relationship, it just kindly requests to systemd, if you start both services, please make sure MySQL starts first and only then attempt Orchestrator.

[...]

Issue #117, opened 2017-03-25
Do not run orchestrator as root


sjmudd:
The provided init scripts make orchestrator start and run as the root user. That's really not necessary, so it would be good to provide sample init scripts that start orchestrator as a (dedicated?) non-root user.

I would suggest
* detecting if running as root and issuing a warning that it's better not to
* making the init scripts try to start orchestrator as the `orchestrator` user or some specific non-root user
* add comments that any routines called by orchestrator may need to run via sudo to gain enhanced privileges if moving from the current setup.
* binding to port 80 (or low ports) may be troublesome, perhaps requiring orchestrator to drop privileges (not sure if that's an issue in `go` or somehow configuring the user to have such privileges. That's likely to depend on the OS being used, but as not everyone will be binding to low ports this may not be such an issue.
* ensure that log writing can write to the appropriate file(s).
* this is really not a new issue and it has been resolved by many other applications and it seems to make sense for orchestrator too.

sjmudd:
I do notice that dropping privileges seems hard in go so maybe this can't be done yet or easily? See: https://github.com/golang/go/issues/1435

shlomi-noach:
Perhaps just run via `start-stop-daemon`? It's a pretty nice wrapper.

```
Usage: start-stop-daemon [<option> ...] <command>

Commands:
  -S|--start -- <argument> ...  start a program and pass <arguments> to it
  -K|--stop                     stop a program
  -T|--status                   get the program status
  -H|--help                     print help information
  -V|--version                  print version

Matching options (at least one is required):
     --pid <pid>                pid to check
     --ppid <ppid>              parent pid to check
  -p|--pidfile <pid-file>       pid file to check
  -x|--exec <executable>        program to start/check if it is running
  -n|--name <process-name>      process name to check
  -u|--user <username|uid>      process owner to check

Options:
  -g|--group <group|gid>        run process as this group
  -c|--chuid <name|uid[:group|gid]>
                                change to this user/group before starting
                                  process
  -s|--signal <signal>          signal to send (default TERM)
  -a|--startas <pathname>       program to start (default is <executable>)
  -r|--chroot <directory>       chroot to <directory> before starting
  -d|--chdir <directory>        change to <directory> (default is /)
  -N|--nicelevel <incr>         add incr to the process' nice level
  -P|--procsched <policy[:prio]>
                                use <policy> with <prio> for the kernel
                                  process scheduler (default prio is 0)
[...]

PR #653, opened 2018-10-10
Systemd init and orchestrator user

mateusduboli:
Target issue:  #229
Using systemd unit files allows us to do nifty things like auto restarts and managing logs via syslog and other features. So we found interesting to move with that.

Target issue:  #117
Running as a privileged user has various security concerns with possible leaks and vulnerabilities when running in the wild, so this additions use `fpm` to create default "orchestrator" user and group on debs and rpm files.

*Disclaimer*
On this PR I'm actually covering 2 issues, which is not ideal, but since I'm relying on changes made on one commit, I'm actually creating them as a single PR, which can be split later on.


shlomi-noach (commented):
This looks good! Thank you for not breaking behavior and for providing a command line option to build the proper init setup.

I'll experiment with this a bit.

sjmudd:
A couple of things here:
* previous experience with rpm has shown that it is often good (if possible) to use a standard uid. Not doing so can lead to issues if the resulting new user has different values and you start copying files from one system to another.  With orchestrator it's likely that more than one orchestrator app is running on a system and it's likely that you want these uids to be consistent.  Consequently if it's possible in the fpm packaging to provide a suggested (but overridable user/group id) that would be good.
* Generally with the choice between using systemd or using older init scripts if possible I'd prefer the  package to be able to figure out which version is appropriate depending on the target OS and version. How possible that is I'm not sure.  So perhaps it's possible to recognise at least some OS/version combinations and auto-configure these accordingly.  A simple example would be CentOS 6 and earlier uses init scripts, CentOS 7 and later uses systemd.  I'd expect that for Debian and Ubuntu similar rules could be found. Others can probably provide support for their faviourite OS and if the combination can't be found force the user to specify the right value.

So it's nice to see support for systemd and also to move away from running as root. Given orchestrator's default port bindings are above port 1024 this change shouldn't make any big change. However, it may mean that scripts expected to run as root may no longer work or require the use of sudo or equivalent if root functionality is required.  I guess people who have already written such scripts to deploy with orchestrator will need to be aware of this change as it might cause surprises during MySQL failover handling which  would not be good. Maybe if orchestrator should/could log if it's not running as root to make this a bit more noticeable. An extra log line won't cause any harm.

shlomi-noach:
This build fails on one of my boxes: `ERROR: Unrecognised option '--user'` in `fpm` execution.

Which means we depend on a specific version of `fpm` to build `orchestrator`, which is not a good thing. For a while now I've been considering replacing `fpm` with [nfpm](https://github.com/goreleaser/nfpm). `fpm` is a real hassle to install what with ruby versions, various gem dependencies, different OS versions.

So I'm inclined to look into `nfpm` and include it as a binary within `orchestrator's` repo.

shlomi-noach:
@mateusduboli which version of `fpm` are you using? I'm using `fpm-1.10.2`

mateusduboli:
So @shlomi-noach, the problem was it needed package prefixes for user and groups.
It's fixed now :)

shlomi-noach:
Extremely sorry for dropping this. Looking into.

shlomi-noach:
> Generally with the choice between using systemd or using older init scripts if possible I'd prefer the package to be able to figure out which version is appropriate depending on the target OS and version. 

@sjmudd I don't see that there's a generic way to do that. Solutions I see use Post-install scripts to copy either `sysv` or `systemd` script to the right place. I'ts not the prettiest.

> So it's nice to see support for systemd and also to move away from running as root. Given orchestrator's default port bindings are above port 1024 this change shouldn't make any big change. However, it may mean that scripts expected to run as root may no longer work or require the use of sudo or equivalent if root functionality is required. I guess people who have already written such scripts to deploy with orchestrator will need to be aware of this change as it might cause surprises during MySQL failover handling which would not be good.

Oh yes, I agree. It's great that this PR moves away from `root` but we also need to consider existing installations. Thinking about this for a bit.

shlomi-noach:
I'm pivoting at https://github.com/github/orchestrator/pull/915

laurent-indermuehle:
Nice work on systemd :)

Is it just me or you need to add :
`[Install]
WantedBy=multi-user.target`
in order to enable this unit file ?
Before I done that, on RedHat reboot, systemd didn't reload orchestrator.service and the command systemctl enable orchestrator.service did nothing.

john9x:
> Is it just me or you need to add :
> `[Install] WantedBy=multi-user.target`
> in order to enable this unit file ?
> Before I done that, on RedHat reboot, systemd didn't reload orchestrator.service and the command systemctl enable orchestrator.service did nothing.

Same problem

shlomi-noach:
@Honiix could you please open a PR?

PR #660, opened 2018-10-17
Added which-broken-replicas to list replicas with errors

cezmunsta:
Adding `which-broken-replicas` to address https://github.com/github/orchestrator/issues/588; Any replicas of instance `xxx` that show with `LastSQLError` or `LastIOError` are reported back to the user.
```
orchestrator-client -c which-broken-replicas -i xxx
```

### DIFF
--- a/resources/bin/orchestrator-client
+++ b/resources/bin/orchestrator-client
@@ -233,8 +233,8 @@ function filter_keys {
   cat - | jq '.[] | .Key'
 }
 
-function filter_broken {
-  cat - | jq 'reduce .[] as $instance ([]; if $instance.LastSQLError != "\"\"" then . + [$instance] elif $instance.LastIOError != "\"\"" then . + [$instance] else . + [] end)'
+function filter_broken_replicas {
+  cat - | jq '.[] | select((.Slave_SQL_Running == false or .Slave_IO_Running == false) and (.LastSQLError != "" or .LastIOError != "")) | [.]'
 }
 
 function print_key {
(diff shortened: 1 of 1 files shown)


resources/bin/orchestrator-client
```diff
@@ -233,6 +233,10 @@ function filter_keys {
   cat - | jq '.[] | .Key'
 }
 
+function filter_broken {
+  cat - | jq 'reduce .[] as $instance ([]; if $instance.LastSQLError != "\"\"" then . + [$instance] elif $instance.LastIOError != "\"\"" then . + [$instance] else . + [] end)'
```
shlomi-noach: here's a shorter way to write this:

```suggestion
  cat - | jq '.[] | select(.LastSQLError != "" or .LastIOError != "")'
```

shlomi-noach (commented):
Thank you for this PR! Let me also dig around and see if I can make something that suits you better

resources/bin/orchestrator-client
shlomi-noach: Or I think I need to re-array the result though.

cezmunsta: Ah, I used `select` before but had a problem with something, but it was late and I probably mistyped something

resources/bin/orchestrator-client
```diff
@@ -233,6 +233,10 @@ function filter_keys {
   cat - | jq '.[] | .Key'
 }
 
+function filter_broken_replicas {
+  cat - | jq '.[] | select((.Slave_SQL_Running == false or .Slave_IO_Running == false) and (.LastSQLError != "" or .LastIOError != "")) | [.]'
```
shlomi-noach, line 237: `[.]` is a trick that wraps each element in its own array, which is not technically an array of objects, but works well for the rest of filtering.

shlomi-noach:
This PR depends on https://github.com/github/orchestrator/pull/662 to be merged.

shlomi-noach (approved):

PR #661, opened 2018-10-18
GTID: predict replication error and prevent relocation

### LINKED ISSUE #642: Make canMoveViaGTID smarter - be aware of erroneous transactions on new master
First mentioned in https://github.com/github/orchestrator/issues/640 but made into a separate issue to keep things clean.

The function `canMoveViaGTID` may need to be smarter as it currently does not take into account some issues which can arise and which may break replication when used from `moveReplicasViaGTID`. They are:

Erroneous transactions on the new master. if `master_auto_position` = 1 then the slave will try to sync any GTIDs it is missing. Often erroneous transactions will have disappeared from binlogs triggering the sync between the slave and it's new master to fail. So recognising if all needed GTIDs are available would be good.

### LINKED ISSUE #643: Make canMoveViaGTID smarter - do not move a delayed slave if doing so will throw away relay logs it needs.
First mentioned in https://github.com/github/orchestrator/issues/640 but made into a separate issue to keep things clean.

The function `canMoveViaGTID` may need to be smarter as it currently does not take into account some issues which can arise and which may break replication when used from `moveReplicasViaGTID`. They are:

Moving a delayed slave which has a considerable amount of delay (and thus relay logs) may also cause issues:
* if the equivalent binlogs are on the new master they need to be fetched again which is wasteful but may be unavoidable
* if the equivalent binlogs are not on the new master then replication will break as the `CHANGE MASTER TO` command will remove any relay logs prior to reconnecting to the new master.
* it seems wise to check `gtid_executed` and `gtid_purged` on the new master with `Retrieved_Gtid_Set` and `Executed_Gtid_Set` (or `gtid_executed`) from the slave to identify if this might happen and not attempt the relocation in this case, providing a helpful error message why this isn't possible.

shlomi-noach:
Fixes #642 
Fixes #643

With this PR `orchestrator` can predict whether server A can replicate from server B via `GTID`, in terms of missing/purged GTID transactions. 

`orchestrator` will now prevent relocating A under B if B has purged GTID transactions unknown to A: `orchestrator` predicts that if you try to `change master to`, replication will be broken.

`orchestrator` thereby also protects you against losing data: a `change master to` drops all relay logs. In the particular case of `DELAYED` replicas this is especially important: `change master to` drops relay logs, then replication is found to be broken, but... operation cannot be reverted; relay logs are lost. And so this PR protects against such incident.

It will also serve as the base for even more elaborate failover/promotion decisions.

cc @sjmudd 
Special thanks @MarkLeith and @evanelias for assisting in understanding the logic.

### DIFF
--- a/go/inst/instance_topology.go
+++ b/go/inst/instance_topology.go
@@ -575,7 +575,7 @@ func canReplicateAssumingOracleGTID(instance, masterInstance *Instance) (canRepl
 	return subtractGtidSet.IsEmpty(), nil
 }
 
-func instancesGTIDCompatible(instance, otherInstance *Instance) (isOracleGTID bool, isMariaDBGTID, compatible bool) {
+func instancesAreGTIDAndCompatible(instance, otherInstance *Instance) (isOracleGTID bool, isMariaDBGTID, compatible bool) {
 	isOracleGTID = (instance.UsingOracleGTID && otherInstance.SupportsOracleGTID)
 	isMariaDBGTID = (instance.UsingMariaDBGTID && otherInstance.IsMariaDB())
 	compatible = isOracleGTID || isMariaDBGTID
@@ -583,7 +583,7 @@ func instancesGTIDCompatible(instance, otherInstance *Instance) (isOracleGTID bo
 }
 
 func checkMoveViaGTID(instance, otherInstance *Instance) (err error) {
-	isOracleGTID, _, moveCompatible := instancesGTIDCompatible(instance, otherInstance)
+	isOracleGTID, _, moveCompatible := instancesAreGTIDAndCompatible(instance, otherInstance)
 	if !moveCompatible {
 		return fmt.Errorf("Instances %+v, %+v not GTID compatible or not using GTID", instance.Key, otherInstance.Key)
 	}
@@ -2569,7 +2569,7 @@ func relocateBelowInternal(instance, other *Instance) (*Instance, error) {
 		return nil, log.Errorf("Relocating binlog server %+v below %+v turns to be too complex; please do it manually", instance.Key, other.Key)
 	}
 	// Next, try GTID
-	if _, _, gtidCompatible := instancesGTIDCompatible(instance, other); gtidCompatible {
+	if _, _, gtidCompatible := instancesAreGTIDAndCompatible(instance, other); gtidCompatible {
 		return moveInstanceBelowViaGTID(instance, other)
 	}
 
--- a/go/inst/instance_topology.go
+++ b/go/inst/instance_topology.go
@@ -554,17 +554,7 @@ Cleanup:
 }
 
 func canReplicateAssumingOracleGTID(instance, masterInstance *Instance) (canReplicate bool, err error) {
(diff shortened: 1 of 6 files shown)


go/inst/instance_topology.go
```diff
@@ -553,25 +553,55 @@ Cleanup:
 	return instance, err
 }
 
-func canMoveViaGTID(instance, otherInstance *Instance) (isOracleGTID bool, isMariaDBGTID, canMove bool) {
-	isOracleGTID = (instance.UsingOracleGTID && otherInstance.SupportsOracleGTID)
-	isMariaDBGTID = (instance.UsingMariaDBGTID && otherInstance.IsMariaDB())
+func canReplicateAssumingOracleGTID(instance, masterInstance *Instance) (canReplicate bool, err error) {
+	instanceTotalGtid := fmt.Sprintf("%s,%s", instance.GtidPurged, instance.ExecutedGtidSet)
```
evanelias: Since `gtid_executed` is a superset of `gtid_purged`, I believe you can just use instance.ExecutedGtidSet as-is for this. (But I'm not too familiar with Orchestrator's codebase, so apologies if I'm misreading!)

shlomi-noach: Thank you!

shlomi-noach:
Added `/api/can-replicate-from-gtid/hostname/port/belowHostname/belowPort` and `orchestrator-client -c can-replicate-from-gtid`. e.g.:

```shell
$ orchestrator-client -c can-replicate-from-gtid -i my-host-12345 -d other-host-98765
other-host-98765

$ orchestrator-client -c can-replicate-from-gtid -i my-host-12345 -d errant-host-44444
Instance errant-host-44444:3306 has purged GTID entries not found on my-host-12345:3306
```

shlomi-noach:
This has recently been battle tested in production and did a good job.

go/inst/oracle_gtid_set.go
```diff
@@ -68,6 +68,45 @@ func (this *OracleGtidSet) RemoveUUID(uuid string) (removed bool) {
...
+		}
+	}
+	if anythingRemoved {
+		this.GtidEntries = filteredEntries
+	}
+	return anythingRemoved
+}
+
+// SharedUUIDs returns UUIDs (range-less) that are shared between the two sets
+func (this *OracleGtidSet) SharedUUIDs(other *OracleGtidSet) (shared []string) {
```
shlomi-noach, line 97: This actually ended up not being used, but I'd like to keep the (tested) functionality.

go/inst/oracle_gtid_set.go
```diff
@@ -68,6 +68,45 @@ func (this *OracleGtidSet) RemoveUUID(uuid string) (removed bool) {
 	return removed
 }
 
+// RetainUUID retains only entries that belong to given UUID.
+func (this *OracleGtidSet) RetainUUID(uuid string) (anythingRemoved bool) {
+	return this.RetainUUIDs([]string{uuid})
+}
+
+// RetainUUIDs retains only entries that belong to given UUIDs.
+func (this *OracleGtidSet) RetainUUIDs(uuids []string) (anythingRemoved bool) {
```
shlomi-noach, line 77: This actually ended up not being used, but I'd like to keep the (tested) functionality.

PR #663, opened 2018-10-18
Adding which-cluster-osc-running-replicas command

shlomi-noach:
Fixes https://github.com/github/orchestrator/issues/588

This PR adds `which-cluster-osc-running-replicas` command to `orchestrator-client` (though not to `orchestrator` binary commands).

`which-cluster-osc-running-replicas` will report OSC replicas which are known to be replicating well.

cc @cezmunsta

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,4 @@ main
 *.pcap
 *.log
 .vendor/go19
-bin
+./bin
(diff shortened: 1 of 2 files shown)


cezmunsta (approved):
LGTM :sheep:

PR #665, opened 2018-10-23
manual recovery overrides globally-disabled recoveries

shlomi-noach:
globally-disabled recoveries were preventing manual recoveries from taking place. This PR ensures that a human can always choose to run a recovery.

PS 
Maybe I'm just pedantic; but since this is a public repo, and the timing of this PR may be of interest to people, it would be wrong to assume that this PR resolves recent events or that recent events were caused by lack of this PR. Just saying.


tomkrouper (approved):

shlomi-noach:
Also, it would be wrong to assume this PR does _not_ have something to do with yesterday's events 😂 

So the problem here did not lead to yesterday's events, but was something we noticed while handling the incident. It was a nuisance during an otherwise busy time.

PR #666, opened 2018-10-24
More visibility in logs in applying post-recovery steps

shlomi-noach:

A few more log entries in recovery audit clarify whether each of the steps:

- `reset slave all` on new master
- `set read_only=0` on new master
- `set read_only=1` on old master

... have been successful or not.

Neither of these steps is a blocker for a recovery, but needs to be better audited.


shlomi-noach:
sample output is as expected:

-- | --
2018-10-24 00:59:32 | - RecoverDeadMaster: applying RESET SLAVE ALL on promoted master: success=true
2018-10-24 00:59:32 | - RecoverDeadMaster: applying read-only=0 on promoted master: success=true
2018-10-24 00:59:32 | - RecoverDeadMaster: applying read-only=1 on demoted master: success=false

shlomi-noach:
noteworthy that the information exists in the general log, but this PR groups it within the recovery-audit log, which is well defined per-recovery and is easier to parse.

PR #673, opened 2018-10-29
Avoiding infinite ancestry on circular replication

shlomi-noach:
Fixes https://github.com/github/orchestrator/issues/672

This PR adds a stop-condition to prevent infinitely growing `ancestry_uuid` in a circular replication situation.

cc @cohenjo

### DIFF
--- a/go/inst/instance_dao.go
+++ b/go/inst/instance_dao.go
@@ -773,17 +773,26 @@ Cleanup:
 	waitGroup.Wait()
 
 	if instanceFound {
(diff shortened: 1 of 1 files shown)


cohenjo:
Hi - Just finished testing - seems to resolve the issue - we don't see the infinite ancestry UUID list.
Thank you very much for the quick response 👍

shlomi-noach:
This is now feature complete.

`ancestry_uuid`: both co-masters include each other in ancestry
`gtid_errant`: a co-master's _own transactions_ are ignored when evaluating errant transactions.

cohenjo:
tested both - it cleared our errant transaction marking from the co-masters.
Thanks 👍

Issue #674, opened 2018-10-30
FR: orchestrator-client user credentials from environment variables


mateusduboli:
Enable orchestrator-client to be configurable via environment variables, my suggestions are:
```bash
ORCHESTRATOR_AUTH_USER
ORCHESTRATOR_AUTH_PASSWORD
```

shlomi-noach:
@mateusduboli can you please clarify, are these credentials for HTTP Auth?

mateusduboli:
Yes, those are for `orchestrator-client` have some way to define that as environment variables, so on the `/etc/profile.d/orchestrator-client` script we can define the credentials

mateusduboli:
Btw one other change for us is to define this file depending on the user like:
```
$PWD/orchestrator-profile
$HOME/.orchestrator-profile
/etc/profile.d/orchestrator-profile
```

shlomi-noach:
closed by #675

PR #675, opened 2018-10-30
orchestrator-client: Add credentials to environment variables

mateusduboli:
Related to issue #674

### DIFF
--- a/resources/bin/orchestrator-client
+++ b/resources/bin/orchestrator-client
@@ -50,7 +50,7 @@ promotion_rule=
 pool=
 hostname_flag=
 api_path=
-basic_auth=":"
+basic_auth="${ORCHESTRATOR_USER}:${ORCHESTRATOR_PASSWORD}"
 
 instance_hostport=
 destination_hostport=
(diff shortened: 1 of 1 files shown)


shlomi-noach (changes requested):
Thank you for this. The PR makes sense.
Can you please rename the variables to something more descriptive, such as `ORCHESTRATOR_AUTH_USER`, `ORCHESTRATOR_AUTH_PASSWORD`?

shlomi-noach (approved):

sjmudd:
This actually breaks having credentials in the `ORCHESTRATOR_API` variable, e.g. of the form: 

```
export ORCHESTRATOR_API=https://some_user:some_password@orchestrator.example.com
```

The fix for this seems to me to be to only set `basic_auth` if `ORCHESTRATOR_AUTH_USER` is not empty, as otherwise the value is populated with `:` which is useless.

sjmudd:
Also the usage of `--basic --user "${basic_auth}"` if `${basic_auth}` is not set will break. This is not taken into account.

I will offer an updated PR for this which should address both issues.

PR #676, opened 2018-10-31
Instance write buffer metrics

mateusduboli:

Related issue: https://github.com/github/orchestrator/issues/113

### Description
This is the rebased version of #146, it:
- Adds fixed collection names to avoid errors when creating new ones
- Adds a write-buffer collection to store individual write-buffer counters to be either aggregated or exposed in raw format
- Adds endpoints to `metrics/instance-write-buffer/{raw,aggregated}/:seconds`
- Updates go-metrics to the latest commit (3113b8401b8a98917cde58f8bbd42a1b1c03b1fd)

### DIFF
--- a/go/collection/collection.go
+++ b/go/collection/collection.go
@@ -85,18 +85,28 @@ type Collection struct {
 	expirePeriod time.Duration // time to keep the collection information for
 }
 
+// This is a collection name
+type Name string
+
+const (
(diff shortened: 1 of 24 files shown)


mateusduboli:
I'm upgrading the go-metrics in this, PR but it's just because my code only works on the latest version of the library, I can move this upgrade to another PR if it the changes are good

PR #681, opened 2018-10-31
Added get_url to handle authentication

cezmunsta:
Fixes https://github.com/github/orchestrator/issues/680


shlomi-noach (changes requested):
Thank you for this!

May I ask for the following changes:

- Extract `"curl"` outside the function. so that we do:
```
leader_check=$(curl $(get_curl_auth_params) -m 1 -s -o /dev/null -w "%{http_code}" "${api}/leader-check")
```

- Only compute the function once, and not per invocation?

shlomi-noach:
Also, can you please take into account the change in https://github.com/github/orchestrator/pull/675?

cezmunsta:
```
# ./orchestrator-client -b orchestrator:xxx -c all-instances
10.0.0.20:3306

# ORCHESTRATOR_AUTH_USER=orchestrator ORCHESTRATOR_AUTH_PASSWORD=xxx ./orchestrator-client -c all-instances
10.0.0.2:3306

# ./orchestrator-client -c all-instances                                                                                                              
parse error: Invalid numeric literal at line 1, column 4
^C^

# edit /etc/profile.d/orchestrator-client.sh

# cut -f1 -d= /etc/profile.d/orchestrator-client.sh
ORCHESTRATOR_API

# ./orchestrator-client -c all-instances
10.0.0.2:3306
```

cezmunsta:
```
# ./orchestrator-client -c all-instances
orchestrator-client[26762]: Cannot access orchestrator at http://localhost:3000/api.  Check ORCHESTRATOR_API is configured correctly and orchestrator is running
```
The value of `curl_auth_params` is tested against `401 Unauthorized` and fails before attempting to execute `curl ...` instead of erroring on `jq`

cezmunsta:
@shlomi-noach OK now?

shlomi-noach:
@cezmunsta this will unfortunately fail when `orchestrator_api` contains multiple entries, as per https://github.com/github/orchestrator/pull/681/files#diff-79f20b792813786c70a731862c24143eL157

shlomi-noach (approved):

shlomi-noach:
As @cezmunsta clarified elsewhere, the check will not fail when orchestrator_api contains multiple entries.

Issue #691, opened 2018-11-06
./tests/integration/test.sh failed because missing default value


MOON-CLJ:
```
+ check_db
+ echo 'select 1;'
++ run_queries /tmp/orchestrator-test.sql
++ queries_file=/tmp/orchestrator-test.sql
++ '[' mysql == sqlite ']'
++ mysql --default-character-set=utf8mb4 test -ss
+ query_result=1
+ '[' 1 '!=' 1 ']'
+ echo '- check_db OK'
- check_db OK
+ test_all
+ deploy_internal_db
+ echo 'Deploying db'
Deploying db
+ cmd='/tmp/orchestrator-test     --config=/tmp/orchestrator.conf.json
    --debug     --stack     -c redeploy-internal-db'
+ echo_dot
+ echo -n .
.+ echo /tmp/orchestrator-test --config=/tmp/orchestrator.conf.json --debug --stack -c redeploy-internal-db
+ echo_dot
+ echo -n .
.+ bash /tmp/orchestrator-test.bash
+ '[' 0 -ne 0 ']'
+ echo '- deploy_internal_db result: 0'
- deploy_internal_db result: 0
+ '[' 0 -ne 0 ']'
+ echo '- deploy_internal_db OK'
- deploy_internal_db OK
+ test_pattern=.
+ find ./tests/integration '!' -path . -type d -mindepth 1 -maxdepth 1
+ xargs ls -td1
+ cut -d / -f 4
+ egrep .
+ read test_name
+ test_single analysis-all-intermediate-master-replicas-not-replicating
+ local test_name
+ test_name=analysis-all-intermediate-master-replicas-not-replicating
+ echo -n 'Testing: analysis-all-intermediate-master-replicas-not-replicating'
Testing: analysis-all-intermediate-master-replicas-not-replicating+ echo_dot
+ echo -n .
.+ run_queries ./tests/integration/create-per-test.sql
+ queries_file=./tests/integration/create-per-test.sql
+ '[' mysql == sqlite ']'
+ mysql --default-character-set=utf8mb4 test -ss
ERROR 1364 (HY000) at line 14: Field 'master_uuid' doesn't have a default value
```

shlomi-noach:
I'm sorry, could you please clarify again what the issue is?

MOON-CLJ:
@shlomi-noach i run "./tests/integration/test.sh", then encounter "ERROR 1364 (HY000) at line 14: Field 'master_uuid' doesn't have a default value"

[...]

PR #692, opened 2018-11-06
fix create-per-test missing default value

MOON-CLJ:
ref: https://github.com/github/orchestrator/issues/691


shlomi-noach (changes requested):
Sorry, can you please clarify what you are changing and why?

shlomi-noach (approved):

shlomi-noach:
Clarified on #691

PR #707, opened 2018-11-11
Errant GTIDs: inject empty transactions on master

shlomi-noach:
Complementing https://github.com/github/orchestrator/pull/617, when a replica has errant transactions (#607) `orchestrator` can now "fix' the problem by injecting the necessary empty transactions on the master of the cluster.

`orchestrator` will auto-explode the transaction set, and will issue a transaction per single entry. i.e. if the errant transaction set is `00020194-3333-3333-3333-333333333333:1-3` then `orchestrator` will inject three empty transactions on the cluster's master:

- `00020194-3333-3333-3333-333333333333:1`
- `00020194-3333-3333-3333-333333333333:2`
- `00020194-3333-3333-3333-333333333333:3`

This operation supports master-master setups.

It requires a writable master.


SQLadmin:
Hi @shlomi-noach Its great news for us. But just want a clarification.

I just tried but the result I got is `Have injected 1 transaction on cluster master staging-mysql-05:3306`
Which means the transaction has been executed on the master, so After the failover, even the new master tries to replicate those transactions to the failed node or de-promoted as a slave, then that mysql node will lookup the gtid_executed table and confirms that this transaction has already executed to skip that transaction and move forward.

Is this the concept behind this?

shlomi-noach:
> then that mysql node will lookup the gtid_executed table and confirms that this transaction has already executed to skip that transaction and move forward.

Yes, that's exactly the idea. It sounds to me like you have a different problem, and this is not the solution to that problem. However, I am not sure I understand what the problem is.

SQLadmin:
@shlomi-noach There are no issues with this. Just wanted to understand how this feature 
 works.

shlomi-noach:
The idea is that you've, for example, issued a `analyze table` on a replica, which unfortunately generated an errant GTID on the replica, and now it's supposedly inconsistent with the rest of the cluster, even though there is no data inconsistency. In cases like this you can inject an empty GTID entry on the master, which will propagate to all replicas, so that every server has that GTID entry in their bookkeeping, even though they applied nothing for that GTID.

SQLadmin:
Great thanks :)

PR #727, opened 2018-11-21
FigureClusterName: Avoid heavy queries if clusterHint is empty

sjmudd:

This can be called from `orchestrator -c which-cluster`.
If no hint is provided then the first 2 queries are called and will trigger a 3 table join in `ReadClustersInfo` without a where clause. Worse the resultset won't match what's expected so they will fail.  This wastes resources and if the number of monitored instances or clusters is high the join can become expensive.

Some of my automation uses `orchestrator -c which-cluster` frequently and this can trigger unnecessary load on orchestrator. Removing these 2 _redundant_ queries helps avoid this.

- [ ] code is built via `./build.sh`   (fails again as my golang is 1.11)


sjmudd:
Feel free to wait while I test this but given what I've seen I'm confident that this should reduce unnecessary queries when no hint is provided.

shlomi-noach (approved):
Good catch!

shlomi-noach:
@sjmudd on my side testing is good.

PR #733, opened 2018-11-22
Global web message

shlomi-noach:
A `WebMessage` [configuration variable](https://github.com/github/orchestrator/pull/733/files#diff-275bf73f26d818144b1ddd4a37797935R259) is added. If non-empty, a web message is displayed on all web pages, just under the title bar, with given text. Example:

```json
  "WebMessage": "LOCAL TESTING",
```
screen shot 2018-11-22 at 07 24 41


shlomi-noach:
@sjmudd I'm merging this for now. You had the idea of running a query to generate the message -- we can iterate on that later.

PR #738, opened 2018-11-25
replace-promoted-replica-with-candidate: relocate replicas 

shlomi-noach:
Upon a 2-step recovery (a candidate replica is promoted on top of the immediate promoted replica), we also relocate-up the remaining replicas. Better illustrated below:

Assume
```
m
+- r1
+- r2
+- r3
+- cand
```

master crashes and the immediate suggestion is to promote `r2`:
```
r2
+- r1
+- r3
+- cand
```

but `cand` is found to be a candiate, that we promote it on top of `r2`:
```
cand
+ r2
   +- r1
   +- r3
```
thus far -- the existing logic in `master` branch.

This PR adds moving `r1, r3` upwards, to replicate from `cand`, as follows:
```
cand
+ r1
+ r2
+ r3
```

The relocation of sub-replicas respects `PostponedFunctionsContainer`, and as such the operation may run asynchronously to the actual promotion of `cand`, and we can speed the availability of the writer while still rewiring the topology.

### DIFF
--- a/go/logic/topology_recovery.go
+++ b/go/logic/topology_recovery.go
@@ -746,6 +746,7 @@ func replacePromotedReplicaWithCandidate(topologyRecovery *TopologyRecovery, dea
 
 			relocatedReplicas, _, err, _ := inst.RelocateReplicas(&promotedReplica.Key, &candidateInstance.Key, "")
 			log.Debugf("replace-promoted-replica-with-candidate: + relocated %+v replicas of %+v below %+v", len(relocatedReplicas), promotedReplica.Key, candidateInstance.Key)
+			AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("relocated %+v replicas of %+v below %+v", len(relocatedReplicas), promotedReplica.Key, candidateInstance.Key))
 			return log.Errore(err)
 		}
 		postponedFunctionsContainer := &topologyRecovery.PostponedFunctionsContainer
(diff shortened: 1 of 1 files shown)


ghost:
@shlomi-noach, I've tested that. I think it will work fine for the case you've described here.
But for my test it was not perfect still. 
I had:
```
M1 -> R1
   -> Cand2
↕
M2 -> R2 
```
In the middle:
```
M2 -> R1
   -> Cand2
   -> R2 
```
At the end:
```
Cand2 -> R1
      -> M2
      -> R2 
```
While the correct result will be:
```
Cand2 -> R1
      -> M2 -> R2 
```
This can also happen in case of single master. When promoted most up-to-date replica has slaves behind it. At the end these slaves become slaves of Cand.

shlomi-noach:
@pasha167 that would be too complex to implement for now, because it involves recording the list of replicas of any intermediate master in the topology prior to beginning the recovery, such that after the 2-step promotion we would extract all replicas of 1-step promoted server that were not on that original list.

I'm going to not work on that, and will leave some work to the human DBA who gets notified that a failover took place. The resulting topology before this PR is valid; it is valid with this PR; it would be valid if I were to implement your requested topology change. And while your request makes absolute sense, my experience suggests that some other user would want the opposite. Seeing that this would be overly complicated, I prefer to let it be.

ghost:
Yes, I perfectly understand. If there is no easy way then, let's cancel this change.
Anyway thank you for trying.

shlomi-noach:
> let's cancel this change.

I'm not sure if you're referring to the the change in this Pr or the additional suggested one.

The change in this PR is good to go as far as I'm concerned and makes sense.

ghost:
I was talking about this PR in general, thinking that changing the hierarchy of slaves may be dangerous. 
For example in the following case:
```
m
+- r1
+- r2 
      +- r3
+- cand
```
If r2 has local writes, then the moving r3 to cand is not good.
But thinking a bit more, I understand that in that case r2 should be filter out from promotion completely. So it is more responsibility of DBA to configure properly.

Yes, this PR is good.

PR #746, opened 2018-12-02
Auto-setting MySQLConnectionLifetimeSeconds

shlomi-noach:
Closes https://github.com/github/orchestrator/issues/559

If `MySQLConnectionLifetimeSeconds` is undefined, we define it to be `@@wait_timeout - 5sec`.


shlomi-noach:
This seems to cause connection leak, and I'm seeing `too many open files` errors.

PR #749, opened 2018-12-07
Give a few clues as to how to use Dockerfile

sjmudd:
Helps users know how to use the Dockerfile otherwise they have to poke around and see.
These are just comments, so no code change.


shlomi-noach (approved):
thank you

PR #761, opened 2018-12-17
Experiment: No retries for RESET MASTER on errant GTID cleanup

shlomi-noach:
Observing recurring server crashes when applying `gtid-errant-reset-master` on production replicas, this PR removes the retries by way of experimentation.

cc @ggunson


shlomi-noach:
closed in favor of #762

PR #762, opened 2018-12-19
reset-master operation: wait for replication to stop

shlomi-noach:
As pointed out by @ggunson , `ErrantGTIDResetMaster()` issues a `reset master` immediately following a `stop slave` operation, but without verifying that replication has indeed stopped. e.g. SQL thread could still be busy.

We've seen crashes in production at running `reset master`. 

In this PR we actively wait (or timeout) for replication to stop, before running `reset master`.

### DIFF
--- a/go/inst/instance_dao.go
+++ b/go/inst/instance_dao.go
@@ -242,7 +242,7 @@ func (instance *Instance) checkMaxScale(db *sql.DB, latency *stopwatch.NamedStop
 }
 
 // areReplicationThreadsRunning checks if both IO and SQL threads are running
-func areReplicationThreadsRunning(instanceKey *InstanceKey) (ioThreadRunning, sqlThreadRunning bool, err error) {
+func areReplicationThreadsRunning(instanceKey *InstanceKey) (ioThreadRunning bool, sqlThreadRunning bool, err error) {
 	db, err := db.OpenTopology(instanceKey.Hostname, instanceKey.Port)
 	if err != nil {
 		return ioThreadRunning, sqlThreadRunning, err
--- a/go/inst/instance_dao.go
+++ b/go/inst/instance_dao.go
@@ -53,6 +53,11 @@ const (
 	errorIOTimeout         = "i/o timeout"
 )
 
+var replicationThreadStatesMapping = map[bool]string{
+	false: "No",
+	true:  "Yes",
+}
+
 var instanceReadChan = make(chan bool, backendDBConcurrency)
 var instanceWriteChan = make(chan bool, backendDBConcurrency)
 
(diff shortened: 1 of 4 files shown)


go/inst/instance_dao.go
```diff
@@ -242,18 +242,17 @@ func (instance *Instance) checkMaxScale(db *sql.DB, latency *stopwatch.NamedStop
 }
 
 // areReplicationThreadsRunning checks if both IO and SQL threads are running
-func areReplicationThreadsRunning(instanceKey *InstanceKey) (replicationThreadsRunning bool, err error) {
+func areReplicationThreadsRunning(instanceKey *InstanceKey) (ioThreadRunning, sqlThreadRunning bool, err error) {
```
tomkrouper: Not sure on go syntax, but would this be more right?
```suggestion
func areReplicationThreadsRunning(instanceKey *InstanceKey) (ioThreadRunning bool, sqlThreadRunning bool, err error) {
```

go/inst/instance_dao.go
```diff
@@ -242,18 +242,17 @@ func (instance *Instance) checkMaxScale(db *sql.DB, latency *stopwatch.NamedStop
...
 	db, err := db.OpenTopology(instanceKey.Hostname, instanceKey.Port)
 	if err != nil {
-		return replicationThreadsRunning, err
+		return ioThreadRunning, sqlThreadRunning, err
 	}
 	err = sqlutils.QueryRowsMap(db, "show slave status", func(m sqlutils.RowMap) error {
-		ioThreadRunning := (m.GetString("Slave_IO_Running") == "Yes")
-		sqlThreadRunning := (m.GetString("Slave_SQL_Running") == "Yes")
-		replicationThreadsRunning = ioThreadRunning && sqlThreadRunning
+		ioThreadRunning = (m.GetString("Slave_IO_Running") == "Yes")
```
ggunson: An issue with this is that you're deciding that `Slave_*_Running = Yes` is the boolean check against which to decide that replication is fully running or fully stopped.

I'm not sure about `Slave_SQL_Running`'s options but `Slave_IO_Running` can also be "Connecting". So that's at least one case where this check would say that replication is not running even though the IO thread is (or, starting to, or trying to).

go/inst/instance_dao.go
shlomi-noach: Both are correct, and I like yours better.

go/inst/instance_dao.go
shlomi-noach: That's a good point.

shlomi-noach: fixed

shlomi-noach:
I will merge this PR even though we haven't yet verified the reasoning for the `reset master` crash, or at least have not provided a reliable way to reproduce the error.

Reiterating an internal issue, @ggunson suggests the retries can be the cause of the crash, as follows:

- a 1st connection breaks on `Connection Invalid`
- but the `reset master` it invoked is still underway
- `orchestrator` issues a retry, running a 2nd, _concurrent_ `reset master`
- crash

We're yet to reproduce this reliably, but will not be working on this actively in the short term.

PR #764, opened 2018-12-26
Remove gh-ost content from issue template

ggunson:
### Description

This PR removes mention of https://github.com/github/gh-ost from the issue creation template.

Ideally I'd like a list of information requested rather than "please provide a test case and error output" but I'm not sure what is needed for orchestrator (version? command? replication settings?). 

No issue connected to this, sorry.


shlomi-noach (approved):
🚢

Issue #763, opened 2018-12-26
Incorrect usage of ReplicaRunning()


ggunson:
In investigating https://github.com/github/orchestrator/pull/761 with the concern that `RESET MASTER` might be being issued before replication had fully stopped, I found that the check for whether replication was running or not wasn't complete.

The check used in some parts of orchestrator to tell that replication was running on a server was:

https://github.com/github/orchestrator/blob/a58ad0fdd55b42fddf4553f7db02002db9d6c9d6/go/inst/instance.go#L256-L259

The function returns true if the host is a replica and both sql and IO threads are "Yes". It can then return false if one thread is still running (or the IO thread is "Connecting") as well as when both threads are stopped. So ReplicaRunning() is an invalid function to use to check that replication has 100% stopped.

Currently the function is being used in a few places, but the places where I see a possible concern are:

- [SyncReplicaRelayLogs()](https://github.com/github/orchestrator/blob/master/go/agent/instance_topology_agent.go#L27) -- if the IO thread is still running then the relay log contents are still being written to, and if the SQL thread is still on then it may still be applying relay log contents.

- [StartSlaveUntilMasterCoordinates()](https://github.com/github/orchestrator/blob/a58ad0fdd55b42fddf4553f7db02002db9d6c9d6/go/inst/instance_topology_dao.go#L479) -- orc might try running `start slave until` on a host where one of the threads is already running, which in the console results in only a warning and **no change** to the `until` values. 

- [EnableMasterSSL()](https://github.com/github/orchestrator/blob/master/go/inst/instance_topology_dao.go#L578) -- `CHANGE MASTER` could try executing on a replica with one of the threads running (which should error in this case)*

- [ChangeMasterTo()](https://github.com/github/orchestrator/blob/master/go/inst/instance_topology_dao.go#L605) -- similar to above

- [ResetSlave()](https://github.com/github/orchestrator/blob/master/go/inst/instance_topology_dao.go#L706) -- [RESET SLAVE](https://dev.mysql.com/doc/refman/5.7/en/reset-slave.html) could be run on a replica where one of the threads is still running, which I believe would result in an error but you'll want to double-check

- [ResetMaster()](https://github.com/github/orchestrator/blob/master/go/inst/instance_topology_dao.go#L739) -- RESET MASTER could happen while, say, the SQL thread is still running, so writes could happen after the RESET MASTER that you didn't intend (and if you're going to change gtid_purged or something after this it'll be incorrect)

[...]

PR #767, opened 2018-12-30
ReplicationThreadState: multi-valued logic

shlomi-noach:
Fixes https://github.com/github/orchestrator/issues/763

Until today, replication thread state was a boolean: `true` when running, `false` otherwise. However, a IO thread could also be in a `Connecting` state, which is neither running nor stopped. And this state, as described in #763, can lead to wrong assumptions.

This PR introduces multi-valued logic. `ReplicationthreadState` is introduced, which can be:

- `ReplicationthreadStateStopped`
- `ReplicationthreadStateRunning`
- `ReplicationthreadStateOther`

In this initial commit, most critical existing logic has been converted to use these states as opposed to relying on a boolean value. 

There's more places in the code to modify, in next commits.

cc @ggunson


shlomi-noach:
Added `ReplicationThreadStateNoThread` state, to strictly identify the case of a master: no replication configured.

ggunson:
Just to clarify the purpose of https://github.com/github/orchestrator/issues/763:

https://github.com/github/orchestrator/issues/763 is not solely about the fact that the IO thread can be in the "Connecting" state, but that the logic of determining the replication status was incomplete.

If true is 

> (SQL_Running=Yes) && (IO_Running=Yes)

Then the opposite of that, based on the distribution rules of negation, is:

> !((SQL_Running=Yes) && (IO_Running=Yes)) = !(SQL_Running=Yes) || !(IO_Running=Yes)

This then leads to all the states that the logic would say replication is stopped:

1. IO_Running=No
   SQL_Running=No 

1. IO_Running=No
   SQL_Running=Yes

1. IO_Running=Yes
   SQL_Running=No

1. IO_Running=Connecting
   SQL_Running=Yes

1. IO_Running=Connecting
   SQL_Running=No

1. (Anything else I might've missed?)

And that was the concern, more than just the extra value that Slave_IO_Running can be. This PR looks like it's dealing with all of that so far; this is just to document the extent of the reasoning for anyone reading this.

go/inst/instance.go
```diff
@@ -255,12 +257,17 @@ func (this *Instance) IsMaster() bool {
 
 // ReplicaRunning returns true when this instance's status is of a replicating replica.
 func (this *Instance) ReplicaRunning() bool {
-	return this.IsReplica() && this.Slave_SQL_Running && this.Slave_IO_Running
+	return this.IsReplica() && this.ReplicationSQLThreadState.IsRunning() && this.ReplicationIOThreadState.IsRunning()
 }
 
 // NoReplicationThreadRunning returns true when neither SQL nor IO threads are running (including the case where isn't even a replica)
```
ggunson, line 263: How does this return true if the host isn't a replica? `IsStopped()` maps to `ReplicationThreadStateStopped` but doesn't include `ReplicationThreadStateNoThread`? Or am I reading this incorrectly?

shlomi-noach: If the host isn't a replica this method returns `false`, which is the expected behavior.

PR #771, opened 2019-01-11
Wait for SQL thread to catch up instead of aborting the failover

jordanwheeler:
Related issue: https://github.com/github/orchestrator/issues/736

### Description

We don't want to promote a new replica if the SQL thread is not up to date, so we set `FailMasterPromotionIfSQLThreadNotUpToDate` to true. Orchestrator will then abort a failover if the SQL thread isn't fully caught up on the promoted replica. If that happens, a human now has to come and reconfigure the cluster properly, since Orchestrator was forced to abort before completing the failover. Rather than abort the failover in this scenario, this PR changes the behaviour to `DelayMasterPromotionIfSQLThreadNotUpToDate` instead. Now if the promoted replica is not up to date, Orchestrator will wait until it fully catches up before proceeding normally.

There are a couple other important changes here too. We have our clusters configured in a co-master setup. [The code that checks](https://github.com/github/orchestrator/blob/master/go/logic/topology_recovery.go#L1269-L1272) `DelayMasterPromotionIfSQLThreadNotUpToDate` in `checkAndRecoverDeadCoMaster` happens _after_ [the promoted replica is detached](https://github.com/github/orchestrator/blob/master/go/logic/topology_recovery.go#L1216-L1219). So I've moved this logic up before the detach step, so that Orchestrator will check the correct positions before that happens.

After realizing that this check happened after the host is already detached, the more interesting question became: how is `SQLThreadUpToDate` _ever_ true on the promoted replica? After detaching its `Read_Master_Log_Pos` is `4` and its `Exec_Master_Log_Pos` is 0.  So that path should always abort a co-master failover if `FailMasterPromotionIfSQLThreadNotUpToDate` is `true`, but it doesn't. That's because `SQLThreadUpToDate` is reading those positions from Orchestrator's backend database, and those values are not updated because `ReadTopologyInstance` is not called in between the promoted replica being detached, and the call to `SQLThreadUpToDate`.

There are other places in the codebase where `ReadTopologyInstance` is explicitly called before checking `SQLThreadUpToDate`, but it wasn't here. To be as safe as possible, I modified `SQLThreadUpToDate` so that it always calls `ReadTopologyInstance` before checking the SQL thread position. The way I've implemented it is not the cleanest possible way to do it, but it is a very reliable way to do it. It could be done a bit differently.

I also realize that simply changing the name of a configuration variable is not the proper way to roll this out. There are also test failures when running `go test ./go/...` on this branch, but I also had similar test failures when running that against the latest master. So this branch still needs some work before merging, but I figured we could at least start the conversation here.

@shlomi-noach @pasha167 @akshaysuryawanshi


shlomi-noach (changes requested):
Thank you very much!

- `FailMasterPromotionIfSQLThreadNotUpToDate` should not be removed; this would break many people's prodction behavior. We should add `DelayMasterPromotionIfSQLThreadNotUpToDate`, and run a configuration test to see that the two parameters do not conflict (ie not both `true`, something which is ambiguous)

- I've recently hit cases where SQL thread does not strictly catch up with IO thread. That is, I _think_ the relay logs I had were cut in a position that is "in between" statements. I haven't analyzed this yet, but I think it's something like a `rows_map` or similar, where the SQL thread does not apply the change until it sees followup entries. Does that make sense to you? (I saw this is some automated testing we did).
If that is the case, then `SQLThreadUpToDate()` could potentially never return `true`. What are your thoughts?

jordanwheeler:
> FailMasterPromotionIfSQLThreadNotUpToDate should not be removed

Yes, I realized that was not going to fly. Are you suggesting we keep both behaviours around? Abort if `FailMasterPromotion` is `true`, but it's `false` and `FailMasterPromotion` is `true`, wait instead?

> If that is the case, then SQLThreadUpToDate() could potentially never return true. What are your thoughts?

That's an interesting one. We thought that adding some timeout configuration for this might make sense, so Orchestrator would wait some configurable amount of seconds for the SQL thread to catch up, and then abort. We were OK without that, since we'd still have to intervene either way, whether Orchestrator aborted or waited forever. But in general, giving it the option of waiting indefinitely in the middle of a dead master recovery may not be the best approach. We did at least confirm that waiting indefinitely wouldn't interfere with other recoveries on other clusters.

I really think the optimal thing to do here would be to check again after X seconds and see if there's a more suitable replica to be promoted. Maybe the initially chosen replica is having issues, and a different one caught up in the meantime. I just didn't think the additional complexity required to implement that was worth the potential benefits. That's not a scenario that comes up very often.

shlomi-noach:
I've made quite a few changes here. Notably:

- Supporting timeout when relay logs coordinates are for some reason not making progress beyond some point.
  I actually have a documented use case (haven't published it in public) where the last statement in a relay log is one which relay log coordinates never reaches.
- Introduced `WaitForSQLThreadUpToDate`, a dedicated function which does what its named. This extracts the logic from `topology_recovery.go` and unifies with `StopSlaveNicely()`
- Retries for instance discovery throughout `WaitForSQLThreadUpToDate`
- Restored `FailMasterPromotionIfSQLThreadNotUpToDate`: some users will still want that. `DelayMasterPromotionIfSQLThreadNotUpToDate` and `FailMasterPromotionIfSQLThreadNotUpToDate` are mutually exclusive.

shlomi-noach (approved):

akshaysuryawanshi (approved):

PR #773, opened 2019-01-13
fix to getPriorityMajorVersionForCandidate on identical counts

### LINKED ISSUE #770: different ouput result of getPriorityMajorVersionForCandidate()
In the function `getPriorityMajorVersionForCandidate()`, the `map` structure: `majorVersionsCount` is used.

The traversal order of the `map` is random in golang.

So, the output result of the function may be different in two calls with the same input.

For example, 4 slaves in a cluster, 

version number as follows: 5.6，5.6，5.7，5.7.

When `getPriorityMajorVersionForCandidate() `called,  the output result may be 5.6 or 5.7.

Is this correct?

Thank you for your precious time.

cc @MOON-CLJ

shlomi-noach:
Fixes https://github.com/github/orchestrator/issues/770

With identical number of appearances for different versions, `orchestrator` would pick an inconsistent value. the fix is that in such scenario it should pick the lower available version.


shlomi-noach:
Also fixed `getPriorityBinlogFormatForCandidate()`, using similar logic

PR #774, opened 2019-01-13
updating Dockerfile: alpine+go

shlomi-noach:
Solves https://github.com/github/orchestrator/issues/769

`orchestrator` now requires `go` >= `1.9`. Current Dockerfile build only pulls `go1.8`.

This PR uses `alpine3.8` in Dockerfile, which then uses `go1.10`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,13 @@ env:
 - MYSQL_USER=root
 
 before_install:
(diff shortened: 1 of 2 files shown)


shlomi-noach:
Docker build success has been overlooked, and is now tested by Travis CI.

Issue #616, opened 2018-09-07
Wrong address (port as a part of IP) when discovering new instance


pznamensky:
Orchestrator (v3.0.12) trying to discover IPv6 host after appending port to address.
How to reproduce:

1. Setup master-slave replication in IPv6 network
2. Use master's IPv6 address in `CHANGE MASTER TO ...`
2. Setup orchestrator
3. Discover topology with the following command:
```
~ # orchestrator -c discover -i slave1 --debug

2018-09-07 17:39:31 INFO starting orchestrator, version: 3.0.12, git commit: d318760701a5649867606f455148e28f4353d288
2018-09-07 17:39:31 INFO Read config: /etc/orchestrator.conf.json
2018-09-07 17:39:31 DEBUG Hostname unresolved yet: slave1
2018-09-07 17:39:31 DEBUG Cache hostname resolve slave1 as slave1
2018-09-07 17:39:31 DEBUG Connected to orchestrator backend: sqlite on /usr/local/orchestrator/orchestrator.sqlite3
2018-09-07 17:39:31 DEBUG Initializing orchestrator
2018-09-07 17:39:31 INFO Connecting to backend :3306: maxConnections: 128, maxIdleConns: 32
2018-09-07 17:39:31 DEBUG Hostname unresolved yet: 2abc::1
2018-09-07 17:39:31 DEBUG Cache hostname resolve 2abc::1 as 2abc::1
2018-09-07 17:39:31 DEBUG Hostname unresolved yet: 2abc::1
2018-09-07 17:39:31 DEBUG Cache hostname resolve 2abc::1 as 2abc::1
```

5. Error occurs:
```
2018-09-07 17:39:39 ERROR dial tcp [2abc::1:3306]:3306: i/o timeout
2018-09-07 17:39:39 ERROR ReadTopologyInstance(2abc::1:3306): dial tcp [2abc::1:3306]:3306: i/o timeout
WARNING: NamedStopwatch.Stop("instance") IsRunning is false
2018-09-07 17:39:39 WARNING  DiscoverInstance(2abc::1:3306) instance is nil in 2.003s (Backend: 0.002s, Instance: 2.001s), error=dial tcp [2abc::1:3306]:3306: i/o timeout
```
As you can see orchestrator trying to connect to `[2abc::1:3306]:3306` instead of `[2abc::1]:3306`

shlomi-noach:
Similar: https://github.com/github/gh-ost/issues/604

[...]

PR #782, opened 2019-01-18
Using net.JoinHostPort to format host:port strings

mialinx:
IPv6 based addrs shoul look like [host]:port instead of host:port

Related issue: https://github.com/github/orchestrator/issues/0123456789

### Description

- [ ] contributed code is using same conventions as original code
- [ ] code is built via `./build.sh`


shlomi-noach:
Thank you. This will unfortunately break a lot of things. Building the correct `[host]:port` is one thing, but there's a myriad of places where we assume the format is `host:port` and not `[host]:port`.

It is very unfortunate, indeed, but the change to `ipv6` would need to be massive.

shlomi-noach:
Hi @mialinx, I see you keep pushing changes. Would you like to perhaps explain what you are doing? Like I wrote earlier, there's a LOT to change if we want to go ipv6 that it's difficult to even find where to begin with.

mialinx:
Hi @shlomi-noach, 

Yep, I'm trying to find all suspicious places like using `':'`, `":"`, `"%s:"` in the code, check if they are related to hostname, port handling and replace them with IPv6 aware code.

As far as I understood, there are two main points:
* Check that all services may be conneсted using IPv6 addr. I have tested using IPv6 for specifing backend MySQL database, RAFT nodes, Zookeper store, and discovering mysql clusters (even `@@hostname` and `show slave hosts` are IPv6 addrs). And it works to me.
* Using `host:port` as identifier for comparing instances. I think it should also work if we carefully change all  occurrences of `host:port`. Anyway for IPv4 and hostnames identifiers should stay the same.

It's very likely that I missed somtheing :(  What other difficulties do you see ?

shlomi-noach:
Your changes are thorough, indeed. What concerns me are all the places where we parse `host:port`. I'm yet to consider all the cases, my focus right now is on a few other high priority issues. Meanwhile I applaud you for this undertaking.

mialinx:
It seems that integration tests flap as they depened on map keys order:
https://github.com/github/orchestrator/pull/785

pznamensky:
@mialinx thanks for your work!
The only reason we don't use orchestrator is IPv6 related issues like this: #616
I hope this PR will be approved.

shlomi-noach:
I completely dropped the ball on this PR. It'll take me time to re0review it.

d3mash:
could you please take a look at this PR?

mainameiz:
We really really need this PR. Please take a look :pray:

shlomi-noach:
I apologize for not giving this attention. I very much appreciate the contribution and your work. It'll take a while before I've fully reviewed this. I'll meanwhile send for production testing in staging environment.

PR #785, opened 2019-01-22
fix flappy integration test

mialinx:
Small maps (of 2 items) are randomized badly,
i.e. their keys has the same order *almost* always, but not always.
This leads to rare test failures in commands that depends on keys order,
such as untag (untag-all-03)

Related issue: https://github.com/github/orchestrator/issues/0123456789

### Description

- [ ] contributed code is using same conventions as original code
- [ ] code is built via `./build.sh`


shlomi-noach (commented):
Thank you. I wonder if this shouldn't be the other way around: that the tests sort the results, and we keep randomized iteration in `GetInstanceKeys()`. I don't feel strongly either way.

mialinx:
If you don't rely on any particular order of GetInstanceKeys(), then this patch breaks nothing.

If you do rely on alphabetic order of GetInstanceKeys() (as it looks like in `expected_output` files and `go/inst/tag_dao.go:Untag` queries), then there is a bug and this patch fixes it.

PR #789, opened 2019-02-05
ChangeMasterCredentials: fixed when server is not a replica

### LINKED ISSUE #788: MakeCoMaster fails, when master does not have any replication settings
If master does not have replication settings, i.e. `slave slave status` is "Empty set", the attempt to make it as co-master with the slave fails with the error:

> ChangeMasterTo: Cannot change master on: master1:3306 because replication is running

As far as I understand it fails in `ChangeMasterCredentials` function, checking whether replication is stopped or not:
```
	if !instance.ReplicationThreadsStopped() {
		return instance, fmt.Errorf("ChangeMasterTo: Cannot change master on: %+v because replication is running", *instanceKey)
	}
```
But replication is neither stopped nor running. I suppose, it is "ReplicationThreadStateNoThread".
PS: this is latest version 3.0.14 issue.

shlomi-noach:
Fixes https://github.com/github/orchestrator/issues/788

This fixes the `ChangeMasterCredentials()` behavior on masters. Behavior was broken by https://github.com/github/orchestrator/pull/767

cc @pasha167

### DIFF
--- a/go/inst/instance_topology.go
+++ b/go/inst/instance_topology.go
@@ -44,6 +44,7 @@ var asciiFillerCharacter = " "
 var tabulatorScharacter = "|"
 
 var countRetries = 5
+var MaxConcurrentReplicaOperations = 5
 
 // getASCIITopologyEntry will get an ascii topology tree rooted at given instance. Ir recursively
 // draws the tree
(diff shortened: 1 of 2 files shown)


ghost:
@shlomi-noach, thank you for quick response. 
I tested that and it did not solve the issue. Could you please double check?
Thanks.

shlomi-noach:
whoops. committed an unrelated change.

shlomi-noach:
@pasha167 please test again now

ghost:
Yes, that works 👍. Thanks.

shlomi-noach:
This PR also introduces a concurrency cap on the number of replicas operated on `relocate-replicas`. There will only be 5 replicas moved at a time.

PR #792, opened 2019-02-06
LeaderURI: self identify, avoid infinite forwarding

shlomi-noach:
Fixes https://github.com/github/orchestrator/issues/696

In https://github.com/github/orchestrator/issues/696, a demoted leader tries to reverse proxy HTTP requests to the leader, but the `LeaderURI` it has is its own, which leads to infinite reverse-proxy to itself.

In this PR:

- A node self-identifies its own URI.
- And can check whether the LeaderURI actually points to itself.
- And avoids an infinite loop in reverse-proxy. 

The change of behavior is:

Instead of an infinite loop, the host will attempt to serve the request. This isn't ideal either: the node is not the leader, but can't figure out who the leader is (yet). A client that attempts to run an operation during this time will get a "read only" error.

Thank you to @mia0x75 @makmanalp for illustrating the problem. Can you please test this PR?

### DIFF
--- a/go/http/raft_reverse_proxy.go
+++ b/go/http/raft_reverse_proxy.go
@@ -25,7 +25,14 @@ func raftReverseProxy(w http.ResponseWriter, r *http.Request, c martini.Context)
 	if orcraft.GetLeader() == "" {
 		return
 	}
-
+	if orcraft.LeaderURI.IsThisLeaderURI() {
+		// Although I'm not the leader, the value I see for LeaderURI is my own.
+		// I'm probably not up-to-date with my raft transaction log and don't have the latest information.
+		// But anyway, obviously not going to redirect to myself.
+		// Gonna return: this isn't ideal, because I'm not really the leader. If the user tries to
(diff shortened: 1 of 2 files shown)


makmanalp:
@shlomi-noach Thanks a billion for being on this so quickly. I'll test the fix first thing today.

makmanalp:
Hi Shlomi, apologies, took me a while to set up our custom orchestrator build, but I now have the docker image with your patch backported onto `v3.0.10` (what we use), tested it yesterday and overnight in our QA environment and so far it looks good. We're letting it sit until Monday until we're comfortable pushing it to production but the issue seems to have been fixed - memory usage is quite stable and restarting or killing the pods doesn't seem to trigger the problem.

shlomi-noach:
Thanks for the update. If all goes well I'll merge it this week.

makmanalp:
@shlomi-noach just deployed it in production, looks great! Thanks again for the speedy fix!

PR #795, opened 2019-02-06
orchestrator-client gsed support for Darwin/BSD

cswingler:
Related issue: https://github.com/github/orchestrator/issues/794

### Description

This PR adds in some logic in `orchestrator-client` to switch over to `gsed` if the system isn't guaranteed to have GNU coreutils installed.


shlomi-noach (commented):
Is Darwin guaranteed to have `gsed` installed?

cswingler:
It unfortunately is not, most people have it installed out of the `gnu-sed` package in Homebrew.

shlomi-noach:
This caused an infinite loop of `sed` function calling itself. Fixed via #848

Issue #796, opened 2019-02-06
orchestrator-client does not handle basic auth appropriately


cswingler:
The way `orchestrator-client` is handling auth in `curl` is dropping the `--basic --user` flags, which appears to be a simple omission.

```
++ get_curl_auth_params
++ local requires_auth=
++ [[ chrisswingler:<redacted> != \: ]]
++ requires_auth='--basic --user chrisswingler:Lech3rY8nymphO'
++ curl --help
++ fgrep -q disallow-username-in-url
++ curl chrisswingler:<redacted> -s --head https://orchestrator.int.sproutsocial.com/api
++ fgrep -q '401 Unauthorized'
++ echo '401 Unauthorized'
++ return
+ local 'curl_auth_params=401 Unauthorized'
+ '[' '401 Unauthorized' == '401 Unauthorized' ']'
+ fail 'Cannot access orchestrator at https://orchestrator.int.sproutsocial.com/api.  Check ORCHESTRATOR_API is configured correctly and orchestrator is running'
+ message='orchestrator-client[22728]: Cannot access orchestrator at https://orchestrator.int.sproutsocial.com/api.  Check ORCHESTRATOR_API is configured correctly and orchestrator is running'
+ echo 'orchestrator-client[22728]: Cannot access orchestrator at https://orchestrator.int.sproutsocial.com/api.  Check ORCHESTRATOR_API is configured correctly and orchestrator is running'
orchestrator-client[22728]: Cannot access orchestrator at https://orchestrator.int.sproutsocial.com/api.  Check ORCHESTRATOR_API is configured correctly and orchestrator is running
+ exit 1
```

PR #797, opened 2019-02-06
Bugfix in orchestrator-client for authentication handling

cswingler:
Related issue: https://github.com/github/orchestrator/issues/796

### Description

Reference the correct variable to set the appropriate flags in `curl` so basic authentication will work.


shlomi-noach (approved):
thank you! Good catch

shlomi-noach:
closes #796

Issue #779, opened 2019-01-16
about DetachReplicaOperation()


yangeagle:
In the function `DetachReplicaOperation`, it run  the command `change master to` to detach a lost replica,
like   `change master to master_log_file='//mysql-bin.000001:4788', master_log_pos=4788`

On the command line, I simulated this scene,  and an error occurred：

```
mysql > stop slave
mysql >change master to master_log_file='//mysql-bin.000001:4788', master_log_pos=4788;
ERROR 1776 (HY000): Parameters MASTER_LOG_FILE, MASTER_LOG_POS, RELAY_LOG_FILE and RELAY_LOG_POS cannot be set when MASTER_AUTO_POSITION is active.
```

**context:**

MySQL 5.7.21   with gtid

Please check it.

@MOON-CLJ

shlomi-noach:
Correct, this won't work with GTID.

You may use `detach-replica-master-host` as a more powerful method.

yangeagle:
The function 
`recoverDeadMaster` call the function `DetachReplicaOperation`, then the error mentioned above will occurr.

The corresponding code for this function `recoverDeadMaster`  is as follows:

```
if promotedReplica != nil && len(lostReplicas) > 0 && config.Config.DetachLostReplicasAfterMasterFailover {
		postponedFunction := func() error {
			//...
			for _, replica := range lostReplicas {
				replica := replica
				inst.DetachReplicaOperation(&replica.Key)
			}
			return nil
		}
		//...
}
```
`detach-replica-master-host` you refer to  is implemented by `DetachReplicaMasterHost`.

Should  `DetachReplicaOperation`  be  replaced by `DetachReplicaMasterHost` ?

shlomi-noach:
Good point. Yes, it should.

shlomi-noach:
I'd like to possibly use a different approach and do `DetachReplica` after all, but in a way that would work with GTID + `auto_position`.

shlomi-noach:
Addressed by https://github.com/github/orchestrator/pull/801

PR #801, opened 2019-02-10
consolidate detach-replica and detach-replica-master-host

shlomi-noach:
Fixes https://github.com/github/orchestrator/issues/779

This PR consolidates `detach-replica` and `detach-replica-master-host`. It makes `detach-replica` behave exactly like `detach-replica-master-host`. This makes `detach-replica` safe to work with GTIDs. See #779 for more details.


ggunson:
What specifically does `detach-replica-master-host` do that `detach-replica` doesn't?

shlomi-noach:
`detach-replica-master-host` changes the master hostname from `some.host.name.com` to `//some.host.name.com`.

`detach-replica` (before this PR) changes `Master_log_file` from `mysql-bin.0001234` to `//mysql-bin.0001234:29043857`. This worked well for non-GTID setups, but for `AUTO_POSITION=1` it would error.

The advantage of the `detach-replica` approach is that the replica still points to the same master, just unable to replicate.

PR #804, opened 2019-02-12
fix executed_gtid_set missing from instances without binlogs

### LINKED ISSUE #803: Can't relocate a slave using GTIDs
Hello,

I've just migrated to latest orchestrator version (3.0.14).

My DB topology is fairly simple 1 active master / 1 passive master and multiples readonly slaves with no binlogs. Percona (Oracle) GTIDs are enabled and in use (auto master position).

When trying to relocate slaves from one master to the other one by using this : 
`orchestrator-client -c can-replicate-from-gtid -i slave1 -d master2`
I got this error : 
`Instance master2 has purged GTID entries not found on slave1`

I checked on both servers and GTID were correct and no subset was missing.

After digging a while, I found that executed_gtid_set information is missing from orchestrator internal DB for MySQL instances w/o binlogs.

The check is done here : https://github.com/github/orchestrator/blob/7bef26f042aafbd956daeaede0cd4aab2ba46e65/go/inst/instance_topology.go#L558

Executed_Gtid_Set is fetched here : 
https://github.com/github/orchestrator/blob/7bef26f042aafbd956daeaede0cd4aab2ba46e65/go/inst/instance_dao.go#L398

I made a fix and moved this into the select that gets other global informations (gtid_purged etc ..) : 
https://github.com/fuyar/orchestrator/commit/02153a47a6c08f317dbbc7242bca8ce4ef58b0e9

I dont know if this is OK and I should push a PR ?

WDYT ?

Thanks !

fuyar:
Fixes #803 

Executed_Gtid_Set should be fetched for any MySQL instance and should not be restricted for "masters" (with binlog ON).

### DIFF
--- a/go/inst/instance_dao.go
+++ b/go/inst/instance_dao.go
@@ -395,7 +395,6 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
 					var err error
 					instance.SelfBinlogCoordinates.LogFile = m.GetString("File")
 					instance.SelfBinlogCoordinates.LogPos = m.GetInt64("Position")
-					instance.ExecutedGtidSet = m.GetStringD("Executed_Gtid_Set", "")
 					return err
 				})
 			}()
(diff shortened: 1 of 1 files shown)


shlomi-noach (approved):
looks good, thank you! I'll send to testing and then merge.

shlomi-noach:
This works correctly.
The downside is that we waste network traffic: `show master status` returns `Executed_Gtid_Set` anyway, and we ignore it where we do have binary logs. `Executed_Gtid_Set` may be a large (long) value for clusters suffering from multiple failovers.

Still, this is more correct to use. I'll merge it as-is, and consider a future optimization that only reads `@@global.gtid_executed` if no binary logs are available.

PR #805, opened 2019-02-13
adding bash to final container so examples work

### LINKED ISSUE #783: bash missing from Dockerfile
Dockerfile is missing /bin/bash and results in errors during recovery steps

Looks like this broke previously in #222 but then broke again when the Dockerfile was updated to use multiple stages.

I'm still trying to wrap my head around it, will try and provide more details as I start digging.

anthonyneto:
Examples from the documentation use echo however echo cannot be used without bash. resolves #783 

Whenever an event happens that uses a hook to shell out I would get this error.


shlomi-noach (approved):
Thank you! Good catch

Dockerfile
```diff
@@ -37,7 +37,9 @@ RUN rsync -av $(find /tmp/orchestrator-release -type d -name orchestrator -maxde
 RUN rsync -av $(find /tmp/orchestrator-release -type d -name orchestrator-cli -maxdepth 2)/ /
 RUN cp /usr/local/orchestrator/orchestrator-sample-sqlite.conf.json /etc/orchestrator.conf.json
 
-FROM alpine:3.6
+FROM alpine:3.8
+
+RUN apk add --no-cache bash
```
shlomi-noach, line 42: wondering: why `--no-cache` for `bash`?

anthonyneto: This avoids the need to use --update and remove /var/cache/apk/* when done installing packages.

PR #807, opened 2019-02-18
Workaround to bug 83713: GTID, MTR and relay log corruption

shlomi-noach:
We have been bitten by what appears to be https://bugs.mysql.com/bug.php?id=83713

In this PR `orchestrator` responds to `Error 1201` by issuing a sequence of operations which are expected to mitigate the error.

We are still to reliably reproduce the error.


shlomi-noach:
While unable to reproduce the bug, this has been tested in production to at least not cause damage 😆  for almost two weeks. Merging.

PR #815, opened 2019-02-25
Add the DiscoveryIgnoreReplicaHostnameFilters in the sample configuration.

jfg956:
Add the DiscoveryIgnoreReplicaHostnameFilters in the sample configuration.

https://github.com/github/orchestrator/issues/814

### Description

Add the DiscoveryIgnoreReplicaHostnameFilters in the sample configuration.

### DIFF
--- a/conf/orchestrator-sample-sqlite.conf.json
+++ b/conf/orchestrator-sample-sqlite.conf.json
@@ -15,6 +15,10 @@
   "DefaultInstancePort": 3306,
   "DiscoverByShowSlaveHosts": true,
   "InstancePollSeconds": 5,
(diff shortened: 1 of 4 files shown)


shlomi-noach (changes requested):
@jfg956 thank you! It's noteworthy that the value is actually a regular expression. A fixed hostname would work, but so will `.*[.]datacenter3[.]example[.]com`. Perhaps we should also provide a regexp as example?

jfg956:
@shlomi-noach: I made explicit that DiscoveryIgnoreReplicaHostnameFilters can be a regexp.
Hopefully, this will do.  Cheers, JFG

shlomi-noach (approved):
thank you!

PR #839, opened 2019-03-12
relocate-replicas: sanity check to avoid invalid circular replication

shlomi-noach:
Fixes https://github.com/github/orchestrator/issues/838

The specific `relocate-replicas` command (e.g. `relocate-replicas -i srv1 -d srv2`) will pre-verify that none of the replicas of instance `srv1` is an ancestor of `srv2`.

The check is based on `ancestry_uuid` which is based on the `server_uuid` variable, which is available `5.6` and onwards.

### DIFF
--- a/go/inst/instance.go
+++ b/go/inst/instance.go
@@ -336,6 +336,16 @@ func (this *Instance) IsMasterOf(replica *Instance) bool {
 	return replica.IsReplicaOf(this)
 }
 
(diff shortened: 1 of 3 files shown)


shlomi-noach:
Deploys and works well. cc @tomkrouper

shlomi-noach:
The fix applies to the `relocate` command, too.

PR #843, opened 2019-03-19
Modify bulk-promotion-rules api call to return the promotion rule expiry timestamp

sjmudd:
### Description

See: https://github.com/github/orchestrator/issues/842 bulk promotion rule information does not include expiry information.


shlomi-noach (approved):
LGTM

PR #849, opened 2019-03-24
Fix MySQL backend TLS/SSL configuration.

burnison:
Setting `MySQLOrchestratorUseMutualTLS=false` prevented _any_ SSL
connection to the backend server, as it guarded the call to
`SetupMySQLOrchestratorTLS`. This behaviour is inconsistent with the
documentation. To keep full backwards compatibility but also correct
this bug, a value set in `MySQLOrchestratorSSLCAFile` will now result in
connections that use SSL.

Further, unlike the topology code path, `MySQLOrchestratorUseMutualTLS`
was not referenced within `SetupMySQLOrchestratorTLS`. As such, all
connections required a (valid) client key pair to be provided,
preventing the application from starting. This change unifies the code
paths.

Related issue: https://github.com/github/orchestrator/issues/847

- [X] code is formatted via `gofmt` (please avoid `goimports`)
- [X] code is built via `./build.sh`
- [X] code is tested via `go test ./go/...`


go/db/tls.go
```diff
@@ -90,53 +87,66 @@ func requiresTLS(host string, port int, mysql_uri string) bool {
...
-		}
-		topologyTLSConfigured = true
+func SetupMySQLOrchestratorTLS(uri string) (string, error) {
+	c := connectionConfig{
+		uri:              uri,
+		caCertFile:       config.Config.MySQLOrchestratorSSLCAFile,
+		useClientAuth:    config.Config.MySQLOrchestratorUseMutualTLS,
+		clientCertFile:   config.Config.MySQLOrchestratorSSLCertFile,
+		clientKeyFile:    config.Config.MySQLOrchestratorSSLPrivateKeyFile,
+		verifyClientCert: true, // FIXME: Should this have been SSLSkipVerify?
```
burnison, line 110: This is to keep existing behaviour. I wasn't completely certain if this is a bug, as it's not consistent with `SetupMySQLTopologyTLS`. In this case, I left the original value.

go/db/tls.go
```diff
@@ -35,12 +35,9 @@ import (
...
 
-// Track if a TLS has already been configured for topology
-var topologyTLSConfigured bool = false
-
-// Track if a TLS has already been configured for Orchestrator
-var orchestratorTLSConfigured bool = false
-
+// This cache exists to prevent re-registering tls.Conf instances while also
+// allowing them to be interrogated in tests.
+var tlsConfigCache *cache.Cache = cache.New(cache.NoExpiration, time.Second)
```
burnison, line 40: Being not _terribly_ familiar with Golang, this seemed to me as the best option to be able to inspect the `tls.Config` objects generated in the methods in the tests. I figure that the ability to test the code outweighs the downside of a couple structs lingering in memory. I'm happy to change this if there's a better way.

PR #850, opened 2019-03-24
Adding locate-gtid-errant command

shlomi-noach:
`locate-gtid-errant` reports the names of the binary logs containing errant GTID for a given instance.

Example usage:

```
orchestrator-client -c locate-gtid-errant -i my-host:3306
mysql-bin.000001
mysql-bin.000004
mysql-bin.000006
```

Execution returns with an error when:

- there is no errant GTID
- part or all of the errant GTID is purged


shlomi-noach:
This PR also adds the `which-gtid-errant` command, which outputs the errant GTID, if such exists.

PR #857, opened 2019-04-03
Add DataCenterLookupValues setting

sjmudd:
See: https://github.com/github/orchestrator/issues/856

This PR provides a patch to add a new configuration setting: `DataCenterLookupValues` which would be applied after applying `DataCenterPattern` on the instance hostname.

- [ ] code is built via `./build.sh`       # this currently fails as you only support up to go 1.10. I'm using go 1.12.

Please let me know if this looks reasonable.


sjmudd:
I'll make a PR to support a higher version of Go.

sjmudd:
Perhaps I should provide some context:

The term "datacentre" used in orchestrator is good but in some cases you may want to group more than one datacentre (located close by) into a *single* _availability zone_. Orchestrator doesn't have a way to reference availability zones as such so an easy way to resolve this is to map parts of a hostname which include a datacentre into a AZ definition, after which orchestrator colouring in the GUI will treat the servers as being in the same "area". This looks better than seeing them mixed when logically they should be considered together.

This change handles the mangling based on the server's hostname which is simpler than doing a more complex `if(...)` type SQL query against each instance being monitored. Doing this against each instance being monitored is an extra query and that increases latency of the time to poll a server so is not ideal if the number of servers being monitored is large.

PR #858, opened 2019-04-05
Create Post Take-Master Processes Hook

daniel-2647:
### Description

This PR creates a hook which would only be called upon a successful Take-Master event. This would enable custom scripts to be called at those times when a post failover hook is not called, i.e. When we do a graceful failover of an intermediary master (See https://github.com/github/orchestrator/issues/799 for full details)

This PR introduces 2 new config options:

1. PostTakeMasterProcessesEnabled : true/false
2. PostTakeMasterProcesses : "some PostTakeMasterHook here"

There is an issue created for this a few months back with details on the reasoning behind this PR: https://github.com/github/orchestrator/issues/799


daniel-2647:
Closing this pull request to create a new one with some corrections https://github.com/github/orchestrator/pull/859

PR #861, opened 2019-04-11
Update support for go 1.12 (also triggers some file reformatting)

sjmudd:
Orchestrator build fails on go1.12 so adjust.

```
[user@host ~/src/orchestrator/src/github.com/github/orchestrator]$ go version
go version go1.12 darwin/amd64
[user@host ~/src/orchestrator/src/github.com/github/orchestrator]$ ./build.sh -b
Build only; no packaging
go version must be between 1.8 and 1.10
```

This also triggers some file reformatting (due to `go fmt` changes) which I've included.
There are no functional changes in the code.


sjmudd:
Related: https://github.com/github/orchestrator/pull/857.

shlomi-noach (commented):
@sjmudd thank you. Can you please also remove go1.8, and update the `.travis.yml` file?

shlomi-noach:
Because `go1.12` has different formatting than previous versions, `go1.12` will be the _minimal_ supported version. I don't wish to support both `1.11` and `1.12` with differing formatting.

shlomi-noach:
Added:

- using go1.12 on Travis
- using go1.12 in Docker
- Updated `script/ensure-go-installed`

PR #863, opened 2019-04-12
add curl and jq for orchestrator-client

marcosvm:
Related issue: https://github.com/github/orchestrator/issues/862

### Description

This pull request adds `curl` and `jq` packages allowing to make runs from inside a docker container like:

```
docker run --rm -ti orchestrator:latest  /usr/local/orchestrator/resources/bin/orchestrator-client
```

which can be very handy.


shlomi-noach (approved):
thank you!

PR #877, opened 2019-05-08
Support countLostReplicas in failover hooks

shlomi-noach:
`{countLostReplicas}` is now reported to `PostFailoverProcesses` and `PostUnsuccessfulFailoverProcesses`


shlomi-noach:
works as expected

IT-xiaoge:
Under what conditions will Lost replicas？Now I will also lose replicas when failing over.

shlomi-noach:
> Under what conditions will Lost replicas？Now I will also lose replicas when failing over.


Explained in https://github.com/github/orchestrator/blob/master/docs/topology-recovery.md#discussion-recovering-a-dead-master

pospelov-v:
Hi. Command line text replacement its it `countLostReplicas`. Which Environment variable? `ORC_COUNT_LOST_REPLICAS`?

PR #878, opened 2019-05-08
Add structure warning to `replication-analysis` when all masters are read_only

jfudally:
Add structure warning to `replication-analysis` when all master nodes are read_only

Related issue: https://github.com/github/orchestrator/issues/865

### Description

This PR adds `NoWriteableMasterStructureWarning` to `replication-analysis` during scenarios when every master node in the cluster is `read_only`, omitting intermediate masters.

```
# Cluster with cyclical replication between two masters, and an intermediate master
bash-4.4# ./orchestrator -c topology -i mysql_1
+ mysql_1:3306     [0s,ok,5.7.25-log,rw,ROW]
  + mysql_2:3306   [0s,ok,5.7.25-log,ro,ROW]
    + mysql_3:3306 [0s,ok,5.7.25-log,ro,ROW]
+ mysql_4:3306     [0s,ok,5.7.25-log,rw,ROW]
  + mysql_5:3306   [0s,ok,5.7.25-log,ro,ROW]

# No warnings for the intermediate master
bash-4.4# ./orchestrator -c replication-analysis -i mysql_1

# mysql_1 master is now R/O
bash-4.4# ./orchestrator -c topology -i mysql_1
+ mysql_1:3306     [0s,ok,5.7.25-log,ro,ROW]
  + mysql_2:3306   [0s,ok,5.7.25-log,ro,ROW]
    + mysql_3:3306 [0s,ok,5.7.25-log,ro,ROW]
+ mysql_4:3306     [0s,ok,5.7.25-log,rw,ROW]
  + mysql_5:3306   [0s,ok,5.7.25-log,ro,ROW]

# No warnings from 1 of 2 co-masters being R/O
bash-4.4# ./orchestrator -c replication-analysis -i mysql_1

# Both co-masters mysql_1 & mysql_3 are now R/O
bash-4.4# ./orchestrator -c topology -i mysql_1
+ mysql_1:3306     [0s,ok,5.7.25-log,ro,ROW]
  + mysql_2:3306   [0s,ok,5.7.25-log,ro,ROW]
    + mysql_3:3306 [0s,ok,5.7.25-log,ro,ROW]
+ mysql_4:3306     [0s,ok,5.7.25-log,ro,ROW]
  + mysql_5:3306   [0s,ok,5.7.25-log,ro,ROW]

🎉 
bash-4.4# ./orchestrator -c replication-analysis -i mysql_1
mysql_4:3306 (cluster mysql_1:3306): NoWriteableMasterStructureWarning
```


go/inst/analysis.go
```diff
@@ -64,6 +64,7 @@ const (
 	DifferentGTIDModesStructureWarning                                       = "DifferentGTIDModesStructureWarning"
 	ErrantGTIDStructureWarning                                               = "ErrantGTIDStructureWarning"
 	NoFailoverSupportStructureWarning                                        = "NoFailoverSupportStructureWarning"
+	NoWriteableNodesWarning                                                  = "NoWriteableNodesStructureWarning"
```
shlomi-noach: We should be concerned with non-writable master, for sure. The rest of the nodes are practically irrelevant (that is, we expect them to be read-only anyhow).

shlomi-noach: I'd name this: `NonWritableMasterWarning`

shlomi-noach:
To elaborate, on a normal topology, we only expect the master to be writable. Right now we're interested in a scenario where the master is read-only. We normally expect (and hope) for replicas to be read-only, too, but I've seen cases out there where intermediate master could be taking writes. Since we're aiming at the global audience, my suggestion is to let go of looking into the replicas.

So we should only care if we have a read-only master. 

Another thing is that we want to check is the case where we have co-masters. In a co-master (circular master-master) setup, it's OK if one of the co-masters is read-only. It's also OK (though generally discouraged) if both co-masters are writable. It's not OK if both are read-only...

go/inst/analysis.go
jfudally: @shlomi-noach Make sense, I've updated to warn if the current master is `read-only`.

jfudally:
I'l take a look at the co-master scenarios as well

jfudally:
@shlomi-noach I've updated the PR body.  I've tested new logic to omit warnings on `intermediate masters` being R/O, and if at least 1 `co-master` is still R/W.

go/inst/analysis_dao.go
```diff
@@ -90,15 +91,17 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
 		    SELECT
 		        master_instance.hostname,
 		        master_instance.port,
-						MIN(master_instance.data_center) AS data_center,
-						MIN(master_instance.physical_environment) AS physical_environment,
+			master_instance.read_only AS read_only,
```
shlomi-noach: 👍

go/inst/analysis_dao.go
```diff
@@ -90,15 +91,17 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
...
+			SUM(master_instance.is_co_master) AS co_master_count,
```
shlomi-noach: this will be either `0` or `1`, not what you'd expect. See later discussion.

go/inst/analysis_dao.go
```diff
@@ -296,6 +301,10 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
 		a.CountDelayedReplicas = m.GetUint("count_delayed_replicas")
 		a.CountLaggingReplicas = m.GetUint("count_lagging_replicas")
 
+		if a.IsCoMaster && m.GetUint("read_only") == 1 {
```
shlomi-noach: Please read `m.GetUint("read_only")` into a variable first. In fact, it's really interesting information, make that a member in the `ReplicationAnalysis` struct

go/inst/analysis_dao.go
```diff
@@ -296,6 +301,10 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
...
+			coMasterROCount++
```
shlomi-noach: This look can iterate over multiple clusters. You only want to count the number of co masters per cluster. So this trick of `coMasterROCount++` won't work. 

A cluster with two co-masters appears in `orchestrator` as a single cluster, not as two clusters. One of the two co-masters is deemed to be the "primary" master.  If one is read-only and another is writable, it's the writable. If both are read-only or both are writable, I believe the choice is alphabetic.

So, as I'm writing this, I'm rethinking our earlier discussion. if the master is a `co_master` and is `read_only`, then the other co-master is also `read_only` (or else IT would have been the primary master!). This simplifies things and you can get rid of `coMasterROCount`.

go/inst/analysis_dao.go
```diff
@@ -477,6 +486,14 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
 			if a.MaxReplicaGTIDErrant != "" {
 				a.StructureAnalysis = append(a.StructureAnalysis, ErrantGTIDStructureWarning)
 			}
+
+			if a.IsMaster && m.GetUint("read_only") == 1 && m.GetString("master_host") == "" {
```
shlomi-noach: `"master_host"` is already read in `a.AnalyzedInstanceMasterKey`. You can ask if `a.AnalyzedInstanceMasterKey.IsValid()` (== non empty)

shlomi-noach: Given earlier discussion, I think we can skip the check about `master_host` altogether.

go/inst/analysis_dao.go
```diff
@@ -477,6 +486,14 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
...
+				a.StructureAnalysis = append(a.StructureAnalysis, NoWriteableMasterStructureWarning)
+			}
+			if a.IsCoMaster && coMasterROCount == m.GetInt("co_master_count") {
```
shlomi-noach: Given earlier discussion, I think we can skip this block altogether.

shlomi-noach (changes requested):

go/inst/analysis_dao.go
```diff
@@ -90,6 +91,7 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
 		    SELECT
 		        master_instance.hostname,
 		        master_instance.port,
+            master_instance.read_only AS read_only,
```
shlomi-noach: inconsistent indentation here

shlomi-noach (approved):
This looks good to me! Let's deploy to our staging environment and experiment

jfudally:
Sounds good...as soon as Travis decides to work I can deploy to our test environment.

shlomi-noach:
Travis is used for our external facing builds; our internal builds are all that is needed (and at this time we will need to kick the build manually).

shlomi-noach:
Good to go! Identifies a read-only master on a single-master topology. Future PR to handle co-masters.

PR #883, opened 2019-05-10
Update documentation for MySQLTopologyCredentialsConfigFile

EagleEyeJohn:
Related issue: https://github.com/github/orchestrator/issues/882

### Description

This PR updates documentation and code comment for MySQLTopologyCredentialsConfigFile so that it is clear what format this file should be.


shlomi-noach (changes requested):
Thank you! Actually, "You may choose to use plaintext credentials:" and the following `json` file are correct; they should just be in a different paragraph. It was misleading to put "You may choose to use plaintext credentials:" in the same paragraph as "`MySQLTopologyCredentialsConfigFile` follows similar rules as `MySQLOrchestratorCredentialsConfigFile`. "

My suggestion is to edit as follows:

---

`MySQLTopologyCredentialsConfigFile` follows similar rules as `MySQLOrchestratorCredentialsConfigFile`, for example:

```
[client]
user=orchestrator
password=orc_topology_password
```

Or, you may choose to use plaintext credentials:

```json
{
  "MySQLTopologyPassword": "orc_topology_password",
}
```

shlomi-noach:
Respectful ping

shlomi-noach (approved):
thank you

PR #905, opened 2019-05-30
Fix comment name that doesn't match actual function name

sjmudd:
Tiny weenie patch in the comment to make the function names match. `golint` and `go vet` sometimes get upset about these things.


sjmudd:
FWIW `golint` seems to find a few other things. I just was looking at this function. I know that sometimes these warnings might be pointless but often they can be helpful.
For now the single patch "helps", if only a tiny, weenie bit...

shlomi-noach (approved):

PR #912, opened 2019-06-14
Provide snapshot-topologies support in orchestrator-client (and via api)

sjmudd:
`orchestrator` has a `--snapshot-topologies` option not available via `orchestrator-client`.

This patch provides:
* support in `orchestrator-client` to support `--snapshot-topologies`
* support in the api interface to handle a `/api/snapshot-topologies` call which will do the same thing.

Output of `orchestrator-client` looks like:

```
$ orchestrator-client -c snapshot-topologies
Took 2.125752ms
```

Output of a call to the orchestrator api (e.g. https://orchestrator.example.com/api/snapshot-topologies`) shows:

```
{"Code":"OK","Message":"Topology Snapshot completed","Details":"Took 2.058672ms"}
```

- [x] code is formatted via `gofmt` (please avoid `goimports`)     # note: using go 1.12 formatting

### DIFF
--- a/go/http/api.go
+++ b/go/http/api.go
@@ -1665,11 +1665,11 @@ func (this *HttpAPI) asciiTopology(params martini.Params, r render.Render, req *
 func (this *HttpAPI) SnapshotTopologies(params martini.Params, r render.Render, req *http.Request) {
 	start := time.Now()
 	if err := inst.SnapshotTopologies(); err != nil {
-		Respond(r, &APIResponse{Code: ERROR, Message: fmt.Sprintf("%+v", err)})
+		Respond(r, &APIResponse{Code: ERROR, Message: fmt.Sprintf("%+v", err), Details: fmt.Sprintf("Took %v", time.Since(start))})
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


shlomi-noach:
`snapshot-topologies` was written a long time ago and I haven't been practicing it in a while. The idea was to have some "git history" for the topologies; I think the feature needs to be revisited and get some serious testing.

shlomi-noach (approved):

sjmudd:
Well Shlomi, you set it up here. It's getting testing... but to be honest I'm not sure if the content it generates is being used. Were you wanting to get rid of it that would be fine I think.

Having said that the PR was because I'm moving all access to orchestrator to use the http interface so that I can switch the orchestrator clusters I talk to more easily (rather than use the config file). So moving to `orchestrator-client` means that any existing scripts etc collecting data which used the `orchestrator` binary need to be adapted.  This was missing.

This patch has been running since I made the PR and seems to be working fine.

shlomi-noach (approved):

PR #915, opened 2019-06-26
optional systemd-based build packaging

shlomi-noach:
A resubmission of https://github.com/github/orchestrator/pull/653 by @mateusduboli (for irrelevant technical reasons), but in this PR we still use the `root` user; while agreed this is not ideal, I'm concerned with backwards compatibility -- how users who already assume the `root` users would be impacted if they install a newer package which uses a different account. So pushing that decision till later.

cc @sjmudd


shlomi-noach:
Gonna merge this as-is. With this PR (again, contributed by @mateusduboli), one would:

```
./build.sh [-i sysv]
```
or
```
./build.sh -i systemd
```

shlomi-noach:
fat fingers. A followup PR will dockerize the build process to generate both options.

PR #920, opened 2019-07-01
Add GracefulMasterTakeoverSkipSiblingTakeover option

jbransen:
### A Pull Request should be associated with an Issue.

Related issue: https://github.com/github/orchestrator/issues/919

### Description

This PR adds the GracefulMasterTakeoverSkipSiblingTakeover, which is `false` by default, but when turned to `true` will not move siblings when doing a graceful master switchover. Read the [related issue](https://github.com/github/orchestrator/issues/919) for details on the use case.


shlomi-noach:

shlomi-noach (commented):
Almost year-old, slight lag. I'm regrettably unsure how I wish to proceed with this. `orchestrator` does not support multisource replication, and this is a hack to make it ignore multisource replication to some extent. The solution of `GracefulMasterTakeoverSkipSiblingTakeover` seems very specific and I fear that by letting it in, there will come requests for other flags that are likewise very specific to people's very specific use cases. 
I'm sorry that I will meanwhile not merge this PR.

PR #952, opened 2019-08-05
Include port in DiscoveryIgnoreReplicaHostnameFilters regexp matching

dougfales:
Related issue: https://github.com/github/orchestrator/issues/923

This PR simply extends the hostname matching for the `DiscoveryIgnoreReplicaHostnameFilters` config setting to also include a server's port. By allowing the match on port number as well as hostname, you can prevent any replica instances (or transient gh-ost processes) which connect on [a secondary port](https://www.percona.com/doc/percona-server/LATEST/performance/threadpool.html#extra_port) from showing up in Orchestrator by using a config setting like this:

```
  "DiscoveryIgnoreReplicaHostnameFilters": [ ".*:3307" ]
```

Notes:
* See [the related issue](https://github.com/github/orchestrator/issues/923) for an explanation of how this affects us with our gh-ost usage.
* You will need to tell Orchestrator to *Forget* an already-discovered instance. This is not a change in functionality, just something I noticed during my testing.

### DIFF
--- a/conf/orchestrator-sample-sqlite.conf.json
+++ b/conf/orchestrator-sample-sqlite.conf.json
@@ -18,7 +18,8 @@
   "InstancePollSeconds": 5,
   "DiscoveryIgnoreReplicaHostnameFilters": [
     "a_host_i_want_to_ignore[.]example[.]com",
-    ".*[.]ignore_all_hosts_from_this_domain[.]example[.]com"
+    ".*[.]ignore_all_hosts_from_this_domain[.]example[.]com",
+    "a_host_with_extra_port_i_want_to_ignore[.]example[.]com:3307"
   ],
   "UnseenInstanceForgetHours": 240,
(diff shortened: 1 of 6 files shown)


conf/orchestrator-sample-sqlite.conf.json
```diff
@@ -18,7 +18,8 @@
   "InstancePollSeconds": 5,
   "DiscoveryIgnoreReplicaHostnameFilters": [
     "a_host_i_want_to_ignore[.]example[.]com",
-    ".*[.]ignore_all_hosts_from_this_domain[.]example[.]com"
+    ".*[.]ignore_all_hosts_from_this_domain[.]example[.]com",
+    "a_host_with_extra_port_i_want_to_ignore[.]example[.]com:3307"
```
shlomi-noach, line 22: Thanks for adding an example!

shlomi-noach (approved):
Thank you

PR #959, opened 2019-08-15
add args commandPayload for manual failover/takeover api/cmd then pass to hooks

MOON-CLJ:

Related issue: https://github.com/github/orchestrator/issues/955

cc @yangeagle


shlomi-noach (commented):
This will only work correctly if invoked from the orchestrator cli; using the API or `orchestrator-client` this has no effect. Correct?
I wonder if we can make this consistent across all methods.

MOON-CLJ:
> using the API or orchestrator-client this has no effect. Correct?

yes, i know。 for orchestrator-client, not works。
orchestrator-client need one character flag for the new commandPayload args。what's your suggestion, which character one should we use? :) 

"-C" ?

MOON-CLJ:
@shlomi-noach 
 
> using the API or orchestrator-client this has no effect. Correct?

For API, i should works，see https://github.com/github/orchestrator/pull/959/files#diff-fbdc8cde28a71961e53c2e83e2483c9a

for now just orchestrator-client not works。

MOON-CLJ:
> orchestrator-client need one character flag for the new commandPayload args。what's your suggestion, which character one should we use? :)

~~or we can use getopt instead of getopts，but introduce some minor change of the usage 。~~

ignore, it's a bad idea. getopt's behavior is not consistent between macos and linux.

MOON-CLJ:
@shlomi-noach i add -C --command-payload in orchestrator-client, pls review。

i also test in local, when "orchestrator-client  .... --command-payload username:clj,a:b ....."，
i can got "ORC_COMMAND_PAYLOAD=username:clj,a:b" in environment variable。

shlomi-noach:

MOON-CLJ:
thx

PR #960, opened 2019-08-19
Write buffer metrics

luisyonaldo:
Related issues: https://github.com/github/orchestrator/issues/113
Related PRs: https://github.com/github/orchestrator/pull/676, https://github.com/github/orchestrator/pull/682

### Description

This PR adds metrics related to the logic behind write buffer instances. The goal is to try to corroborate that what we assume is what is really happening and improve how we can tune config variables like `InstanceWriteBufferSize` and `InstanceFlushIntervalMilliseconds`.

While testing this PR and looking at the metrics some things came up.

- The metrics showed orchestrator was constantly flushing instances even when `InstanceWriteBufferSize` was not reached and without waiting at least `InstanceFlushIntervalMilliseconds`

Here is a small sample of the metrics `api/write-buffer-metrics-raw/60`:
```json
    {
        "Timestamp": "2019-08-16T17:11:58.394757488+02:00",
        "WaitLatency": 964,
        "FlushLatency": 1524,
        "EnqueuedInstances": 5
    },
    {
        "Timestamp": "2019-08-16T17:11:58.402999622+02:00",
        "WaitLatency": 1018,
        "FlushLatency": 1206,
        "EnqueuedInstances": 4
    },
    {
        "Timestamp": "2019-08-16T17:11:58.416573318+02:00",
        "WaitLatency": 1720,
        "FlushLatency": 1978,
        "EnqueuedInstances": 5
    },
...
```
Latency is showed in milliseconds and our config is `InstanceFlushIntervalMilliseconds: 50` and `InstanceWriteBufferSize: 1200` for our primary cluster.

Here we can see how we are flushing almost instantly (wait latency is almost non-existent) and with just a few instances in the buffer. Which is the opposite of what we expected: the buffer should be flushed if `InstanceFlushIntervalMilliseconds` has passed or if `InstanceWriteBufferSize` if reached.

After this, we added some logging and debugged dipper and we found the following:

- There were many discoveries goroutines blocked/waiting to signal to flush

https://github.com/github/orchestrator/blob/master/go/inst/instance_dao.go#L2487

There are `DiscoveryMaxConcurrency` discovery goroutines filling the buffer and as soon as the buffer reached the max size, the goroutines try to send a message to `forceFlushInstanceWriteBuffer` in order to signal the "flushing" goroutine that there's work (flush the buffer). But only one goroutine can send the message and the other goroutines are blocked (non-buffered channel) and waiting to be able to send the signal. As soon as the flushing goroutine finish then the next discovery goroutine is ready to send a message and so on. Which caused the weird behavior we saw on the metrics.
When the buffer reaches its size then the flushing goroutine gets called (which is good) but as soon as it finishes flushing it has many more discoveries goroutines waiting to signal to flush even when there is no need anymore (the buffer has just been flushed).

- Also every time we call the "flushing" goroutine, it flushes half of the buffer only

This is due to how the buffer is being looped over:
https://github.com/github/orchestrator/blob/master/go/inst/instance_dao.go#L2501

```go
for i := 0; i < len(instanceWriteBuffer); i++ {
    upd := <-instanceWriteBuffer
...
}
```
On every loop iteration `len(instanceWriteBuffer)` is called and one instance is flushed out of the buffer until we had flushed half of the buffer and `i` is bigger than the current length of the buffer, e.g.:
```
i=0, len(instanceWriteBuffer)=10
i=1, len(instanceWriteBuffer)=9
i=2, len(instanceWriteBuffer)=8
....
i=5, len(instanceWriteBuffer)=5
exit the loop,  i < len(instanceWriteBuffer) is not valid anymore
```
This is fixed by changing the for loop to:
```go
for len(instanceWriteBuffer) > 0 {..}
```
and then on every flush, we stop only when there are no more instances in the flush. But this will cause a weird behavior also. Since many discovery goroutines can [enqueue](https://github.com/github/orchestrator/blob/master/go/inst/instance_dao.go#L2483) an instance into the buffer while we are flushing. Thus, we could end flushing more than the expected `InstanceWriteBufferSize`.
In our case there were times where it flushed around 2x 3x times `InstanceWriteBufferSize`.

Solutions:
This [commit](https://github.com/github/orchestrator/commit/dce8f197c2c2eededa51fbeaa5f6aa909d0ea4f4) fixes both issues.
- To avoid blocking a discovery goroutine on sending a signal to flush, when flush is already ongoing, we do a non-blocking send, so just the one goroutine will send the signal and the others will continue and not be blocked there. If `forceFlushInstanceWriteBuffer` channel is not ready to receive a signal is because the "flushing" goroutine is running at the moment (it was triggered already).
```go
select {
    case forceFlushInstanceWriteBuffer <- true:
    default:
}
```
- We flush the buffer untill a max of `InstanceWriteBufferSize`
```go
for i := 0; i < config.Config.InstanceWriteBufferSize && len(instanceWriteBuffer) > 0; i++ {..}
```

Note: open to discussions about the solutions. If preferred then I can remove the solution for this PR and discuss the fixes on a different issue.


luisyonaldo:
[Here](https://gist.github.com/luisyonaldo/db3a5a98f287a98132c040b8e7adf621) are the metrics after the fixes.

We can see how now the behavior match what we expect.

`WaitLatency` is on average around `InstanceFlushIntervalMilliseconds: 50` except when we flush due to max size `InstanceWriteBufferSize: 1200` on the buffer has been reached.

Time to time we see big increases in the `WriteLatency` that produces an unusual `WaitLatency`. We are inspecting the reason behind this.

luisyonaldo:
We have been running this from more than a month in our systems. Any comment on this PR?

shlomi-noach:

shlomi-noach:
again, sorry for the delay. Hope to look into this soon.

go/inst/instance_dao.go
```diff
@@ -2492,9 +2496,14 @@ var forceFlushInstanceWriteBuffer = make(chan bool)
...
-		// Signal the "flushing" gorouting that there's work.
+		// Signal the "flushing" goroutine that there's work.
 		// We prefer doing all bulk flushes from one goroutine.
-		forceFlushInstanceWriteBuffer <- true
+		// Non blocking send to avoid blocking goroutines on sending a flush,
+		// if the "flushing" goroutine is not able read is because a flushing is ongoing.
+		select {
+		case forceFlushInstanceWriteBuffer <- true:
+		default:
+		}
```
shlomi-noach, line 2537: the `select` clause is confusing to read for me. Would you mind rewriting as a goroutine?
```
go func() {
  forceFlushInstanceWriteBuffer <- true
}()
```

go/inst/instance_dao.go
```diff
@@ -2492,9 +2496,14 @@ var forceFlushInstanceWriteBuffer = make(chan bool)
 
 func enqueueInstanceWrite(instance *Instance, instanceWasActuallyFound bool, lastError error) {
 	if len(instanceWriteBuffer) == config.Config.InstanceWriteBufferSize {
```
shlomi-noach, line 2529: Should this run if `config.Config.InstanceWriteBufferSize` == 0?

shlomi-noach (changes requested):
One minor request to change.

Also, is it possible to avoid the new logic when `InstanceWriteBufferSize == 0` or `BufferInstanceWrites == false`, i.e. in systems (such as most systems except your) where buffer writes are never enabled?

go/inst/instance_dao.go
luisyonaldo: oh my bad, added some checks to disable buffer writes if the buffer size is not valid:
https://github.com/github/orchestrator/blob/357fbb86e84bc136f53e41f26fb33e3b3b2fcac9/go/config/config.go#L575-L577

go/inst/instance_dao.go
luisyonaldo: the reason I use a non-blocking send (using a `select`), it's because I do not want to wait on sending the flush signal if the `forceFlushInstanceWriteBuffer` channel is blocked, since the channel being block means that the flushing process it's already happening, so I don't want to have multiple go-routines "waiting" to flush again a soon as the current flushing finish (the channel is unblocked).

shlomi-noach (approved):

PR #963, opened 2019-08-22
discovery: avoid backend writes of invalid instance

sjmudd:
### Description

The logic here looks wrong though it's unlikely to be called (frequently).

If the instance is invalid there's no point in even trying to indicate that we were attempting to update the instance, so remove the backend check (which would waste backend resources and time) and just return the current error correcting the function name shown.

- [x] code is formatted via `gofmt` (please avoid `goimports`)    [ using go1.12 ]


shlomi-noach:

Issue #966, opened 2019-08-23
delete from kv when forgetting instance or cluster


MaxFedotov:
Hello @shlomi-noach 

Right now all forget operations (ForgetInstance, ForgetCluster, ForgetLongUnseenInstances, ForgetUnseenInstancesDifferentlyResolved) do not remove entries from KV stores, which leads to outdated records in KV and can result in misconfiguration when you are using tools like consul-template in order to populate configuration for different proxies.

MaxFedotov:
PR #967 will fix this issue

PR #967, opened 2019-08-23
forget operations support removing from kv

MaxFedotov:
Related issue: #966 

Description:
Support for removing data from KV stores for different forget operations(ForgetInstance, ForgetCluster, ForgetLongUnseenInstances, ForgetUnseenInstancesDifferentlyResolved).


shlomi-noach:

MaxFedotov:
Hi, @shlomi-noach! Are there any updates?

shlomi-noach:
/o\ the update is: sorry for the delay 😬

MaxFedotov:
Hello @shlomi-noach how are you? :) Is there anything I should do in order to proceed with this PR?

shlomi-noach:
@MaxFedotov thanks for your patience!
So, I've been banging my head around this.

On one hand, the essence of this PR makes perfect sense. On the other hand, there is a destructive aspect to it. I'm trying to think whether stale KV entries should be kept around. e.g. the fact `orchestrator` forgot about an instance doesn't necessarily mean the instance does not exist anymore. Since external apps will depend on KV, I can't anticipate what assumptions those apps will make about existing entries.

For these reasons I find it difficult to approve the PR.

MaxFedotov:
> @MaxFedotov thanks for your patience!
> So, I've been banging my head around this.
> 
> On one hand, the essence of this PR makes perfect sense. On the other hand, there is a destructive aspect to it. I'm trying to think whether stale KV entries should be kept around. e.g. the fact `orchestrator` forgot about an instance doesn't necessarily mean the instance does not exist anymore. Since external apps will depend on KV, I can't anticipate what assumptions those apps will make about existing entries.
> 
> For these reasons I find it difficult to approve the PR.

Yeah, but on the other hand, if Orchestrator forgets the instance and it is still in KV, from which external tools get routing data - that will also lead to inconsistencies, e.g. routing to old\stale servers :)

What do you think about adding a configurable parameter to config.go, with the name like RemoveForgottenClustersFromKV, so everyone can decide for each own if they need this or not.

shlomi-noach:
> What do you think about adding a configurable parameter to config.go, with the name like RemoveForgottenClustersFromKV, so everyone can decide for each own if they need this or not.

@MaxFedotov sounds good to me!

go/kv/internal.go
```diff
@@ -65,3 +65,14 @@ func (this *internalKVStore) GetKeyValue(key string) (value string, found bool,
 func (this *internalKVStore) DistributePairs(kvPairs [](*KVPair)) (err error) {
 	return nil
 }
+
+func (this *internalKVStore) DeleteRecursive(key string) (err error) {
+	_, err = db.ExecOrchestrator(`
+		delete
+			from kv_store 
+		where
+		store_key rlike ?
```
shlomi-noach: will this delete `/mysql/master/important` when we forget the `import` cluster, seeing that `import` is a prefix of `important`?

shlomi-noach (changes requested):

go/kv/internal.go
MaxFedotov: That's true, my mistake, sorry. Replaced it to use `like`

MaxFedotov:
@shlomi-noach, so i've added RemoveForgottenClustersFromKV config option, fixed a bug with `rlike` in DeleteRecursive function for internal KV and updated the docs :) And tested everything one more time

MaxFedotov:
@shlomi-noach ping :) Is there anything i need to do in order to proceed with this PR?

go/kv/internal.go
shlomi-noach: Now it's even worse 😂 .

go/kv/internal.go
```diff
@@ -65,3 +67,9 @@ func (this *internalKVStore) GetKeyValue(key string) (value string, found bool,
 func (this *internalKVStore) DistributePairs(kvPairs [](*KVPair)) (err error) {
 	return nil
 }
+
+func (this *internalKVStore) DeleteRecursive(key string) (err error) {
+	query := fmt.Sprintf("DELETE FROM kv_store WHERE store_key LIKE '%s%%'", key)
```
shlomi-noach: This will delete `/mysql/master/important` when we forget the `import` cluster, seeing that `import` is a substring of `important`.
If we forget a cluster called `po`, this will forget `import`, `important`, `poo`, `topology` and any other cluster which has `po` as substring.

shlomi-noach (changes requested):

go/kv/internal.go
MaxFedotov: Yes, you are right with the first case, but the second won't be true, as the key is already in the form {kv_prefix}/key.
Think changing it to:
`query := fmt.Sprintf("DELETE FROM kv_store WHERE store_key = '%s' OR store_key LIKE '%s/%%'", key, key)`
should solve the issue.

shlomi-noach (approved):

leiless:
```diff
type KVStore interface {
	PutKeyValue(key string, value string) (err error)
	PutKVPairs(kvPairs []*KVPair) (err error)
	GetKeyValue(key string) (value string, found bool, err error)
+	DeleteRecursive(key string) (err error)
	DistributePairs(kvPairs [](*KVPair)) (err error)
}
```

```
[DEBUG] Build only; no packaging
[DEBUG] Building via go version go1.17.6 linux/amd64
go build: -i flag is deprecated
# github.com/openark/orchestrator/go/kv
go/kv/consul_txn.go:104:2: cannot use store (type *consulTxnStore) as type KVStore in return argument:
	*consulTxnStore does not implement KVStore (missing DeleteRecursive method)
```

Hi, @MaxFedotov, it seems that `go/kv/consul_txn.go` does not implemented `DeleteRecursive(key string) (err error)`, I don't know about Consul transaction store, can you refine this PR further?

Can I simply copy the same code from `consulStore.DeleteRecursive` to `consulTxnStore.DeleteRecursive`?

Like:
```go
func (this *consulTxnStore) DeleteRecursive(key string) (err error) {
	if this.client == nil {
		return nil
	}
	_, err = this.client.KV().DeleteTree(key, nil)
	return err
}
```

PR #968, opened 2019-08-26
Recovery processes ending with "&" are executed asynchronously

shlomi-noach:
This applies to all recovery processes: 

- `PreGracefulTakeoverProcesses`
- `PreFailoverProcesses`
- `PostMasterFailoverProcesses`
- `PostIntermediateMasterFailoverProcesses`
- `PostFailoverProcesses`
- `PostUnsuccessfulFailoverProcesses`
- `PostGracefulTakeoverProcesses`

A failure of an async process has no effect; the error is logged, but the rest of the processes continue to run and the flow proceeds.

Also: `PreGracefulTakeoverProcesses` now gets `{successorHost}`, `{successorPort}` information, since with graceful takeover we pre-select designated instance.


shlomi-noach:
Tests well in production

PR #974, opened 2019-09-02
support for routed-leader-check

shlomi-noach:
Recall that in a raft setup nodes relay (reverse proxy) API calls to the leader.

A `api/leader-check` call never routes through reverse proxy.

A new endpoint, `api/routed-leader-check` _does_. And this is useful for `orchestrator-client`. A misconfiguration in `ORCHESTRATOR_API` may not include the current leader (perhaps changes are pending, `puppet` to be deployed, whatever). But another node, which _is_ listed in `ORCHESTRATOR_API` can relay traffic to the leader. In which case `orchestrator-client` is happy to work with said node.


shlomi-noach:
Works as expected. Some non-proxied commands such as `raft-leader` will respond with an error message, e.g. `Cannot determine raft state`

PR #985, opened 2019-09-25
MySQLOrchestratorSSLSkipVerify to apply on backend TLS config

shlomi-noach:
This PR fixes what seems to be a bug: when certificate can be skipped (`MySQLOrchestratorSSLSkipVerify = true`) the existing code still demanded to add X509 key pair.

With this PR, when `MySQLOrchestratorSSLSkipVerify = true` and when no certificate files exist, then we don't attempt to load them.

Another logic seems to have been flawed in `SetupMySQLTopologyTLS`: again, certificates were required when `MySQLTopologySSLSkipVerify = true`.

The new logic is as follows:
```
		if (config.Config.MySQLTopologyUseMutualTLS && !config.Config.MySQLTopologySSLSkipVerify) &&
			config.Config.MySQLTopologySSLCertFile != "" &&
			config.Config.MySQLTopologySSLPrivateKeyFile != "" {
...
```


shlomi-noach:
cc @ggunson for visibility

sjmudd:
Code looks good. Going to double check it works here, so should be able to give you some feedback shortly.

shlomi-noach:
@sjmudd any thoughts per chance?

shlomi-noach:
I'm merging this PR given some constraints.

PR #1010, opened 2019-11-07
emergentlyRestartReplicationOnTopologyInstance fixes

shlomi-noach:
- restart replication fully (both SQL and IO threads)
 - fix concurrency/itration bug
 - slightly more logging

Yesterday we observed a situation where `stop slave io_thread; start slave io_thread;` was not enough to make a replica identify that it's master is broken. By stopping replication completely I hope this mitigates such behavior.

cc @timvaillancourt


timvaillancourt:
Thanks @shlomi-noach 👍

PR #1013, opened 2019-11-11
fix bug : concurrency error is modified by other goroutines

fengve:
https://github.com/github/orchestrator/issues/1000

Concurrent bug in ReadTopologyInstanceBufferable


shlomi-noach (approved):
Thank you -- looks good!

shlomi-noach (commented):
I very much appreciate this PR, and it is good and correct. ~~However, I prefer to use [atomic.Error](https://godoc.org/go.uber.org/atomic#Error) since this seems to be an already existing mechanism.
I'll draft an alternative PR.~~

shlomi-noach:
Uhhhh... forget about that! My mistake.

PR #1018, opened 2019-11-14
support for DiscoveryIgnoreMasterHostnameFilters

shlomi-noach:
A peculiar situation: a normal replication cluster needs to be fed some data from an external source. We set the `master` to replicate from that external source, and we _do not want_ `orchestrator` to crawl up to that external source.

To be experimented with.


shlomi-noach:
Also added `DiscoveryIgnoreHostnameFilters` for general-purpose ignoring of hosts.

PR #1020, opened 2019-11-14
Add IO thread lag to Instance struct

kbanman:
Related issue: https://github.com/github/orchestrator/issues/1019

### Description

This PR adds a `IOThreadLagSeconds` property to the Instance struct, allowing for awareness of network and disk latency slowing binlogs from being read by replica nodes.

This implementation uses the timestamp encoded by pseudo-gtid to infer how far behind the binlogs are on a given node.

It adds this calculated lag time to a new column on the `database_instance` table.


go/inst/instance_dao.go
```diff
@@ -804,6 +804,28 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
 		}
 	}
 
+	// PseudoGTID allows for measurement of IO thread lag
+	if instance.UsingPseudoGTID {
+		waitGroup.Add(1)
+		go func() {
+			defer waitGroup.Done()
+			_, pseudoGtidText, err := FindLastPseudoGTIDEntry(instance, instance.RelaylogCoordinates, nil, false, &instance.Binlog_format)
```
shlomi-noach, line 812: This is a relatively resource heavy operation. It's fine to run when you're in need to relocate a replica, but I wouldn't want to run this every 5 seconds as part of probing a server.

go/inst/instance_topology_dao.go
```diff
@@ -1093,6 +1095,17 @@ func injectPseudoGTID(instance *Instance) (hint string, err error) {
 	return hint, log.Errore(err)
 }
 
+func extractPseudoGTIDTime(pseudoGTIDText string) (entryTime time.Time, err error) {
+	queryExp, err := regexp.Compile(`drop view if exists (.+)._asc:(.+)`)
```
shlomi-noach, line 1099: This regex is only true assuming you use `AutoPseudoGTID`

go/inst/instance_topology_test.go
```diff
@@ -568,3 +569,10 @@ func TestChooseCandidateReplicaPromoteRuleOrdering3(t *testing.T) {
 	test.S(t).ExpectEquals(len(laterReplicas), 0)
 	test.S(t).ExpectEquals(len(cannotReplicateReplicas), 0)
 }
+
+func TestExtractPseudoGTIDTime(t *testing.T) {
+	gtidText := "drop view if exists `_pseudo_gtid_`.`_asc:5dcc60bc:01cf8718:0f1e4130686bc1bc`"
+	entryTime, err := extractPseudoGTIDTime(gtidText)
```
shlomi-noach, line 575: Thank you for adding a test!

shlomi-noach (changes requested):
Thanks you! As per discussion on https://github.com/github/orchestrator/issues/1019, I think Pseudo-GTID does not have a good-enough interval to measure replication lag; a `pt-heartbeat` or similar mechanism would be better. Please also see comments inline. I'm sorry!

kbanman:
Thanks for the feedback @shlomi-noach. 

Would you be open to adding the functionality of `pt-heartbeat` to orchestrator? Seems it would be a good fit for an opt-in feature.

shlomi-noach:
> Would you be open to adding the functionality of pt-heartbeat to orchestrator? Seems it would be a good fit for an opt-in feature.

I'm ambivalent. It's certainly within `orchestrator`'s capabilities: `orcehstrator` knows which is the `master` host, can validate it's writable, etc.

But that would mean `orchetrator` turns into a critical path in the topology's replication health. If `orchestrator` went down - the cluster would _seem to be lagging_ (even if not really lagging). Not sure I like this dependency on `orchestrator`.

PR #1033, opened 2019-12-06
Fix upstream issue 1032: Instance alias display not working in aggregation.

jfg956:
Related issue: https://github.com/github/orchestrator/issues/1032

Add fixing alias in addition to title in aggregation logic.


shlomi-noach (approved):
Thank you!

Issue #1053, opened 2020-01-27
Allow sorting of clusters in the dashboard


jhriggs:
The ordering of clusters in the dashboard was a mystery to me until I looked at the code. To me (and our team), sorting by the cluster instance count seemed strange and unintuitive. It may just be us, but regardless, it seems this should be configurable.

PR incoming to allow sorting by count (current behavior, default), cluster name, and cluster alias.

jhriggs:
Merged.

PR #1054, opened 2020-01-27
Allow sorting clusters on dashboard by count, name, or alias

jhriggs:
Related issue: #1053

In addition to the current default of sorting clusters by decreasing instance count, allow the user to choose sorting by cluster name or cluster alias via drop down in the nav bar. The user's selection is stored in a long-lived cookie.


shlomi-noach (approved):
thank you

PR #1056, opened 2020-01-27
instance problems dropdown menu: use fixed max-height and overflow

### LINKED ISSUE #1055: not posible to scroll down on instance problems dropdown menu
When there are too many instance problems (sadly we do :)), is it not posible to see all issues since the height of the dropdown-menu is bigger than the windows' height.

Would be nice to have a fixed height and be able to scroll (like in the clusters dropdown menu).

luisyonaldo:
Fixes: https://github.com/github/orchestrator/issues/1055

### Description

Add fixed `max-height` and `overflow` into instance problems dropdown menu, so we can scroll in case of multiple problems.


shlomi-noach (approved):
looks good - thank you!

Issue #873, opened 2019-05-07
remote error: tls: bad certificate


git001:
## Question
How can I debug the `remote error: tls: bad certificate` below. It's not clear for me which part of orchestrator have tls problems.

## config

```
cat /var/lib/orchestrator/orchestrator-sqlite.conf.json
{
    "Debug": true,
    "EnableSyslog": false,
    "ListenAddress": ":3000",
    "AutoPseudoGTID": true,
    "RaftEnabled": true,
    "RaftDataDir": "/var/lib/orchestrator",
    "RaftBind": "104.248.131.78",
    "RaftNodes": ["mysql-001.livesystem.at", "mysql-002.livesystem.at", "mysql-003.livesystem.at"] ,
    "BackendDB": "sqlite",
    "SQLite3DataFile": "/var/lib/orchestrator/data/orchestrator.sqlite3",
    "MySQLTopologyCredentialsConfigFile": "/var/lib/orchestrator/orchestrator-topology.cnf",
    "InstancePollSeconds": 5,
    "DiscoverByShowSlaveHosts": false,
    "FailureDetectionPeriodBlockMinutes": 60,
    "UseSSL": true,
    "SSLPrivateKeyFile": "/var/lib/orchestrator/pki/mysql-001.livesystem.at_privatekey.pem",
    "SSLCertFile": "/var/lib/orchestrator/pki/mysql-001.livesystem.at_cert.pem",
    "SSLCAFile": "/var/lib/orchestrator/pki/ca_cert.pem",
    "SSLSkipVerify": false,
  }
```

## debug output

```
root@mysql-001:~# cd /usr/local/orchestrator && orchestrator --debug --config=/var/lib/orchestrator/orchestrator-sqlite.conf.json --stack http
2019-05-07 10:14:49 INFO starting orchestrator, version: 3.0.14, git commit: f4c69ad05010518da784ce61865e65f0d9e0081c
2019-05-07 10:14:49 INFO Read config: /var/lib/orchestrator/orchestrator-sqlite.conf.json
2019-05-07 10:14:49 DEBUG Parsed topology credentials from /var/lib/orchestrator/orchestrator-topology.cnf
2019-05-07 10:14:49 DEBUG Connected to orchestrator backend: sqlite on /var/lib/orchestrator/data/orchestrator.sqlite3
2019-05-07 10:14:49 DEBUG Initializing orchestrator
2019-05-07 10:14:49 DEBUG Migrating database schema
[...]

PR #1059, opened 2020-02-03
Fix healthcheck, support mTLS on client

smirnov-vs:
Related issue: #873 

### Description

If API uses SSL and mTLS is enabled, healthcheck requests to leader are unsuccessful.
In logs you can see something like:
`remote error: tls: bad certificate`

So, `setupHttpClient` is loading TLS configuration now - root certs and mTLS certs.

### DIFF
--- a/go/ssl/ssl.go
+++ b/go/ssl/ssl.go
@@ -54,19 +54,30 @@ func NewTLSConfig(caFile string, verifyCert bool) (*tls.Config, error) {
 		log.Info("verifyCert requested, client certificates will be verified")
 		c.ClientAuth = tls.VerifyClientCertIfGiven
 	}
+	caPool, err := ReadCAFile(caFile)
+	if err != nil {
+		return &c, err
+	}
+	c.ClientCAs = caPool
+	c.BuildNameToCertificate()
+	return &c, nil
+}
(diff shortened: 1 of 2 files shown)


go/ssl/ssl.go
```diff
@@ -54,19 +54,30 @@ func NewTLSConfig(caFile string, verifyCert bool) (*tls.Config, error) {
 		log.Info("verifyCert requested, client certificates will be verified")
 		c.ClientAuth = tls.VerifyClientCertIfGiven
 	}
+	caPool, err := ReadCAFile(caFile)
+	if err != nil {
+		return &c, err
+	}
+	c.ClientCAs = caPool
+	c.BuildNameToCertificate()
```
shlomi-noach, line 62: I wonder if this enforces an Authority, or will this still allow self-signed certificates?

smirnov-vs: We use self-signed certs with a local root CA certificate without any problems.
Actually problem was not in this line. This isn't my code, i just reused old version of code for reading CA certificate in new `ReadCAFile` function.
This function is used in client and server now because for server you need `ClientCAs` and `RootCAs` for client.

Issue #1063, opened 2020-02-10
Output message in moveReplicasViaGTID


rluisr:
ref: #1061 

I don't understand why replicas couldn't move to new master currently.
We need add `log.Errorf` in here.

PR #1064, opened 2020-02-10
Add log.Errorf in moveReplicasViaGTID

rluisr:
issue: #1063


go/inst/instance_topology.go
```diff
@@ -692,6 +692,7 @@ func moveReplicasViaGTID(replicas [](*Instance), other *Instance, postponedFunct
 				movedReplica, replicaErr := moveInstanceBelowViaGTID(replica, other)
 				if replicaErr != nil && movedReplica != nil {
```
rluisr, line 693: I think `moveReplica` always not nil.

PR #1070, opened 2020-02-12
Turn seconds into a more meaningful Wd Xh Ym Zs output

sjmudd:
# Description

Displaying of replication delay or time since seeing a host in seconds is fine except when you may have some boxes that have a very high value. Then it's better to show the value in terms of days/hours/minutes/seconds as that's easier to read and avoids the need for mental conversion.

This tiny patch presents the delays in the web interface in for form `1d 2h 3m 4s lag` or `1d 2h 3m 4s ago` which makes these delays easier to read.

- [x] code is tested


shlomi-noach:
superseded by #1184

PR #1075, opened 2020-02-13
Add option to set rejectReadOnly option of mysql driver

luisyonaldo:

Related issue: https://github.com/github/orchestrator/issues/1074

### Description

Adds option to configure rejectReadOnly for MySQL orchestrator database config

This was tested in our current setup:

before
```
Feb 13 10:50:50 xxx orchestrator[7714]: Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
```
after (setting `MySQLOrchestratorRejectReadOnly: true`)
```
Feb 13 12:35:42 xxx orchestrator[64671]: driver: bad connection
```

bad connection error forces a new connection, which will re-sync DNS entry for the database and point to the new promoted and writable master.


luisyonaldo:
before:
```log
Feb 13 10:50:50 xxx orchestrator[7714]: Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
```

after:
```log
Feb 13 12:35:42 xxx orchestrator[64671]: driver: bad connection
```

**bad connection** error forces a new connection, which will re-sync DNS entry for the database and point to the new promoted and writable master.

sjmudd:
Shlomi: context. When using group replication the primary server may change to be a read-only server. As the connection to the server is good golang will not drop the connection and attempts to write to that connection will fail. Given this is supposed to be a writeable connection the change Luis has made (provided by the Go MySQL driver) basically treats a failed write to a `read-only` server as an error and bad connection. The connection will be closed and a new connection will be open and hopefully that will reach the new primary server.

shlomi-noach:
@sjmudd is this about cluster as backend for `orchestrator`, or is this about `orchestrator` probing cluster? I'm a bit confused.

luisyonaldo:
> @sjmudd is this about cluster as backend for `orchestrator`, or is this about `orchestrator` probing cluster? I'm a bit confused.

this is about MySQL as backend for `orchestrator`

luisyonaldo:
@shlomi-noach is the issue clear enough?

shlomi-noach:
I believe it is! Sorry for delays, I’m ooo and only briefly looking at issues and PRs.

shlomi-noach:
Question: is there a scenario where this will be thrashing? e.g. `orchestrator` sees a `read-only` server, errors the connection, connects again, it's still `read-only`, errors the connection ad infinitum?

sjmudd:
I think it might happen, so it might make sense to improve this further by a small additional change:

-  add extra logging to indicate what's happening and why, even if the code and logic behaves the same.  e.g. `Dropping read-only connection and forcing a driver re-connection` or something similar. We can figure out appropriate text afterwards.
- It may also make sense to add a small/random/configurable delay before doing this.  @luisyonaldo what do you think about that?

luisyonaldo:
@sjmudd this logging happens on mysql driver so not much we can do there. And right now what @shlomi-noach describes is what's happening, application get stuck on a infinite loop logging 
```
Feb 13 10:50:50 xxx orchestrator[7714]: Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
```
with the `rejectReadOnly` a re-connection will be forced, that's it, if this doesn't fix the connection, then the application will be stuck without being able to connect to database (nothing much to do at this point).

sjmudd:
Hi @shlomi-noach,

Is there any chance you can look to push this? We bumped into this again and didn't notice and it would be good as I think  this removes any difference from you (our upstream) so that's good.

Also a master orchestratordb backend failure which gets resolved can lead to orchestratorapp talking to the old host which is now read only and thus while everything else works all attempts to write will fail which breaks the expected behaviour.

Sample logging seen earlier today:

```
2020-07-03T16:44:00.898891+02:00 orchestratorapp orchestrator[39845]: Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
2020-07-03T16:44:00.899073+02:00 orchestratorapp orchestrator[39845]: ContinuousRegistration: RegisterNode failed: Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
2020-07-03T16:44:00.899254+02:00 orchestratorapp orchestrator[39845]: Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
2020-07-03T16:44:00.914533+02:00 orchestratorapp orchestrator[39845]: 2020-07-03 16:44:00 ERROR Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
2020-07-03T16:44:00.914732+02:00 orchestratorapp orchestrator[39845]: 2020-07-03 16:44:00 ERROR Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
2020-07-03T16:44:00.914899+02:00 orchestratorapp orchestrator[39845]: 2020-07-03 16:44:00 ERROR Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
2020-07-03T16:44:00.915065+02:00 orchestratorapp orchestrator[39845]: Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
2020-07-03T16:44:00.915236+02:00 orchestratorapp orchestrator[39845]: Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
2020-07-03T16:44:00.915416+02:00 orchestratorapp orchestrator[39845]: Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
2020-07-03T16:44:01.756528+02:00 orchestratorapp orchestrator[39845]: 2020-07-03 16:44:01 ERROR Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
2020-07-03T16:44:01.756829+02:00 orchestratorapp orchestrator[39845]: 2020-07-03 16:44:01 ERROR Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement
```

So please if you can, consider applying this patch.
Thanks!

Simon

shlomi-noach (approved):

PR #1083, opened 2020-02-24
Web interface read only

### LINKED ISSUE #1082: Adding new config option to make the GUI read-only
Whilst `ReadOnly` can be used in conjunction with `orchestrator` in CLI mode, it cannot be used when using `orchestrator-client`.

Adding a new option, `WebInterfaceReadOnly`, will allow this so that the GUI can be force into read-only mode, regardless of authentication mechanism and user.

cezmunsta:
Resolves #1082


go/http/httpbase.go
```diff
@@ -27,20 +28,39 @@ import (
...
 	return ""
 }
 
+// isForcedReadOnly checks config to see if the WebInterfaceReadOnly flag is set
+// This is independent of any authentication mechanisms
+func isForcedReadOnly(uri string) bool {
+	if fmt.Sprintf("%v", regexURI) == regexInitRequired {
+		log.Debug("Compiling regexp for isForcedReadOnly")
+		regexURI = regexp.MustCompile(fmt.Sprintf("^%s/web/.*", config.Config.URLPrefix))
+	}
```
shlomi-noach, line 52: Can you please clarify the use of the regex here? I confess I do not understand what this code is doing.
Also, please note `regexURI` is a global variable, which is being compared to, and manipulated, in `isForcedReadOnly(...)`, which gets called by `isAuthorizedForAction(...)`, which gets called by various API/Web endpoints, potentially concurrently. Thus, regardless of the purpose of the logic above, it is incorrect seeing that multiple concurrent calls to API/Web will corrupt the value of `regexURI`.

shlomi-noach (changes requested):

go/http/httpbase.go
cezmunsta: So, the regex is intended to explicitly match the configured URLPrefix, but that is not available at compile time. Instead of repeatedly compiling a regex, I was looking to have it set correctly by the first call that triggers it. Are you happy with a broader match?
```
var regexURI = regexp.MustCompile("^(.*/)?web/.*")
```

go/http/httpbase.go
```diff
@@ -27,20 +28,39 @@ import (
 	"github.com/github/orchestrator/go/inst"
 	"github.com/github/orchestrator/go/os"
 	"github.com/github/orchestrator/go/process"
-	"github.com/github/orchestrator/go/raft"
+	orcraft "github.com/github/orchestrator/go/raft"
```
EagleEyeJohn, line 31: is `orcraft ` expected here?

cezmunsta: > is `orcraft ` expected here?

John, it is an alias and distinguishes the `orchestrator` raft module from any other. It is then used when referencing variables and calling functions etc. This is not a change from this PR, but exists in the main branch. 

e.g.
```
if orcraft.IsRaftEnabled() && !orcraft.IsLeader() {
```

cezmunsta: Also, this PR will not be merged and remains as a reminder for @shlomi-noach for token-controlled authentication

PR #1088, opened 2020-03-03
Added dynamic cluster messages

### LINKED ISSUE #1086: Dynamic web message on cluster page
Hello.
First of all, thanks for your great app.

Our recovery process is based on moving VIP from old master to a new one. Some operations from our bash scripts might fail during the recovery, e.g. the VIP didn't shut down on old master and potentially we would have split brain in some scenarios.

I'd like to add some alerts on cluster page in UI to visually notify administrator if something went wrong or if cluster maintenance requires some manual actions after failover.

Is it possible to do from hooks or orchestrator-client?

pparshin:
It might be useful during or after the failover to show custom messages to DBA team, e.g. if some failover operations fail and cluster recovery requires manual actions to be executed.

Each message can be acknowledged and removed from cluster page.

Resolves https://github.com/openark/orchestrator/issues/1086


shmel1k:
@shlomi-noach , can you please take a look at this PR? :)

shlomi-noach:
I'm on vacation/ooo (as my status indicates) so will only look at this when I'm back.

shlomi-noach:
Hi again. While reviewing the code (and associated issue #1086) I'm not sure I understand the use case. Can you please give an example scenario? What happens, what's missing; how you would use the new changes?
On first looks, it seems to put some functionality into `orchestrator` that can be put elsewhere (are those playbooks for your team? Can these be opened as a GitHub issue or a Jira ticket instead?)

pparshin:
Hi, 

When manual or auto failover happens we realized that status messages on the cluster page can be helpful to quickly understand is everything is going OK or our failover process has been failed and we need to do some manual actions. Likely in case of crash I'm going to check orhestrator page firstly and after will read logs, alerts, metrics, etc. 

Creating Jira tickets is a good alternative but it seems like a solution for long-running or not urgent  problems, e.g. investigate the root cause of the failover.

Screenshot how it might looks like:

image

If our VIP migration is failed, it would be possible to understand just looking on this page. It is important because orchestrator can rebuild the topology successfully but to move the traffic to a new master we have to migrate the VIP.

jackiesre721:
Excuse me, the VIP switching information displayed on the picture is, we are writing directly in the hook script.

Issue #1089, opened 2020-03-06
audit recoveries by cluster alias is not working


luisyonaldo:
Not posible to audit recoveries by searching by cluster alias.

https://orchestrator.booking.com/web/audit-recovery/alias/$clusterAlias *always* shows the same view as https://orchestrator.booking.com/web/audit-recovery/alias (i.e. recoveries are not being filter by cluster alias)

PR #1090, opened 2020-03-06
search recoveries by cluster alias

luisyonaldo:
Fixes issue: https://github.com/openark/orchestrator/issues/1089

https://orchestrator.booking.com/web/audit-recovery/alias/$alias

is not filtering by cluster alias as expected and show results for all clusters instead.

Also add filtering by `page` on both **recoveries** and **failure detection** audits when searching by `clusterAlias`.


resources/public/js/audit-recovery.js
```diff
@@ -2,8 +2,11 @@ $(document).ready(function() {
   $("#audit_recovery_steps").hide();
   showLoader();
   var apiUri = "/api/audit-recovery/" + currentPage();
-  if (auditCluster()) {
-    apiUri = "/api/audit-recovery/cluster/" + auditCluster() + "/" + currentPage();
+  if (clusterName()) {
+    apiUri = "/api/audit-recovery/cluster/" + clusterName() + "/" + currentPage();
+  }
```
koorgoo: Why not `else if`? Name & alias cannot exist at the same time. State it in the code.

luisyonaldo: just to follow convention, current code is like this in multiple places (e.g. `audit-failure-detection.js`, can change on a follow up MR if wanted

shlomi-noach (approved):
Looks good

PR #1092, opened 2020-03-08
    Allow maps to assign instances to data center, region, and environment

### LINKED ISSUE #1091: [Feature] Allow maps to assign instances to data center, region, and environment
Not all environments allow for determining DC, region, and environment from a hostname regex or database query. In my particular case, I can determine region and environment from our DNS naming standard, however, the DC is not available from the name or database. I could use configuration management or a cron job or something to inject something into the DB, but it is extraneous information and becomes an ongoing maintenance item.

But, it would be quite simple for me to build a list or mapping from our AWS subnets to DCs (AWS AZs) that I can bake into the Orchestrator configuration. I imagine it would be useful to add an additional map setting for each of these, `DataCenterMap`, `RegionMap`, and `PhysicalEnvironmentMap` that map an instance to each of these bits of metadata by:

1. hostname + port
2. hostname
3. IP(s) + port
4. IP(s)
5. CIDR

Incoming PR that does just this…

jhriggs:
This adds three new settings `DataCenterMap`, `RegionMap`, and `PhysicalEnvironmentMap` that allow specifying an instance's DC, region, or environment using a map whose keys match an instance as follows:

    1. <hostname>:<port>
    2. <hostname>
    3. <ip>:<port> (loop over host IPs)
    4. <ip> (loop over host IPs)
    5. CIDR network specification (loop over host IPs)

    Example:
    ```
    "DataCenterMap": {
      "db-foo-bar.myservice.com": "dc1",
      "192.168.123.123:3307": "dc2",
      "192.168.0.0/24": "dc1",
      "192.168.1.0/24": "dc2",
      "192.168.3.0/24": "dc3"
    },
    ```

    Fixes #1091.


jhriggs:
(Note that I am no Go developer; I am just learning as I go along, so feel free to correct/adjust/suggest if anything doesn't match standards or norms!) :-)

PR #1096, opened 2020-03-12
skip AddReplicaKey if it is specified in config.Config.DiscoveryIgnoreReplicaHostnameFilters

MaxFedotov:
Hi @shlomi-noach. 
As I understand, there is a special parameter in config DiscoveryIgnoreReplicaHostnameFilters, which is used especially to skip adding some replicas to Orchestrator (like different binlog dump routines). Right now, although replica is not added, it is still mentioned in slave_hosts field in database_instance table in Orchestrator DB, so API endpoints like /cluster/ still return such replicas in Json responses and it confuses all other services, which may rely on this data. 

This PR prevents such situations, as all hosts specified in DiscoveryIgnoreReplicaHostnameFilters wouldn't be added to Orchestrator DB.


shlomi-noach (approved):
Thank you -- this makes sense.

PR #1115, opened 2020-04-06
Support for FailMasterPromotionOnLagMinutes

### LINKED ISSUE #83: Slaves lagging by couple of hours are elected as master by orchestrator

shlomi-noach:
Fixes #83 

Scenario: `M -> R` topology was running with replica `R` broken for a few hours without anyone noticing. Then, master `M` fails.

Curent behavior: `orchestrator` promotes `R`. But by this we lose:

- potentially hours of worth of relay logs obtained by `R` but not executed (e.g. because of some SQL error), or
- the ability to recover hours of worth of binary logs from the master `M`.

New config variable `FailMasterPromotionOnLagMinutes` tells `orchestrator` to fail a promotion if, at the time a candidate replica is chosen, it is determined to be lagging too much ( >= `FailMasterPromotionOnLagMinutes`).

cc @sougou @mcrauwel


mcrauwel:
thanks for the mention @shlomi-noach! I will be testing it out!

shlomi-noach:
Behavior of this change is now validated via the (new) system tests framework:

- https://github.com/openark/orchestrator/pull/1115/files#diff-efcf9a672f1479fff669af09798f8006
- https://github.com/openark/orchestrator/pull/1115/files#diff-ae05c7c070d7188d3209b979c4881bed

Issue #1125, opened 2020-04-14
the analysis log can not tell which cluster


yangeagle:
https://github.com/openark/orchestrator/blob/244aa5eef45df43a52eff9adfb05599ed8dba52e/go/inst/analysis_dao.go#L310-L317

the log here can not tell which cluster.

if two clusters have the same analysis at the same time, 
only one `analysisMessage` printed.

If agreed, I can give PR with pleasure.

cc @MOON-CLJ

shlomi-noach:
Behavior confirmed. Thank you!

yangeagle:
close it.

PR #1126, opened 2020-04-14
add cluster name in analysis log

yangeagle:
ref: https://github.com/openark/orchestrator/issues/1125


go/inst/instance_dao.go
```diff
@@ -301,6 +301,8 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
 	errorChan := make(chan error, 32)
 	var resolveErr error
 
+	instance.Key = *instanceKey
+
```
shlomi-noach: Can you please clarify why you moved this line to this location? `if !instanceKey.IsValid()` is important context. If the key is invalid, we do not assign `instance.Key`.

shlomi-noach (commented):
Sorry for the late response. Adding cluster name is 👍 . Asking to clarify what seems to be an unrelated change.

yangeagle:
> Sorry for the late response. Adding cluster name is 👍 . Asking to clarify what seems to be an unrelated change.

Sorry for extra change. I have fixed it
@shlomi-noach

shlomi-noach:
resubmitted as https://github.com/openark/orchestrator/pull/1152 because this PR originates at `master` branch

PR #1130, opened 2020-04-21
graceful master takeover to support automatically picking replica to …

### LINKED ISSUE #908: Graceful-switchover to support automatically picking the replica to promote
Currently, graceful switchover requires the user to select a replica to promote to the new master, however it would be nice if orchestrator could automatically pick a (if not "best") replica to promote.

"Best" replica for promotion could take into account promotion rules and replication status, i.e. a replica with a lot of lag might not be the best to promote.

mzhou151:
…promote

fix https://github.com/openark/orchestrator/issues/908 

**What’s the problem**
Currently orchestrator's graceful master takeover API implementation supports two scenarios for a graceful takeover
1) the master has only only replica replicating from it, or
2) the master can have many replicas replicating from it, but the caller needs to designate one replica as the candidate when using the API.

Neither meets our needs for a graceful automatic master takeover.

**What’s needed**
The majority of our clusters are master-replica set ups with multiple replicas of a single master. When we initiate a graceful master takeover (either manually or during a maintenance period), we need Orchestrator to automatically select the best candidate to promote.  If no valid candidate is found, the takeover should be aborted and no topology changes should happen on the given cluster. The best candidate is a replicant where the replication delay is within a threshold and the replicant is also in the same datacenter as the current master.

**How it is implemented**
This code change is to add such support to Orchestrator’s graceful master takeover API to automatically select a best candidate during a graceful master takeover, by adding an additional parameter “suggestReplica=true”.

First we do basic checks on all replicas given a cluster

1. The replica’s replication delay (in seconds) needs to be within a configurable threshold.
2. The replica should not have an errant GTID.
3. The replica is not banned by a promotion rule.

If none of the replicas pass all the above checks, we abort and return the error.
- If any replicas pass all the above checks, we proceed with locking the master by issuing ‘flush tables with read lock’ on the master. 
- If the command fails, or times out after a defined config, we abort the takeover and return the error. If the command succeeds, we then wait for the replicas’ executed binlog coordinates to reach that of the master. 
- If there are multiple replicas whose executed binlog reaches that of the master, we prefer the replica that’s in the same datacenter as the master. 
- We then return the selected replica as the best candidate for Orchestrator to continue with the rest of the graceful takeover process.


shlomi-noach (commented):
Thank you for this pull request and sorry for the delays. A few questions:

1. `orchestrator` uses a `read-only`/`super-read-only` on the `master` rather than `flush table with read lock`/`unlock tables` for a reason. That is, on VeryBusySystems™ a `flush tables with read lock` is a time consuming operation on an active `master`; I've experiences *minutes* of lockdown while `flush tables...` was running. To be honest, I haven't tried it in production in a long time (at least 4 years) because I was not willing to take the chance.

2. I'm not sure about the use of `GetMatchingExecBinlogCoordinatesReplicas`. 
  a. it naively waits for _all_ replicas to catch up. What if one of them is lagging? Your `master` is locked down during this time, correct? So this operation is risky.
  b. why not first `RegroupReplicas...` under your desired replica, if possible? Tkae an optimistic approach using the existing mechanisms. If you're able to make your candidate replica an intermediate master for all other replicas, you're good to proceed. If you can't, well, then you should not run a graceful takeover (because it will not be graceful).

Does that make sense?

shlomi-noach:
Thank you again. Inspired by this PR I've chosen to implement it differently in #1165 , and using existing building blocks. The new command, `graceful-master-takeover-auto` both auto-picks a replica to promote, and also starts replication on demoted master after promotion.

shlomi-noach:
superseded by #1165

PR #1151, opened 2020-04-28
Show reason if begin_maintenance fails

sjmudd:
When trying to relocate a server via the GUI I saw the message:

`Cannot begin maintenance on host.example.com:3306`

This is not very helpful as it does not say *why* which would be useful. It was not apparent to me what the actual problem was and looking in the code there are quite a few code paths where this message can be generated, basically a call to `BeginMaintenance()`.

So this patch attempts to retrieve the error from `BeginMaintenance` and expose it to the error message I was seeing hopefully to explain what the problem actually is.


go/inst/instance_topology.go
```diff
@@ -511,13 +511,13 @@ func MoveBelow(instanceKey, siblingKey *InstanceKey) (*Instance, error) {
 	log.Infof("Will move %+v below %+v", instanceKey, siblingKey)
 
 	if maintenanceToken, merr := BeginMaintenance(instanceKey, GetMaintenanceOwner(), fmt.Sprintf("move below %+v", *siblingKey)); merr != nil {
-		err = fmt.Errorf("Cannot begin maintenance on %+v", *instanceKey)
+		err = fmt.Errorf("Cannot begin maintenance on %+v: merr", *instanceKey, merr)
```
shlomi-noach: You probably meant this:
```suggestion
		err = fmt.Errorf("Cannot begin maintenance on %+v: %v", *instanceKey, merr)
```

shlomi-noach (changes requested):
typo/syntax error in one of the changes. Otherwise looks good.

go/inst/instance_topology.go
sjmudd: yep, my bad. I just wanted to see if this made sense. will fix.

sjmudd: So I think patch should be good.

shlomi-noach (approved):

PR #1155, opened 2020-05-02
Towards system tests dependencies

shlomi-noach:
Each test will be able to specify a list of dependency tests via `depends-on` file, listing one test per line.

A test will only run iff all of its dependencies have run successfully.
A test will be automatically marked as failed if one of its dependency has failed.

This implies ordering of system tests.


huipeng11:
#1155



---Original---

Issue #1137, opened 2020-04-24
No way to handle the only semi-sync replica failing


kklimonda:
(updated topology screenshot to match the question)

I'm testing the following topology:

<img width="1021" alt="Screenshot 2020-04-24 at 11 56 27" src="https://user-images.githubusercontent.com/221724/80200407-e360bf80-8622-11ea-833c-1d5cb9255547.png">

Single master that has semi-sync replication enabled, with single semi-sync slave running in another DC. The idea here is to gracefully failover to another DC in case master's DC goes offline.

I have a PostFailover hook that makes sure that, in case of master failover, semi-sync replication is enabled on either replica from another DC (preferably) or replica in the same DC as master and disabled on all other replicas.

This seems to be working fine, however I can't figure out how to handle failure of semi-sync replica - there is no hook executed when replica is lost, and so master stops accepting writes as it's waiting for ACK from its remaining replicas (none of which have semi-sync replication enabled).

Is this somehow related to Semi-sync enforcement? I've already raised similar question on the ML, but it's not obvious from documentation and code that this feature is a correct approach here. 

One idea I had was to turn that topology into something similar to that:

<img width="1022" alt="Screenshot 2020-04-24 at 11 30 42" src="https://user-images.githubusercontent.com/221724/80197611-26209880-861f-11ea-804f-eb18019c9553.png">

In this topology semi-sync replica is also intermediate master, which triggers hooks on failure and lets me reconfigure other replica for semi-sync replication.

shlomi-noach:
> how to handle failure of semi-sync replica - there is no hook executed when replica is lost, and so master stops accepting writes as it's waiting for ACK from its remaining replicas (none of which have semi-sync replication enabled).

If I understand correctly, in the post failover hook you try and set up a new replica as semi-sync, and, since that could fail, you'd wish to have some hook, so you can try another replica?

I think of a different solution. Post failover, you can check with orchestrator which replicas are OK, and then choose one to enable semi-sync. Does that make sense?

Regardless, am I correct to understand you never ever want to have a write on a master unless backed by a replica? How about, for the duration of the failover, you set `rpl_semi_sync_master_timeout` to a reasonably low value, to allow for some recovery mishaps and delays, and eventually set `rpl_semi_sync_master_timeout` to a high value?

kklimonda:
Should PostFailover hook be running not only on master failover, but also on failure of replicas? Or perhaps other hook that is executed in that case?

[...]

PR #1173, opened 2020-05-21
WIP: Analysis: LockedSemiSyncMaster

shlomi-noach:
A new analysis entry aimed to identify a blocked semi-sync master situation.

Example scenario depicted in https://github.com/openark/orchestrator/issues/1137

e.g. a master has a single semi-sync replica and a configuration of:

- `rpl_semi_sync_master_enabled=ON`
- `rpl_semi_sync_master_timeout=10000000000`
- `rpl_semi_sync_master_wait_for_slave_count=1`

if the semi-sync replica goes down, the master blocks writes. We want to identify this as a failure scenario (a recovery might be: enable semi-sync on a different replica).

However, we probably also want to make this detection only after some X seconds, because due to the nature of the setup, it may be normal for the master to block for 1 or 2 seconds occasionally.
The issue right now (which is why this PR is WIP at this time) is to be able to reliably say "the master has been blocked since 20sec ago". There's various ways to go about that, which I'm exploring.


shlomi-noach:
#1175 replaces this PR. Closing without merging.

Issue #1179, opened 2020-05-31
Feature Request: Add support for MySQL Group Replication


ejortegau:
I would like to have at least some level of support for Group Replication set-ups in Orchestrator. The bare minimum support would include:

* At least the (simpler) single primary mode, since it is somewhat more similar to the traditional single primary master in async/semi-sync replication.
* The API returns all group members as part of the same cluster in calls to `/api/cluster/<cluster_alias>`
* The web UI shows the group members in the same view, instead of what happens today, in which each group member looks like a separate cluster.
* Orchestrator does not consider GTIDs from the group as errant anymore.
* Orchestrator prevents the user from performing incorrect topology for group members (e.g. set a primary to replicate from a secondary; or move a secondary to replicate from somewhere else, since secondaries actually replicate from the group's primary).

PR #1180, opened 2020-05-31
Add basic support for group replication

ejortegau:
Related issue: https://github.com/openark/orchestrator/issues/1179
### Description

This PR adds some initial support for group replication in orchestrator.

ToDo:
- [X] code is formatted via gofmt (please avoid goimports)
- [X] code is built via ./build.sh
- [x] relevant documentation has been updated.
- [x] upgrade from previous version (`master` branch) is successful

Assume a 3 member group, plus an async slave replicating from one of them. Orchestrator shows all three members as separate clusters, as it does not understand group replication and there is no traditional replication configured in any group member. The async slave is shown as a slave in the cluster whose master is the group member the slave is set up to replicate from.

Orchestrator also shows the async slave with a problem of errant GTIDs.

With this  PR, some basic support for single primary replication groups is added. So, a replication group shown like this in mysqlshell:
```json
{
    "clusterName": "test_cluster", 
    "defaultReplicaSet": {
        "name": "default", 
        "primary": "127.0.0.1:5306", 
        "ssl": "REQUIRED", 
        "status": "OK", 
        "statusText": "Cluster is ONLINE and can tolerate up to ONE failure.", 
        "topology": {
            "127.0.0.1:4306": {
                "address": "127.0.0.1:4306", 
                "mode": "R/O", 
                "readReplicas": {}, 
                "replicationLag": null, 
                "role": "HA", 
                "status": "ONLINE", 
                "version": "8.0.20"
            }, 
            "127.0.0.1:5306": {
                "address": "127.0.0.1:5306", 
                "mode": "R/W", 
                "readReplicas": {}, 
                "replicationLag": null, 
                "role": "HA", 
                "status": "ONLINE", 
                "version": "8.0.20"
            }, 
            "127.0.0.1:6306": {
                "address": "127.0.0.1:6306", 
                "mode": "R/O", 
                "readReplicas": {}, 
                "replicationLag": null, 
                "role": "HA", 
                "status": "ONLINE", 
                "version": "8.0.20"
            }
        }, 
        "topologyMode": "Single-Primary"
    }, 
    "groupInformationSourceMember": "127.0.0.1:5306"
}
```
Is now understood by orchestrator, as two group secondary group members that replicate from the primary. In addition, the async/replica continues to be shown normally:

image

Notice also that the errant GTID problem is not shown.

The UI also shows group membership information, including:
* An icon showing whether the instance is a group member or not. A different text style is used for
  the primary and secondary instances so that they can be easily  identified.
* Hovering over the icon shows the instance state and role in the replication group.
* Group members that are not online are shown as with problems, as shown below:

image

On top of this, certain replication operations are prevented from taking place. Namely, attempting to relocate group secondaries to replicate from outside the group, since they actually replicate from the primary, as well as attempting to set-up a group primary to replicate from a secondary.


ejortegau:
cc @dveeden , @sjmudd , @luisyonaldo

dveeden:
I think group replication is an important feature for Orchestrator as it looks like this is at the core of the long term strategy of Oracle MySQL. Group Replication is a core component of InnoDB Cluster.

Note that *some* info about group replication is available in the same way as normal replication. There is nothing in `SHOW SLAVE STATUS`, but `performance_schema.replication_connection_status` etc show a `group_replication_applier` and a `group_replication_recovery` channel. Other newer / advanced replication features like multi source replication and parallel replication use the same performance_schema tables. So I think that where possible these should be used. For GR there are additional tables like `replication_group_members`.

@ejortegau, Could you squash your commits? Or are there multiple on purpose?

ejortegau:
> @ejortegau, Could you squash your commits? Or are there multiple on purpose?

There are multiple on purpose, as I have worked on it and added more and more support so far. I figure they would/could be squashed once the merge is accepted.

go/db/generate_patches.go
```diff
@@ -576,4 +576,40 @@ var generateSQLPatches = []string{
 			database_instance
 			ADD COLUMN semi_sync_available TINYINT UNSIGNED NOT NULL DEFAULT 0 AFTER semi_sync_enforced
 	`,
+	// Fields related to Replication Group the instance belongs to
+	`
+		ALTER TABLE
+			database_instance
+			ADD COLUMN replication_group_name VARCHAR(64) CHARACTER SET ascii NOT NULL DEFAULT '' AFTER gtid_mode
```
dveeden, line 588: So this is a UUID?
https://dev.mysql.com/doc/refman/5.7/en/group-replication-options.html#sysvar_group_replication_group_name

go/inst/instance_dao.go
```diff
@@ -478,6 +491,10 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
 				} else if masterInfoRepositoryOnTable {
 					_ = db.QueryRow("select count(*) > 0 and MAX(User_name) != '' from mysql.slave_master_info").Scan(&instance.ReplicationCredentialsAvailable)
 				}
+				// Populate GR information for the instance in Oracle MySQL 5.7+
+				if !instance.IsSmallerMajorVersionByString("5.7") {
```
dveeden: What about MariaDB? Should we exclude that?

dveeden:
Would this work ok with a group in multi primary mode?

https://dev.mysql.com/doc/refman/8.0/en/group-replication-multi-primary-mode.html

go/db/generate_patches.go
ejortegau: Yes, it is a UUID. I considered storing it with a binary, but chose not to in order to be consistent with how other UUIDs (e.g. `server_uuid`) are being stored in the orchestrator database.

go/inst/instance_dao.go
ejortegau: This is already in an if block with this:
```go
if (instance.IsOracleMySQL() || instance.IsPercona()) && !instance.IsSmallerMajorVersionByString("5.6") {
```
So MariaDB is  already excluded.

go/inst/instance_dao.go
```diff
@@ -478,6 +491,10 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
...
+					PopulateGroupReplicationInformation(instance, db)
+				}
```
luisyonaldo: add a new `waitGroup` and do this inside a new goroutine

ejortegau (commented):
> Would this work ok with a group in multi primary mode?
> 
> https://dev.mysql.com/doc/refman/8.0/en/group-replication-multi-primary-mode.html

It will not. I have missed pushing an update to the FAQ, indicating this. For now, this only works with single primary mode.

go/inst/instance_dao.go
```diff
@@ -3068,6 +3155,79 @@ func FigureInstanceKey(instanceKey *InstanceKey, thisInstanceKey *InstanceKey) (
 	return figuredKey, nil
 }
 
+// Obtain information about Group Replication  for this host as well as other hosts who are members of the same group
+// (if any). This method swallows up any errors, interpreting them as if the host does not support group replication
+// (e.g. because the group replication plugin is not loaded, therefore the corresponding performance schema tables and
+// global variables are not known to the server.
+func PopulateGroupReplicationInformation(instance *Instance, db *sql.DB) {
```
luisyonaldo: I'm not sure ignoring the error is the best way here, I would rather catch the error and ignore only the expected errors (no group replication plugin, etc), instead of ignoring all errors (which might bypass some error that we would like to see)

go/inst/instance_dao.go
```diff
@@ -744,6 +761,10 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
 	}
 
 	{
+		// wait for routines to finish before reading cluster attributes. This seems to be needed, as,
+		// otherwise, detecting whether a host is a replication group secondary to set up cluster attributes from its primary
+		// does not work. ToDo: Is there a better way to sync this?
+		waitGroup.Wait()
```
luisyonaldo: this seems wrong, why do you need to wait here?

go/inst/instance_dao.go
```diff
@@ -3068,6 +3155,79 @@ func FigureInstanceKey(instanceKey *InstanceKey, thisInstanceKey *InstanceKey) (
 	return figuredKey, nil
 }
 
+// Obtain information about Group Replication  for this host as well as other hosts who are members of the same group
```
luisyonaldo: `// PopulateGroupReplicationInformation  obtain information about ... `

go/inst/instance_dao.go
```diff
@@ -3068,6 +3155,79 @@ func FigureInstanceKey(instanceKey *InstanceKey, thisInstanceKey *InstanceKey) (
...
+		MEMBER_HOST,
+		MEMBER_PORT,
+		MEMBER_STATE, 
+		MEMBER_ROLE,
+		@@global.group_replication_group_name,
+		@@global.group_replication_single_primary_mode
+  	FROM
+		performance_schema.replication_group_members
+  	 `
+	rows, err := db.Query(q)
```
luisyonaldo, line 3188: why not `db.QueryRow`? https://golang.org/pkg/database/sql/#DB.QueryRow 
 I think is being used everywhere already, otherwise please add `defer rows.Close()`

go/inst/instance_dao.go
ejortegau: Without this, it would sometimes think that a group member was not actually a group member, because the `ReplicationGroupName` attribute was not populated yet. Hence I added the wait.

go/inst/instance_dao.go
ejortegau: Added the defer `rows.Close()`

dveeden:
For groups we should draw a line around them and maybe show the name of the group.
This is true for Group Replication, NDB and maybe PXC.

With D3.js this can be done with `d3.geom.hull`. Example in g1 in the image.

The `d3.layout.tree()` that is used doesn't allow more complex topologies like groups, multi source etc. This *could* be fixed by moving to `d3.layout.force()`. Example in g2 in the image.

I also notice that D3.js v3 is used, which is two versions behind the latest major release (D3.js v5).

Screenshot from 2020-06-05 15-21-07

The code for the example is in:
https://gist.github.com/dveeden/8a5cc4c1230f0442bb8b74c8e0eecad7

Some of this could be done outside of this PR, but we should make sure that the nodelist is populated with group info to allow the frontend part to be done later on.

go/inst/instance.go
```diff
@@ -204,6 +222,21 @@ func (this *Instance) IsNDB() bool {
...
 
+// IsReplicationGroup checks whether the host thinks it is part of a known replication group. Notice that this might
+// return True even if the group has decided to expel the member represented by this instance, as the instance might not
+// know that under certain circumstances
+func (this *Instance) IsReplicationGroupMember() bool {
+	return this.ReplicationGroupName != ""
+}
+
+func (this *Instance) IsReplicationGroupPrimary() bool {
+	return this.IsReplicationGroupMember() && this.ReplicationGroupPrimaryKey == this.Key
```
shlomi-noach: Please use `Equals()`
```suggestion
	return this.IsReplicationGroupMember() && this.ReplicationGroupPrimaryKey.Equals(&this.Key)
```

go/inst/instance.go
```diff
@@ -204,6 +222,21 @@ func (this *Instance) IsNDB() bool {
...
+}
+
+func (this *Instance) IsReplicationGroupSecondary() bool {
+	return this.IsReplicationGroupMember() && this.ReplicationGroupPrimaryKey != this.Key
```
shlomi-noach:
```suggestion
	return this.IsReplicationGroupMember() && !this.ReplicationGroupPrimaryKey.Equals(&this.Key)
```

go/inst/instance.go
```diff
@@ -271,7 +304,10 @@ func (this *Instance) IsReplica() bool {
 
 // IsMaster makes simple heuristics to decide whether this instance is a master (not replicating from any other server)
 func (this *Instance) IsMaster() bool {
-	return !this.IsReplica()
+	return !this.IsReplica() &&
+		((this.IsReplicationGroupMember() &&
+			this.ReplicationGroupMemberRole != GroupReplicationMemberRoleSecondary) ||
+			!this.IsReplicationGroupMember())
```
shlomi-noach: too much `&& || (( ))` complexity. I can hardly parse.
```suggestion
        if this.IsReplica() {
          return false
        }
	if !this.IsReplicationGroupMember() {
	  return false
	}
	if this.IsReplicationGroupMember() && this.ReplicationGroupMemberRole == GroupReplicationMemberRoleSecondary {
	  return false
	}
	return true
```

go/inst/instance.go
```diff
@@ -121,13 +121,31 @@ type Instance struct {
...
+	// Replication group members information. See
+	// https://dev.mysql.com/doc/refman/8.0/en/replication-group-members-table.html for details.
+	ReplicationGroupMemberState string
+	ReplicationGroupMemberRole  string
+
+	// List of all known members of the same group
+	ReplicationGroupMembers InstanceKeyMap
+
+	// Primary of the replication group
+	ReplicationGroupPrimaryKey InstanceKey
```
shlomi-noach: Because `PrimaryKey` is a term in databases, let's use `PrimaryInstanceKey`
```suggestion
	ReplicationGroupPrimaryInstanceKey InstanceKey
```

go/inst/instance_dao.go
shlomi-noach: This will hurt parallelism. I see the dependency problem. Let me see if moving `ReadInstanceClusterAttributes` itself can  break dependencies.

go/inst/instance_dao.go
```diff
@@ -866,12 +887,15 @@ Cleanup:
 		} else {
 			instance.AncestryUUID = fmt.Sprintf("%s,%s", instance.AncestryUUID, instance.ServerUUID)
 		}
+		// Add replication group ancestry UUID as well. Otherwise, Orchestrator thinks there are errant GTIDs in group
+		// members and its slaves, even though they are not.
+		instance.AncestryUUID = fmt.Sprintf("%s,%s", instance.AncestryUUID, instance.ReplicationGroupName)
```
shlomi-noach, line 909: is `ReplicationGroupName` a `UUID`?

shlomi-noach (changes requested):
Only an initial, partial review. I'll need to revisit.

go/inst/instance_dao.go
shlomi-noach: Agree

shlomi-noach: Is it not possible to validate whether this host at all supports group replication, and based on that finding to invoke/skip this function?

go/inst/instance_dao.go
```diff
@@ -3068,6 +3155,79 @@ func FigureInstanceKey(instanceKey *InstanceKey, thisInstanceKey *InstanceKey) (
...
+	SELECT
+		MEMBER_ID,
+		MEMBER_HOST,
+		MEMBER_PORT,
+		MEMBER_STATE, 
+		MEMBER_ROLE,
+		@@global.group_replication_group_name,
+		@@global.group_replication_single_primary_mode
+  	FROM
+		performance_schema.replication_group_members
```
shlomi-noach, line 3186: Does this require any special privileges? (I don't know)

go/inst/instance_topology.go
```diff
@@ -2695,10 +2695,19 @@ func RelocateBelow(instanceKey, otherKey *InstanceKey) (*Instance, error) {
 	if err != nil || !found {
 		return instance, log.Errorf("Error reading %+v", *instanceKey)
 	}
+	// Relocation of group secondaries makes no sense, group secondaries, by definition, always replicate from the group
+	// primary
+	if instance.IsReplicationGroupSecondary() {
```
shlomi-noach, line 2734: I'm thinking the same logic needs to be applied in `MoveBelow`, `MoveBelowGTID`, `MakeCoMaster`, `Repoint`, `TakeMaster`, `MatchBelow`, `MatchUp` functions.

resources/public/js/orchestrator.js
```diff
@@ -600,8 +600,17 @@ function normalizeInstance(instance) {
...
   instance.masterTitle = instance.MasterKey.Hostname + ":" + instance.MasterKey.Port;
-  instance.masterId = getInstanceId(instance.MasterKey.Hostname,
-    instance.MasterKey.Port);
+  // If this host is a replication group member, we set its masterId to the group primary, unless the instance is itself
+  // the primary. In that case, we set it to its async/semi-sync master (if configured). Notice that for group members
+  // whose role is not defined (e.g. because they are in ERROR state) we still set their master ID to the group primary.
+  // Setting the masterId to the group primary is what allows us to visualize group secondary members as replicating
+  // from the group primary.
+  if (instance.ReplicationGroupName != "" && (instance.ReplicationGroupMemberRole == "SECONDARY" || instance.ReplicationGroupMemberRole == ""))
+    masterKey = instance.ReplicationGroupPrimaryKey
```
shlomi-noach:
```suggestion
    masterKey = instance.ReplicationGroupPrimaryKey;
```

resources/public/js/orchestrator.js
```diff
@@ -600,8 +600,17 @@ function normalizeInstance(instance) {
...
+  else
+    masterKey = instance.MasterKey
```
shlomi-noach:
```suggestion
    masterKey = instance.MasterKey;
```

resources/public/js/orchestrator.js
```diff
@@ -600,8 +600,17 @@ function normalizeInstance(instance) {
...
+  instance.masterId = getInstanceId(masterKey.Hostname,
+    masterKey.Port);
```
shlomi-noach:
```suggestion
  instance.masterId = getInstanceId(masterKey.Hostname, masterKey.Port);
```

go/inst/instance_dao.go
```diff
@@ -1182,6 +1250,10 @@ func readInstanceRow(m sqlutils.RowMap) *Instance {
 	if instance.GtidErrant != "" {
 		instance.Problems = append(instance.Problems, "errant_gtid")
 	}
+	// Group replication problems
+	if instance.ReplicationGroupName != "" && instance.ReplicationGroupMemberState != "ONLINE" {
+		instance.Problems = append(instance.Problems, "group_replication_member_not_online")
+	}
```
shlomi-noach: Can we "upgrade" this to be analyzed in `instance.go`?

shlomi-noach (changes requested):
Some further comments. Review is not yet complete.

Thank you!

go/inst/instance_dao.go
luisyonaldo: this shouldn't be in a wait group since it has to be applied in sequential order, same as how `show slave status` query is read.

I would put this after `show slave status` code:
```go
instance.ReplicationIOThreadState = ReplicationThreadStateNoThread
instance.ReplicationSQLThreadState = ReplicationThreadStateNoThread
err = sqlutils.QueryRowsMap(db, "show slave status", func(m sqlutils.RowMap) error {
...
if err != nil {
	goto Cleanup
}
if instance.IsOracleMySQL() && !instance.IsSmallerMajorVersionByString("5.7") {
	err = PopulateGroupReplicationInformation(instance, db)
        if err != nil {
	        goto Cleanup
        }
}
```

go/inst/instance_dao.go
ejortegau: AFAIK just regular `SELECT`  on the relevant `performance_schema` tables. I will check to validate this assumption.

go/inst/instance_dao.go
ejortegau: I can move it there, though I thought this was a better place for it. From what I see, `instance.go` does not populate problems anywhere; instead, they seem to be populated here.

go/inst/instance_dao.go
ejortegau: Yes, we can check whether the plugin is installed and active, but this requires an extra query. I will modify this to check the error codes from the query and only ignore them if they are either because the group replication global variables or performance schema tables are not known, which would mean that the host does not support group replication.

go/inst/instance_dao.go
ejortegau: So this took me a while to fully understand, but the problem comes from the fact that in order to populate  GR attributes of the instance, the Server UUID needs to be known. This UUID is obtained asynchronously, so I added a wait group for it and defer execution of the `PopulateGroupReplicationInformation` until after the UUID is known. `PopulateGroupReplicationInformation` itself is run sequentially, as per @luisyonaldo 's suggestion.

go/inst/instance_topology.go
ejortegau: Will apply there.

ejortegau: Fixed

go/inst/instance_dao.go
ejortegau: Indeed, it is, see https://dev.mysql.com/doc/refman/5.7/en/group-replication-options.html#sysvar_group_replication_group_name

resources/public/js/orchestrator.js
ejortegau: Fixed

resources/public/js/orchestrator.js
ejortegau: Fixed

go/inst/instance.go
ejortegau: Fixed

go/inst/instance.go
ejortegau: Fixed.

go/inst/instance.go
ejortegau: Fixed

go/inst/instance.go
ejortegau: Fixed

resources/public/js/orchestrator.js
ejortegau: Moved into a single line instead

go/inst/instance_dao.go
ejortegau: Done.

go/inst/instance_dao.go
ejortegau: Confirmed, a user with only `SELECT` privilege in `performance_schema.replication_group_members` can run the query.

shlomi-noach: So, it _does_ require special privileges. Can you please document this in https://github.com/openark/orchestrator/blob/master/docs/configuration-discovery-basic.md ?

ejortegau: Done

shlomi-noach:
Sorry for the delays. I hope to be able to re-review and merge early next week (Sunday/Monday)

go/inst/analysis_dao.go
```diff
@@ -160,11 +160,19 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
...
-			OR substr(master_instance.master_host, 1, 2) = '//'
+			(
+				master_instance.master_host IN ('', '_')
+				OR master_instance.master_port = 0
+				OR substr(master_instance.master_host, 1, 2) = '//'
+			)
+			AND (
+				master_instance.replication_group_name = ''
+				OR master_instance.replication_group_member_role = 'PRIMARY'
+			)
```
shlomi-noach, line 175: :+1:

docs/faq.md
```diff
@@ -62,7 +62,21 @@ No.
...
+* The orchestrator web UI displays single primary group members. They are shown like this:
+    * All group secondary members as replicating from the primary.
+    * All group members have an icon that shows they are group members (as opposed to traditional async/semi-sync replicas).
+    * Hovering over the icon mentioned above provides information about the state and role of the DB instance in the
+      group.
+* Some relocation operations are forbidden for group members. In particular, orchestrator will refuse to relocate a secondary group member, as it, by definition, replicates always from the group primary. It will also reject an attempt to relocate a group primary under a secondary of the same group.
+
+No support has been added (yet) to handling group member failure. If all you have is a single replication group, this is fine, because you don't need it; the group will handle all failures as long as it can secure a majority.
+
+If, however, you have the primary of a group as a replica to another instance; or you have replicas under your group
```
shlomi-noach, line 78: :+1:

go/db/generate_patches.go
```diff
@@ -580,5 +580,41 @@ var generateSQLPatches = []string{
...
+	`,
+	`
+		ALTER TABLE
+		database_instance
+			ADD COLUMN replication_group_primary_host varchar(128) CHARACTER SET ascii NOT NULL DEFAULT '' AFTER replication_group_members
+	`,
+	`
+		ALTER TABLE
+		database_instance
+			ADD COLUMN replication_group_primary_port smallint(5) unsigned NOT NULL DEFAULT 0 AFTER replication_group_primary_host
```
shlomi-noach, line 618: :+1:

go/inst/instance.go
```diff
@@ -129,13 +129,32 @@ type Instance struct {
...
+	// Replication group members information. See
+	// https://dev.mysql.com/doc/refman/8.0/en/replication-group-members-table.html for details.
+	ReplicationGroupMemberState string
+	ReplicationGroupMemberRole  string
+
+	// List of all known members of the same group
+	ReplicationGroupMembers InstanceKeyMap
+
+	// Primary of the replication group
+	ReplicationGroupPrimaryInstanceKey InstanceKey
```
shlomi-noach, line 148: :+1:

go/inst/instance.go
```diff
@@ -129,13 +129,32 @@ type Instance struct {
...
 
 // NewInstance creates a new, empty instance
+
 func NewInstance() *Instance {
 	return &Instance{
-		Replicas: make(map[InstanceKey]bool),
-		Problems: []string{},
+		Replicas:                make(map[InstanceKey]bool),
+		ReplicationGroupMembers: make(map[InstanceKey]bool),
+		Problems:                []string{},
```
shlomi-noach, line 157: :+1:

go/inst/instance.go
```diff
@@ -228,6 +247,21 @@ func (this *Instance) IsNDB() bool {
...
+	return this.ReplicationGroupName != ""
+}
+
+func (this *Instance) IsReplicationGroupPrimary() bool {
+	return this.IsReplicationGroupMember() && this.ReplicationGroupPrimaryInstanceKey.Equals(&this.Key)
+}
+
+func (this *Instance) IsReplicationGroupSecondary() bool {
+	return this.IsReplicationGroupMember() && !this.ReplicationGroupPrimaryInstanceKey.Equals(&this.Key)
+}
```
shlomi-noach, line 263: :+1:

go/inst/instance.go
```diff
@@ -293,9 +327,27 @@ func (this *Instance) IsReplica() bool {
...
+		} else {
+			// If traditional replication is not configured, and this host is part of a group, it is only considered a
+			// master if it has the role of group Primary. Otherwise it is not a master.
+			if this.ReplicationGroupMemberRole == GroupReplicationMemberRolePrimary {
+				return true
+			} else {
+				return false
+			}
+		}
+	}
```
shlomi-noach, line 336: I'm replacing this code:
```golang
func (this *Instance) IsMaster() bool {
	// If traditional replication is configured, it is for sure not a master
	if this.IsReplica() {
		return false
	} else {
		// If traditional replication is not configured, and it is also not part of a replication group, this host is
		// a master
		if !this.IsReplicationGroupMember() {
			return true
		} else {
			// If traditional replication is not configured, and this host is part of a group, it is only considered a
			// master if it has the role of group Primary. Otherwise it is not a master.
			if this.ReplicationGroupMemberRole == GroupReplicationMemberRolePrimary {
				return true
			} else {
				return false
			}
		}
	}
}
```
with simplified nesting:
```golang
func (this *Instance) IsMaster() bool {
	// If traditional replication is configured, it is for sure not a master
	if this.IsReplica() {
		return false
	}
	// If traditional replication is not configured, and it is also not part of a replication group, this host is
	// a master
	if !this.IsReplicationGroupMember() {
		return true
	}
	// If traditional replication is not configured, and this host is part of a group, it is only considered a
	// master if it has the role of group Primary. Otherwise it is not a master.
	if this.ReplicationGroupMemberRole == GroupReplicationMemberRolePrimary {
		return true
	}
	return false
}
```

go/inst/instance.go
```diff
@@ -358,6 +407,11 @@ func (this *Instance) AddReplicaKey(replicaKey *InstanceKey) {
 	this.Replicas.AddKey(*replicaKey)
 }
 
+// AddGroupMemberKey adds a group member to the list of this instance's group members.
+func (this *Instance) AddGroupMemberKey(groupMemberKey *InstanceKey) {
+	this.ReplicationGroupMembers.AddKey(*groupMemberKey)
+}
```
shlomi-noach, line 413: :+1:

go/inst/instance_topology.go
```diff
@@ -485,6 +485,11 @@ func MoveBelow(instanceKey, siblingKey *InstanceKey) (*Instance, error) {
 	if err != nil {
 		return instance, err
 	}
+	// Relocation of group secondaries makes no sense, group secondaries, by definition, always replicate from the group
+	// primary
+	if instance.IsReplicationGroupSecondary() {
+		return instance, log.Errorf("MoveBelow: %+v is a secondary replication group member, hence, it cannot be relocated", instance.Key)
+	}
```
shlomi-noach, line 492: :+1:

go/inst/instance_topology.go
```diff
@@ -656,6 +661,11 @@ func MoveBelowGTID(instanceKey, otherKey *InstanceKey) (*Instance, error) {
 	if err != nil {
 		return instance, err
 	}
+	// Relocation of group secondaries makes no sense, group secondaries, by definition, always replicate from the group
+	// primary
+	if instance.IsReplicationGroupSecondary() {
+		return instance, log.Errorf("MoveBelowGTID: %+v is a secondary replication group member, hence, it cannot be relocated", instance.Key)
+	}
```
shlomi-noach, line 668: :+1:

go/inst/instance_topology.go
```diff
@@ -766,7 +776,11 @@ func Repoint(instanceKey *InstanceKey, masterKey *InstanceKey, gtidHint Operatio
 	if !instance.IsReplica() {
 		return instance, fmt.Errorf("instance is not a replica: %+v", *instanceKey)
 	}
-
+	// Relocation of group secondaries makes no sense, group secondaries, by definition, always replicate from the group
+	// primary
+	if instance.IsReplicationGroupSecondary() {
+		return instance, fmt.Errorf("repoint: %+v is a secondary replication group member, hence, it cannot be relocated", instance.Key)
+	}
```
shlomi-noach, line 783: :+1:

go/inst/instance_topology.go
```diff
@@ -926,6 +940,11 @@ func MakeCoMaster(instanceKey *InstanceKey) (*Instance, error) {
 	if err != nil {
 		return instance, err
 	}
+	// Relocation of group secondaries makes no sense, group secondaries, by definition, always replicate from the group
+	// primary
+	if instance.IsReplicationGroupSecondary() {
+		return instance, fmt.Errorf("MakeCoMaster: %+v is a secondary replication group member, hence, it cannot be relocated", instance.Key)
+	}
```
shlomi-noach, line 947: :+1:

go/inst/instance_topology.go
```diff
@@ -1522,6 +1541,11 @@ func MatchBelow(instanceKey, otherKey *InstanceKey, requireInstanceMaintenance b
 	if err != nil {
 		return instance, nil, err
 	}
+	// Relocation of group secondaries makes no sense, group secondaries, by definition, always replicate from the group
+	// primary
+	if instance.IsReplicationGroupSecondary() {
+		return instance, nil, fmt.Errorf("MatchBelow: %+v is a secondary replication group member, hence, it cannot be relocated", *instanceKey)
+	}
```
shlomi-noach, line 1548: :+1:

go/inst/instance_topology.go
```diff
@@ -1731,6 +1755,11 @@ func TakeMaster(instanceKey *InstanceKey, allowTakingCoMaster bool) (*Instance,
 	if err != nil {
 		return instance, err
 	}
+	// Relocation of group secondaries makes no sense, group secondaries, by definition, always replicate from the group
+	// primary
+	if instance.IsReplicationGroupSecondary() {
+		return instance, fmt.Errorf("takeMaster: %+v is a secondary replication group member, hence, it cannot be relocated", instance.Key)
+	}
```
shlomi-noach, line 1762: :+1:

go/inst/instance_topology.go
```diff
@@ -2026,6 +2055,11 @@ func MatchUp(instanceKey *InstanceKey, requireInstanceMaintenance bool) (*Instan
 	if !instance.IsReplica() {
 		return instance, nil, fmt.Errorf("instance is not a replica: %+v", instanceKey)
 	}
+	// Relocation of group secondaries makes no sense, group secondaries, by definition, always replicate from the group
+	// primary
+	if instance.IsReplicationGroupSecondary() {
+		return instance, nil, fmt.Errorf("MatchUp: %+v is a secondary replication group member, hence, it cannot be relocated", instance.Key)
+	}
```
shlomi-noach, line 2062: :+1:

go/inst/instance_topology.go
```diff
@@ -2695,10 +2729,19 @@ func RelocateBelow(instanceKey, otherKey *InstanceKey) (*Instance, error) {
 	if err != nil || !found {
 		return instance, log.Errorf("Error reading %+v", *instanceKey)
 	}
+	// Relocation of group secondaries makes no sense, group secondaries, by definition, always replicate from the group
+	// primary
+	if instance.IsReplicationGroupSecondary() {
+		return instance, log.Errorf("relocate: %+v is a secondary replication group member, hence, it cannot be relocated", instance.Key)
+	}
```
shlomi-noach, line 2736: :+1:

go/inst/instance_topology.go
```diff
@@ -2695,10 +2729,19 @@ func RelocateBelow(instanceKey, otherKey *InstanceKey) (*Instance, error) {
...
 	other, found, err := ReadInstance(otherKey)
 	if err != nil || !found {
 		return instance, log.Errorf("Error reading %+v", *otherKey)
 	}
+	// Disallow setting up a group primary to replicate from a group secondary
+	if instance.IsReplicationGroupPrimary() && other.ReplicationGroupName == instance.ReplicationGroupName {
+		return instance, log.Errorf("relocate: Setting a group primary to replicate from another member of its group is disallowed")
+	}
```
shlomi-noach, line 2744: :+1:

go/logic/orchestrator.go
```diff
@@ -279,8 +279,8 @@ func DiscoverInstance(instanceKey inst.InstanceKey) {
 		return
 	}
 
-	// Investigate replicas:
-	for _, replicaKey := range instance.Replicas.GetInstanceKeys() {
+	// Investigate replicas and members of the same replication group:
+	for _, replicaKey := range append(instance.ReplicationGroupMembers.GetInstanceKeys(), instance.Replicas.GetInstanceKeys()...) {
```
shlomi-noach, line 283: :+1:

resources/public/js/cluster-analysis-shared.js
```diff
@@ -46,5 +46,9 @@ var errorMapping = {
 	"errant_gtid": {
 		"badge": "label-errant",
 		"description": "Errant GTID"
+	},
+	"group_replication_member_not_online": {
+		"badge": "label-danger",
+		"description": "Replication group member is not ONLINE"
```
shlomi-noach, line 52: :+1:

shlomi-noach, line 52: :+1:

resources/public/js/orchestrator.js
```diff
@@ -600,8 +600,16 @@ function normalizeInstance(instance) {
...
-    instance.MasterKey.Port);
+  // If this host is a replication group member, we set its masterId to the group primary, unless the instance is itself
+  // the primary. In that case, we set it to its async/semi-sync master (if configured). Notice that for group members
+  // whose role is not defined (e.g. because they are in ERROR state) we still set their master ID to the group primary.
+  // Setting the masterId to the group primary is what allows us to visualize group secondary members as replicating
+  // from the group primary.
+  if (instance.ReplicationGroupName != "" && (instance.ReplicationGroupMemberRole == "SECONDARY" || instance.ReplicationGroupMemberRole == ""))
+    masterKey = instance.ReplicationGroupPrimaryInstanceKey;
+  else
+    masterKey = instance.MasterKey;
```
shlomi-noach, line 611: :+1:

resources/public/js/orchestrator.js
```diff
@@ -665,6 +673,9 @@ function normalizeInstanceProblem(instance) {
   instance.errantGTIDProblem = function() {
     return instanceProblemIfExists('errant_gtid');
   }
+  instance.replicationGroupMemberStateProblem = function() {
+    return instanceProblemIfExists("group_replication_member_not_online")
+  }
```
shlomi-noach, line 678: :+1:

resources/public/js/orchestrator.js
```diff
@@ -691,6 +702,9 @@ function normalizeInstanceProblem(instance) {
   } else if (instance.errantGTIDProblem()) {
     instance.problemDescription = "Replica has GTID entries not found on its master";
     instance.problemOrder = 6;
+  } else if (instance.replicationGroupMemberStateProblem()) {
+    instance.problemDescription = "Replication group member in state " + instance.ReplicationGroupMemberState;
+    instance.problemOrder = 7;
```
shlomi-noach, line 707: :+1:

resources/public/js/orchestrator.js
```diff
@@ -908,6 +922,16 @@ function renderInstanceElement(popoverElement, instance, renderType) {
...
       popoverElement.find("h3 div.pull-right").prepend('<span class="glyphicon glyphicon-forward" title="Logs replication updates"></span> ');
     }
+    // Icons for GR
+    var text_style;
+    if (instance.ReplicationGroupName != "") {
+      if (instance.ReplicationGroupMemberRole == "PRIMARY")
+        text_style = "";
+      else
+        text_style = "text-muted";
+      popoverElement.find("h3 div.pull-right").prepend('<span class="glyphicon '+ text_style +' glyphicon-tower" title="Group replication '+ instance.ReplicationGroupMemberRole + ' ' + instance.ReplicationGroupMemberState +'"></span> ');
```
shlomi-noach, line 932: :+1:

resources/public/js/orchestrator.js
```diff
@@ -940,6 +964,15 @@ function renderInstanceElement(popoverElement, instance, renderType) {
...
       instance.renderHint = "warning";
+    } else if (instance.replicationGroupMemberStateProblem()) {
+      switch (instance.ReplicationGroupMemberState) {
+        case "RECOVERING":
+          instance.renderHint = "warning";
+          break;
+        default:
+          instance.renderHint = "danger";
+          break;
+      }
```
shlomi-noach, line 975: :+1:

go/inst/instance_dao.go
```diff
@@ -67,6 +68,29 @@ func (this InstancesByCountReplicas) Less(i, j int) bool {
...
+// See https://dev.mysql.com/doc/refman/8.0/en/replication-group-members-table.html for additional information.
+const (
+	// Group member roles
+	GroupReplicationMemberRolePrimary   = "PRIMARY"
+	GroupReplicationMemberRoleSecondary = "SECONDARY"
+	// Group member states
+	GroupReplicationMemberStateOnline     = "ONLINE"
+	GroupReplicationMemberStateRecovering = "RECOVERING"
+	GroupReplicationMemberStateOffline    = "OFFLINE"
+	GroupReplicationMemberStateError      = "ERROR"
```
shlomi-noach, line 81: :+1:

go/inst/instance_dao.go
```diff
@@ -67,6 +68,29 @@ func (this InstancesByCountReplicas) Less(i, j int) bool {
...
+
+// We use this map to identify whether the query failed because the server does not support group replication or due
+// to a different reason.
+var GroupReplicationNotSupportedErrors = map[uint16]bool{
+	// If either the group replication global variables are not known or the
+	// performance_schema.replication_group_members table does not exist, the host does not support group
+	// replication, at least in the form supported here.
+	1193: true, // ERROR: 1193 (HY000): Unknown system variable 'group_replication_group_name'
+	1146: true, // ERROR: 1146 (42S02): Table 'performance_schema.replication_group_members' doesn't exist
+}
```
shlomi-noach, line 92: :+1:

go/inst/instance_dao.go
```diff
@@ -580,6 +607,17 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
 	if err != nil {
 		goto Cleanup
 	}
+	// Populate GR information for the instance in Oracle MySQL 8.0+. To do this we need to wait for the Server UUID to
+	// be populated to be able to find this instance's information in performance_schema.replication_group_members by
+	// comparing UUIDs. We could instead resolve the MEMBER_HOST and MEMBER_PORT columns into an InstanceKey and compare
+	// those instead, but this could require external calls for name resolving, whereas comparing UUIDs does not.
+	serverUuidWaitGroup.Wait()
```
shlomi-noach, line 614: :+1:

go/inst/instance_dao.go
```diff
@@ -580,6 +607,17 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
...
+	if instance.IsOracleMySQL() && !instance.IsSmallerMajorVersionByString("8.0") {
+		err := PopulateGroupReplicationInformation(instance, db)
+		if err != nil {
+			goto Cleanup
+		}
+	}
```
shlomi-noach, line 620: :+1:

go/inst/instance_dao.go
```diff
@@ -866,12 +904,15 @@ Cleanup:
 		} else {
 			instance.AncestryUUID = fmt.Sprintf("%s,%s", instance.AncestryUUID, instance.ServerUUID)
 		}
+		// Add replication group ancestry UUID as well. Otherwise, Orchestrator thinks there are errant GTIDs in group
+		// members and its slaves, even though they are not.
+		instance.AncestryUUID = fmt.Sprintf("%s,%s", instance.AncestryUUID, instance.ReplicationGroupName)
```
shlomi-noach, line 909: :+1:

go/inst/instance_dao.go
```diff
@@ -943,20 +984,45 @@ func ReadClusterAliasOverride(instance *Instance) (err error) {
...
+		resolvedGroupPrimary, err := NewResolveInstanceKey(groupPrimaryHost, groupPrimaryPort)
+		if err != nil {
+			return err
+		}
+		instance.ReplicationGroupPrimaryInstanceKey = *resolvedGroupPrimary
+		return nil
+	})
+	return err
+}
+
```
shlomi-noach, line 1011: :+1:

go/inst/instance_dao.go
```diff
@@ -943,20 +984,45 @@ func ReadClusterAliasOverride(instance *Instance) (err error) {
...
-	var masterReplicationDepth uint
+	var masterOrGroupPrimaryInstanceKey InstanceKey
+	var masterOrGroupPrimaryClusterName string
+	var masterOrGroupPrimarySuggestedClusterAlias string
+	var masterOrGroupPrimaryReplicationDepth uint
 	var ancestryUUID string
-	var masterExecutedGtidSet string
-	masterDataFound := false
+	var masterOrGroupPrimaryExecutedGtidSet string
+	masterOrGroupPrimaryDataFound := false
```
shlomi-noach, line 1023: :+1:  to be later all renamed as: `s/masterOrGroupPrimary/primary`

go/inst/instance_dao.go
```diff
@@ -969,17 +1035,24 @@ func ReadInstanceClusterAttributes(instance *Instance) (err error) {
...
-	args := sqlutils.Args(instance.MasterKey.Hostname, instance.MasterKey.Port)
-
+	// For instances that are part of a replication group, if the host is not the group's primary, we use the
+	// information from the group primary. If it is the group primary, we use the information of its master
+	// (if it has any). If it is not a group member, we use the information from the host's master.
+	if instance.IsReplicationGroupSecondary() {
+		masterOrGroupPrimaryInstanceKey = instance.ReplicationGroupPrimaryInstanceKey
+	} else {
+		masterOrGroupPrimaryInstanceKey = instance.MasterKey
+	}
```
shlomi-noach, line 1045: :+1:

go/inst/instance_dao.go
```diff
@@ -988,9 +1061,9 @@ func ReadInstanceClusterAttributes(instance *Instance) (err error) {
 
 	var replicationDepth uint = 0
 	var clusterName string
-	if masterDataFound {
-		replicationDepth = masterReplicationDepth + 1
-		clusterName = masterClusterName
+	if masterOrGroupPrimaryDataFound {
+		replicationDepth = masterOrGroupPrimaryReplicationDepth + 1
+		clusterName = masterOrGroupPrimaryClusterName
```
shlomi-noach, line 1066: :+1:

go/inst/instance_dao.go
```diff
@@ -1015,11 +1088,11 @@ func ReadInstanceClusterAttributes(instance *Instance) (err error) {
...
 	}
 	instance.ClusterName = clusterName
-	instance.SuggestedClusterAlias = masterSuggestedClusterAlias
+	instance.SuggestedClusterAlias = masterOrGroupPrimarySuggestedClusterAlias
 	instance.ReplicationDepth = replicationDepth
 	instance.IsCoMaster = isCoMaster
 	instance.AncestryUUID = ancestryUUID
-	instance.masterExecutedGtidSet = masterExecutedGtidSet
+	instance.masterExecutedGtidSet = masterOrGroupPrimaryExecutedGtidSet
 	return nil
```
shlomi-noach, line 1096: :+1:

go/inst/instance_dao.go
```diff
@@ -1182,6 +1265,10 @@ func readInstanceRow(m sqlutils.RowMap) *Instance {
 	if instance.GtidErrant != "" {
 		instance.Problems = append(instance.Problems, "errant_gtid")
 	}
+	// Group replication problems
+	if instance.ReplicationGroupName != "" && instance.ReplicationGroupMemberState != GroupReplicationMemberStateOnline {
+		instance.Problems = append(instance.Problems, "group_replication_member_not_online")
+	}
```
shlomi-noach, line 1271: :+1:

go/inst/instance_dao.go
```diff
@@ -2440,6 +2528,13 @@ func mkInsertOdkuForInstances(instances []*Instance, instanceWasActuallyFound bo
 		"semi_sync_replica_status",
 		"instance_alias",
 		"last_discovery_latency",
+		"replication_group_name",
+		"replication_group_is_single_primary_mode",
+		"replication_group_member_state",
+		"replication_group_member_role",
+		"replication_group_members",
+		"replication_group_primary_host",
+		"replication_group_primary_port",
```
shlomi-noach, line 2537: :+1:

go/inst/instance_dao.go
```diff
@@ -2526,6 +2621,13 @@ func mkInsertOdkuForInstances(instances []*Instance, instanceWasActuallyFound bo
 		args = append(args, instance.SemiSyncReplicaStatus)
 		args = append(args, instance.InstanceAlias)
 		args = append(args, instance.LastDiscoveryLatency.Nanoseconds())
+		args = append(args, instance.ReplicationGroupName)
+		args = append(args, instance.ReplicationGroupIsSinglePrimary)
+		args = append(args, instance.ReplicationGroupMemberState)
+		args = append(args, instance.ReplicationGroupMemberRole)
+		args = append(args, instance.ReplicationGroupMembers.ToJSONString())
+		args = append(args, instance.ReplicationGroupPrimaryInstanceKey.Hostname)
+		args = append(args, instance.ReplicationGroupPrimaryInstanceKey.Port)
```
shlomi-noach, line 2630: :+1:

go/inst/instance_dao.go
```diff
@@ -3068,6 +3170,92 @@ func FigureInstanceKey(instanceKey *InstanceKey, thisInstanceKey *InstanceKey) (
...
+	// likely to have been expelled from the group. Still, try to find out the primary of the group and set it for the
+	// instance being discovered, so that it is identified as part of the same cluster
+	if !foundGroupPrimary {
+		err = ReadReplicationGroupPrimary(instance)
+		if err != nil {
+			return log.Errorf("Unable to find the group primary of instance %+v even though it seems to be "+
+				"part of a replication group", instance.Key)
+		}
+	}
+	return nil
```
shlomi-noach, line 3256: :+1:

shlomi-noach (approved):
Approved. I've made a few formatting changes.

earl86:
I find when MGR changed master, orchestrator has no call PostFailoverProcesses hooks.

earl86:
orchestrator  monitor mysql group replication(MGR),when  MGR PRIMARY changed, can orchestrator   update consul key-value info for this MGR cluster?

sjmudd:
For GR members there are no specific hooks yet to handle a gr member failure. This is a current known limitation so if you have asynchronous replicas under a failed GR member then no action will currently be taken.
That does need to be addressed but has not been handled yet.

PR #1184, opened 2020-06-04
Format delays in days / hours / minutes / seconds

sjmudd:
Delays are shown in seconds which is fine but often you want to see this in days / hours / minutes and seconds without doing mental arithmetic to figure out the delay.

This patch shows time since a box has been seen and replication delay in days / hours / minutes and seconds and thus saves time with a large delay in seeing what it represents.

I'll include a couple of examples with the new formatting.
screenshot-1
screenshot-2
screenshot-3

I'm exposing lots of warts :-)


sjmudd:
cc: @luisyonaldo @dveeden

koorgoo:
Did you insert spaces between units intentionally, i.e. `1d 2h 3m 0s`?
Don't you want the function to work similar to https://golang.org/pkg/time/#Duration.String and return `1d2h3m0s`?

dveeden:
Wouldn't there be some JS function that does what `formattedInterval()` does? 

Otherwise: Useful functionality and code looks good to me, but didn't test it. 

side note: You know that instead of editing screenshots you can get orchestrator to mask hostnames?

dveeden:
> Wouldn't there be some JS function that does what `formattedInterval()` does?

I've searched and couldn't find anything that's obviously better.

shlomi-noach:
I believe JavaScript has nothing to offer. But we should use
golang's existing formatting.

koorgoo:
> But we should use golang's existing formatting.

Then we should use only **h**ours, **m**inutes, and **s**econds.

And maybe rename `formattedInterval()` to `durationString()`.

sjmudd:
To answer the questions:
* spaces are intentional. I think they read better.  YMMV.
* I have no clue about javascript so am quite happy to be provided a better solution. I just want to see this using a more human readable output.
* I am just sick of seeing unformatted seconds.
* the patch seemed reasonable and I'm sure we can improve it later but as the first attempt this looks reasonable.
* Dima if you don't like the name I can change it.

koorgoo:
The current implementation already looks good to me 👌

luisyonaldo (approved):

shlomi-noach (approved):

PR #1188, opened 2020-06-09
API: transition into new terminology

shlomi-noach:
Followup to #1182 and #1187 : this PR changes the `Instance` API by _adding_ new fields using "replica" terminology, on top of existing fields.
This is done by adding `MarshalJSON()` function which, at point of marshaling, updates new-terminology fields to the proper values. 

This is step 1 in transitioning into new terminology.

Clients are now able to use new terminology and drop usage of old terminology.

Future followups will be:
- Switch internally to new terminology, reverse marshaling logic
- Removal of old terminology fields. This is in the longer future since all users will have to first switch to new terminology, or else `orchestrator` breaks.


shlomi-noach:
It's best to review the changes hiding whitespace diffs, see https://github.com/openark/orchestrator/pull/1188/files?diff=unified&w=1

PR #1191, opened 2020-06-12
Skip Verify should be SSLSkipVerify Instead the mysql backend config.

martinarrieta:
Hello.

Related issue: https://github.com/openark/orchestrator/issues/dev/null (couldn't create an issue :D)

### Description

This PR fixes the configuration in the Raft http_client setup. Currently to skip the validation for raft SSL setup is with the variable MySQLOrchestratorSSLSkipVerify which should be used only to skip the validation against the MySQL backend for orchestrator. Instead, I have changed the variable for SSLSkipVerify which I guess is the right one.

### DIFF
--- a/go/raft/http_client.go
+++ b/go/raft/http_client.go
@@ -40,7 +40,7 @@ func setupHttpClient() error {
 	}
 
 	tlsConfig := &tls.Config{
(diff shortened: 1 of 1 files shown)


shlomi-noach (approved):
Good catch. Thank you!

PR #1194, opened 2020-06-16
Strict default value in backend table

shlomi-noach:
Strictly setting NULL as default value for a TIMESTAMP column. This is in favor of the vitess parser


shlomi-noach:
Related:
- https://github.com/vitessio/vitess/issues/6329
- https://github.com/vitessio/vitess/issues/6330

shlomi-noach:
I should probably note that there is no intention to run `orchestrator` on top of `vitess`, and this PR is the result of some unrelated experimentation for education purposes.

PR #1201, opened 2020-06-26
fix CountDistinctOkInstanceKeys always 0

MOON-CLJ:
### Description

CountDistinctOkInstanceKeys is is not assigned. This PR fix it。

@shlomi-noach @sjmudd cc @yangeagle


shlomi-noach (approved):
@sjmudd this looks to be in your neighborhood ; but the PR looks legit.

Issue #1203, opened 2020-07-03
bug report，InstanceWriteBufferSize cant be more than about 1000


MOON-CLJ:
If this is a bug report, please provide a test case and the error output. Useful information:

 - your orchestrator.conf.json config file/contents

most related

```
InstanceWriteBufferSize=10000
BufferInstanceWrites=true
InstanceFlushIntervalMilliseconds =10000
```

 - what did you expect to happen?

database_service table's record be updated every InstanceFlushIntervalMilliseconds。

 - what happened?

database_service table's record dont update for long time。

 - orchestrator logs; please use `--debug --stack` for maximum verbosity

```
2020-07-03 22:31:46 ERROR goroutine 22 flushInstanceWriteBuffer: 2020-07-03 22:31:46 ERROR goroutine 22 flushInstanceWriteBuffer last_seen: Error 1390: Prepared statement contains too many placeholders
2020-07-03 22:31:56 ERROR goroutine 22 Error 1390: Prepared statement contains too many placeholders
2020-07-03 22:31:56 ERROR goroutine 22 flushInstanceWriteBuffer last_seen: Error 1390: Prepared statement contains too many placeholders
2020-07-03 22:31:56 ERROR goroutine 22 flushInstanceWriteBuffer: 2020-07-03 22:31:56 ERROR goroutine 22 flushInstanceWriteBuffer last_seen: Error 1390: Prepared statement contains too many placeholders
2020-07-03 22:32:07 ERROR goroutine 22 Error 1390: Prepared statement contains too many placeholders
2020-07-03 22:32:07 ERROR goroutine 22 flushInstanceWriteBuffer last_seen: Error 1390: Prepared statement contains too many placeholders
2020-07-03 22:32:07 ERROR goroutine 22 flushInstanceWriteBuffer: 2020-07-03 22:32:07 ERROR goroutine 22 flushInstanceWriteBuffer last_seen: Error 1390: Prepared statement contains too many placeholders
```

After a long debug，we found the above anomaly in our logs.

so the problem's InstanceWriteBufferSize=10000 is so big that mkInsertOdku ran into Error 1390. After a quick search，https://stackoverflow.com/a/24447922, There's a limit 65535 of  "Prepared statement placeholders"，

we sugguest to check that InstanceWriteBufferSize does not exceed 1000 when loading the configuration，because one instance have about 64 columns/placaholders。

@shlomi-noach cc @yangeagle @sjmudd

sjmudd:
[...]

PR #1204, opened 2020-07-05
Provide some suggestions for larger orchestrator environments

sjmudd:
Add some suggestions on settings for orchestrator in a large environment.


MOON-CLJ:
In addition to https://github.com/openark/orchestrator/issues/1203, we have one more modification。

we comment this line out.
https://github.com/openark/orchestrator/blob/7bf32d18d182de3459542d1444a2126646b9c4b1/go/logic/topology_recovery.go#L916 

because we do frequent integration tests in our product environment. and because we patch AuditPurgeDays to 31(We've already revert it to 7)，we have a lots of data in our backend db。

```
select count(*) from topology_recovery;
+----------+
| count(*) |
+----------+
|    62421 |
+----------+
1 row in set (0.02 sec)

select count(*) from topology_failure_detection;
+----------+
| count(*) |
+----------+
|    95159 |
+----------+
1 row in set (0.02 sec)
```

Snapshot reads the full amount of data in the backend db as well as the raft state each time, taking up a lot of memory and golang gc resources. Then takes up a lot of goroutine scheduling resources, then raft heartbeat timeouts and frequent elections.

So we only do snapshots periodically now.
https://github.com/openark/orchestrator/blob/7bf32d18d182de3459542d1444a2126646b9c4b1/go/raft/raft.go#L44

maybe we need a config for this behavior https://github.com/openark/orchestrator/blob/7bf32d18d182de3459542d1444a2126646b9c4b1/go/logic/topology_recovery.go#L916

cc @shlomi-noach

sjmudd:
Hm. The snapshot thing I have configured without touching code.

`  "SnapshotTopologiesIntervalHours": 0,`

Judging from the [configuration description](https://github.com/openark/orchestrator/blob/master/go/config/config.go#L146) I think this is sufficient. You should not need to change the code to make orchestrator work the way you need it to.

If extra configuration is required I'm sure that @shlomi-noach would prefer you providing a PR with a new configuration setting to make the setting configurable as needed.  The process for doing that is quite straight forward and this is much better long term.  Having had experience of maintaining a patched fork of orchestrator I would really recommend that you push changes back upstream to Shlomi explaining why they are needed and if they make sense I'm sure he'll accept them.

shlomi-noach:
CI fails because there's no incoming link to this new documentation page. Please add a link in [configuration.md](https://github.com/openark/orchestrator/blob/master/docs/configuration.md)

sjmudd:
@shlomi-noach good morning. I hadn't realised this was necessary and it's good to require that, so should be fixed I think with the extra commit I've pushed.

MOON-CLJ:
> 

https://github.com/openark/orchestrator/blob/7bf32d18d182de3459542d1444a2126646b9c4b1/go/logic/topology_recovery.go#L916

> Hm. The snapshot thing I have configured without touching code. "SnapshotTopologiesIntervalHours": 0,

@sjmudd SnapshotTopologiesIntervalHours is no the same thing with snapshotInterval，snapshotInterval was about raft snapshot.

so what we suggest is to add a configuration item here below
https://github.com/openark/orchestrator/blob/7bf32d18d182de3459542d1444a2126646b9c4b1/go/logic/topology_recovery.go#L916

shlomi-noach (approved):

Issue #1205, opened 2020-07-07
semi-sync: Out of range value for column 'semi_sync_master_timeout'


luisyonaldo:
`rpl_semi_sync_master_timeout` should be `bigint` instead of `int`. Since this value is in milliseconds, is reasonable to set this for a value bigger than `int`.

Error:
```error
orchestrator[14049]: flushInstanceWriteBuffer last_seen: Error 1264: Out of range value for column 'semi_sync_master_timeout' at row 143
orchestrator[14049]: Error 1264: Out of range value for column 'semi_sync_master_timeout' at row 143
orchestrator[14049]: flushInstanceWriteBuffer: 2020-07-07 14:26:59 ERROR flushInstanceWriteBuffer last_seen: Error 1264: Out of range value for column 'semi_sync_master_timeout' at row 143
```

sjmudd:
The setting being used on the server that triggered this and which is accepted by MySQL was:

```
rpl_semi_sync_master_timeout=10000000000
```

PR #1207, opened 2020-07-07
fix out of range in semi_sync_master_timeout

luisyonaldo:
Solves issue: https://github.com/openark/orchestrator/issues/1205

### Description

This PR  sets `semi_sync_master_timeout` as `bigint` to avoid out of range (overflow) errors.

### DIFF
--- a/go/db/generate_patches.go
+++ b/go/db/generate_patches.go
@@ -577,7 +577,7 @@ var generateSQLPatches = []string{
 			ADD COLUMN semi_sync_available TINYINT UNSIGNED NOT NULL DEFAULT 0 AFTER semi_sync_enforced
 	`,
 	`
-		ALTER TABLE
+		ALTER TABLE /* sqlite3-skip */
 			database_instance
 			MODIFY COLUMN semi_sync_master_timeout BIGINT UNSIGNED NOT NULL DEFAULT 0
 	`,
--- a/go/db/generate_patches.go
+++ b/go/db/generate_patches.go
@@ -579,6 +579,6 @@ var generateSQLPatches = []string{
 	`
 		ALTER TABLE /* sqlite3-skip */
 			database_instance
-			MODIFY COLUMN semi_sync_master_timeout BIGINT UNSIGNED NOT NULL DEFAULT 0
+			MODIFY semi_sync_master_timeout BIGINT UNSIGNED NOT NULL DEFAULT 0
 	`,
 }
--- a/go/db/generate_patches.go
+++ b/go/db/generate_patches.go
@@ -576,4 +576,9 @@ var generateSQLPatches = []string{
 			database_instance
 			ADD COLUMN semi_sync_available TINYINT UNSIGNED NOT NULL DEFAULT 0 AFTER semi_sync_enforced
 	`,
+	`
+		ALTER TABLE
+			database_instance
+			MODIFY COLUMN semi_sync_master_timeout BIGINT UNSIGNED NOT NULL DEFAULT 0
+	`,
 }
--- a/go/inst/instance_dao.go
+++ b/go/inst/instance_dao.go
@@ -429,7 +429,7 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
 						instance.SemiSyncMasterEnabled = (m.GetString("Value") == "ON")
 						semiSyncMasterPluginLoaded = true
 					} else if m.GetString("Variable_name") == "rpl_semi_sync_master_timeout" {
-						instance.SemiSyncMasterTimeout = m.GetInt64("Value")
+						instance.SemiSyncMasterTimeout = m.GetUint64("Value")
 					} else if m.GetString("Variable_name") == "rpl_semi_sync_master_wait_for_slave_count" {
 						instance.SemiSyncMasterWaitForReplicaCount = m.GetUint("Value")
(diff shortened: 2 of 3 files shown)


go/db/generate_patches.go
```diff
@@ -576,4 +576,9 @@ var generateSQLPatches = []string{
 			database_instance
 			ADD COLUMN semi_sync_available TINYINT UNSIGNED NOT NULL DEFAULT 0 AFTER semi_sync_enforced
 	`,
+	`
+		ALTER TABLE
```
shlomi-noach:
```suggestion
		ALTER TABLE /* sqlite3-skip */
```
required for `sqlite3` backend

go/db/generate_patches.go
```diff
@@ -576,4 +576,9 @@ var generateSQLPatches = []string{
...
+			database_instance
+			MODIFY COLUMN semi_sync_master_timeout BIGINT UNSIGNED NOT NULL DEFAULT 0
```
shlomi-noach:
```suggestion
			MODIFY semi_sync_master_timeout BIGINT UNSIGNED NOT NULL DEFAULT 0
```
just to match convention in rest of the file

shlomi-noach (changes requested):
thank you, a couple changes requested

shlomi-noach:
(you can see all CI jobs failed -- that's because of the `sqlite3` missing hint)

shlomi-noach (commented):
we need to further change the matching `go` variable to `int64` or `uint64`, correct? Let's do this in this PR, too

go/inst/instance_dao.go
```diff
@@ -429,7 +429,7 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
 						instance.SemiSyncMasterEnabled = (m.GetString("Value") == "ON")
 						semiSyncMasterPluginLoaded = true
 					} else if m.GetString("Variable_name") == "rpl_semi_sync_master_timeout" {
-						instance.SemiSyncMasterTimeout = m.GetUint("Value")
+						instance.SemiSyncMasterTimeout = m.GetInt64("Value")
```
shlomi-noach: Don;t we need to support max UINT64?
```suggestion
						instance.SemiSyncMasterTimeout = m.GetUint64("Value")
```

go/inst/instance_dao.go
```diff
@@ -1139,7 +1139,7 @@ func readInstanceRow(m sqlutils.RowMap) *Instance {
 	instance.SemiSyncEnforced = m.GetBool("semi_sync_enforced")
 	instance.SemiSyncAvailable = m.GetBool("semi_sync_available")
 	instance.SemiSyncMasterEnabled = m.GetBool("semi_sync_master_enabled")
-	instance.SemiSyncMasterTimeout = m.GetUint("semi_sync_master_timeout")
+	instance.SemiSyncMasterTimeout = m.GetInt64("semi_sync_master_timeout")
```
shlomi-noach:
```suggestion
	instance.SemiSyncMasterTimeout = m.GetUint64("semi_sync_master_timeout")
```

shlomi-noach (changes requested):
I think last requested changes

go/inst/instance_dao.go
luisyonaldo: there is no `GetUint64` in https://github.com/openark/golib/blob/master/sqlutils/sqlutils.go

go/inst/instance_dao.go
luisyonaldo: there is no `GetUint64` in https://github.com/openark/golib/blob/master/sqlutils/sqlutils.go

luisyonaldo:
> I think last requested changes

I didn't do `uint64` because is not supported by `sqlutils` library, and all other cases that use `bigint unsigned` have `int64` as counterpart in go

shlomi-noach:
> there is no GetUint64

Whoops. But I think we must use it, right? Because the value reported is `2^64-1`, correct?

shlomi-noach:
I know why `sqlutils` dosn't have `GetUint64` -- that's because golang's [ParseInt](https://golang.org/pkg/strconv/#ParseInt) doen't support parsing a `uint64`.

shlomi-noach:
I wonder if we should literally first compare the string with `18446744073709551615`

luisyonaldo:
> I know why `sqlutils` dosn't have `GetUint64` -- that's because golang's [ParseInt](https://golang.org/pkg/strconv/#ParseInt) doen't support parsing a `uint64`.

it does, [ParseUint](https://golang.org/pkg/strconv/#ParseUint)

shlomi-noach:
> it does, ParseUint

!!!

The name confused me, I assumed it returns a `uint`, not `uint64`. So inconsistent

shlomi-noach:
See https://github.com/openark/orchestrator/pull/1209

luisyonaldo:
> See #1209

Looks good, would be good if we can make https://github.com/openark/orchestrator/pull/1104 work, so we can backport all this changes to their own repo.

Will wait for #1209 to be merged and then apply changes here

shlomi-noach:
`GetUint64` is now available

shlomi-noach:
> Looks good, would be good if we can make #1104 work

Yeah, I'm :100: :+1: on that. `sqlutils` has diverged a lot from the original codebase; iirc there's a bit of mess because I also made local changes to `sqlutils.go` in https://github.com/github/gh-ost

shlomi-noach (approved):

PR #1210, opened 2020-07-08
WIP: orchestrator as a library

shlomi-noach:
Looking into embedding `ochestrator` as a library within another Go process. Work in Progress.


shlomi-noach:
Not planning to pursue this in the neat future.

Issue #355, opened 2017-12-05
Expose Prometheus metrics


guidoiaquinti:
It would be good to integrate the [Prometheus client library](https://github.com/prometheus/client_golang) to expose Orchestrator metrics like we already do for [graphite](https://github.com/github/orchestrator/blob/master/go/metrics/graphite.go). 

I know that we expose the same metrics in expvar-like format via `/debug/metrics` but that's not exactly Prometheus compatible (see related [discussion](https://github.com/github/gh-ost/issues/104) in the `gh-ost` repo).

I'm opening this issue to keep track of this enhancement.

Thank you!

shlomi-noach:
@guidoiaquinti are you suggesting to add promotheus metrics? Otherwise how should this be done?

guidoiaquinti:
My suggestion is to add an HTTP route (e.g. `/metrics`) that can expose Orchestrator metrics in a Prometheus compatible format. 

Here's a [list](https://prometheus.io/docs/instrumenting/exporters/#software-exposing-prometheus-metrics) of some third-party software natively exposing metrics in the Prometheus format.

shlomi-noach:
@guidoiaquinti thanks :) I meant to ask whether you intend to implements us or want us to.

I'm not familiar with promotheus and this isn't on my immediate roadmap; given that I'll need to invest learning promotheus/metrics, I'm unlikely to do this in the near future.

guidoiaquinti:
Sorry for the misunderstanding @shlomi-noach 😄 

I've opened this issue only to keep track of the enhancement. I don't think I'll have time in the near future to implement it and it's currently not a high-pri for us.

shlomi-noach:
Cool. Thank you!

ameetkotian:
PR for review https://github.com/openark/orchestrator/pull/1414

PR #1211, opened 2020-07-11
Prometheus metrics

mhumpula:
### Description

This is a basic prometheus metrics exporter of instances metrics. The first commit adds a prometheus depdency libraries (would be nice if #1104 were merged first). The second commit exports instance status information so it's possible to create alerting rules in prometheus to alert on undesired states like not replicating slave. 

I think it would be possible to add existing orchestrator metrics from go-metrics module. 

Related issue: #355


shlomi-noach (commented):
Thank you for this! 
Question: are 1396 files really necessary? That's three times more than the current number of dependency files. I'm a bit afraid of uncontrolled dependency list. at some time we will switch to `go modules`, where I feel control is even weaker.

mhumpula:
> Thank you for this!
> Question: are 1396 files really necessary? That's three times more than the current number of dependency files. I'm a bit afraid of uncontrolled dependency list. at some time we will switch to `go modules`, where I feel control is even weaker.

I don't know... technically speaking I'm not a golang programmer, so I have no idea, what is the proper way to actually do this. I have copied one module at time to make the code compile successfully, so I think that this is the minimal set of modules. But it should be possible to edit the source code of prometheus module, so it doesn't need some dependencies (like protobuf). Not sure if it should be part of the vendor after that.

Issue #1214, opened 2020-07-22
Include Instance Alias in the output from the command line


martinarrieta:
Currently, when you display the topology from the command line, it only displays the Hostname and Port, however, it could be helpful (it is for us that we manage everything with IP's) to display as well the InstanceAlias if is set.

So instead of this:
```
10.0.0.2  [0s,ok,5.7.30-33-log,rw,MIXED,>>,P-GTID]
+ 10.0.0.3 [0s,ok,5.7.30-33-log,ro,MIXED,>>,P-GTID]
+ 10.0.0.4 [0s,ok,5.7.30-33-log,ro,MIXED,>>,P-GTID]
```

Display this:
```
10.0.0.8 (mysql1) [0s,ok,5.7.30-33-log,rw,MIXED,>>,P-GTID]
+ 10.0.0.3 (mysql2) [0s,ok,5.7.30-33-log,ro,MIXED,>>,P-GTID]
+ 10.0.0.4 (mysql3) [0s,ok,5.7.30-33-log,ro,MIXED,>>,P-GTID]
```

PR #1215, opened 2020-07-22
Added instance Alias to the topology display from the command line

martinarrieta:
Related issue: https://github.com/openark/orchestrator/issues/1214

### Description

This PR adds the instance Alias to the output of the topology command from the command line interface.


shlomi-noach (approved):
Makes sense. Thank you

Issue #1218, opened 2020-07-25
Packagecloud release schedule


thehunmonkgroup:
Curious what the process is for new GA releases to make it to the builds at https://packagecloud.io/github/orchestrator

I noticed the latest GA, v3.2.2, has been out for about a month, and the most recent version on Packagecloud is v3.1.4.

Is it on some kind of periodic build schedule, or did something break, or am I just not patient enough? ;)

Anything I can do to help get the latest release up there?

shlomi-noach:
Since I no longer work for GitHub, I don't have access to https://packagecloud.io/github/orchestrator and it's not going to be updated. I'm still looking into an alternative, whether I'll host in https://packagecloud.io/openark or what.

thehunmonkgroup:
Thanks for the clarification.

For anybody else in the meantime, it's pretty trivial to build your own packages using the very clear and handy https://github.com/openark/orchestrator/blob/master/build.sh

shlomi-noach:
> For anybody else in the meantime, it's pretty trivial to build your own packages

Even more trivial is via Docker. Run `script/dock pkg` to build and generate all packages without installing dependencies on your machine. See https://github.com/openark/orchestrator/blob/master/docs/docker.md#create-package-files

PR #1221, opened 2020-07-26
disable RPM build on missing rpmbuild binary

thehunmonkgroup:
The build script already warns the user that an RPM package will not be
built -- this includes the check for rpmbuild and disabling the build
if it's not found

Related issue: https://github.com/openark/orchestrator/issues/1218


shlomi-noach (approved):
Belated thank you!

PR #1246, opened 2020-09-29
Expire cluster_alias entries

### LINKED ISSUE #1245: Old clusters don't error out on web interface (/web/cluster/alias/<cluster>)
If I go to the webui of the cluster `foo` that never existed, I'll get the error message.
```
{"Code":"ERROR","Message":"No cluster found for alias foo","Details":null}
```

If I got to the webui of the cluster `bar` that existed in the past but no longer has any hosts, I'll get an empty page.

The logs do show:
```
[martini] Started GET /web/cluster/alias/bar for <ipaddress>
[martini] Completed 200 OK in 236.58µs
```

Then a couple lines later, I see. The `<fqdn>` matches the information in the `cluster_alias` table.
```
2020-09-28 11:50:32 ERROR Unable to determine cluster name. clusterHint=<fqdn>:3306
```

I was able to test this on several "old" clusters, by finding them via:
```
select * from cluster_alias where last_registered < DATE(DATE_SUB(NOW(), INTERVAL 1 DAY)) order by 3
```

My guess is the `forget` should also remove the `cluster_alias` information or the webui should behave differently, I'm not sure which.

shlomi-noach:
Fixes https://github.com/openark/orchestrator/issues/1245

This PR expires `cluster_alias` entries, based on `UnseenInstanceForgetHours`, same as `database_instance` entries are expired.

The effect is that `orchestrator` properly forgets aliases for clusters that have no instances (ie. created long ago, but which have probably been decomissioned, seeing that there's no instances that associate themselves with said alias).

cc @tomkrouper

### DIFF
--- a/go/inst/cluster_alias_dao.go
+++ b/go/inst/cluster_alias_dao.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/openark/golib/log"
 	"github.com/openark/golib/sqlutils"
+	"github.com/openark/orchestrator/go/config"
 	"github.com/openark/orchestrator/go/db"
 )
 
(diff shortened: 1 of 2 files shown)


tomkrouper (approved):
Thanks so much. Looks good.

Issue #1217, opened 2020-07-23
prevent multiple concurrent recoveries on a given cluster 


mzhou151:
Hi, we have an Orchestrator cluster with three nodes in raft setup run in kubernetes. We had an outage a few weeks ago that caused multiple mysql clusters becoming unresponsive at the same time, and Orchestrator initiated recoveries on all of them. For some unknown reason, Orchestrator wasn't able to complete these recoveries within 60min which is the default value of `RecoveryPeriodBlockSeconds`. So when the block window had passed, we saw new recoveries initiated on the same set of failed clusters. After some time, there were 3-4 recoveries running concurrently on a given cluster, and many of these recoveries did not select the same candidate to promote for the given cluster. At the end of the outage, we saw multiple clusters end up in a split brain state. We could extend `RecoveryPeriodBlockSeconds` value as a workaround, though that might put us in risk of not recovering a legitimate failure, so we're wondering if there's anything else built in Orchestrator that would prevent concurrent recoveries on a given cluster. Thanks

shlomi-noach:
> so we're wondering if there's anything else built in Orchestrator that would prevent concurrent recoveries on a given cluster

This can be improved. One of the ideas was that a recovery can unexpectedly terminate (`orchestrator` node itself going down or getting network isolated) and another `orchestrator` node picking up from there.

But, please tell me more. What happened during that 1 hour? Why was `orchestrator` unable to complete a failover within 1 hour? What _did_ `orchestrator` manage to do? Really need much more detail to understand better the situation.

mzhou151:
So we weren't sure what caused Orchestrator to be slow on recovering at the time, Orchestrator was logging intensively during the outage, mostly related to polling and analysis, so it was very active, but from the recovery logs, we saw long gaps between each recovery step. Below are the first few lines from one recovery, even though it only took 2.69ms to execute the hook, but the whole process took 23min. 
```
2020-06-20T18:03:27Z	Running 1 PreFailoverProcesses hooks
2020-06-20T18:15:14Z	Running PreFailoverProcesses hook 1 of 1: echo 'Will recover from DeadMasterAndSomeSlaves on host1:1234' >> /tmp/recovery.log
2020-06-20T18:26:22Z	Completed PreFailoverProcesses hook 1 of 1 in 2.690399ms
```
Within that 23min, we saw mostly messages from functions such as `executeCheckAndRecoverFunction`, `AttemptRecoveryRegistration`, `CheckAndRecover` .

[...]

PR #1249, opened 2020-10-11
check for active recovery at registration and clearance

mzhou151:
Related issue: #1217

This PR adds additional checks when attempting to register a new recovery and when clearing an active recovery. Specifically, at `AttemptRecoveryRegistration` it checks for any ongoing active recovery on the given cluster via `ReadActiveClusterRecovery`, and at `ClearActiveRecoveries` it adds one more condition that checks for `end_recovery` in the sql query, so that only resolved recovery will be cleared. These two added checks are to ensure only one active recovery is occurring at a time for a given cluster. 

Thanks


shlomi-noach:
One concern I have over this change is: what happens if some recovery hangs? For whatever reason, perhaps it calls a pre-or post- failover hook that takes ages. What ensures that this recovery is eventually cleared up?

mzhou151:
@shlomi-noach the motive behind this PR is in fact related to a incident where a recovery was hung for a long time (see related issue #1217), it eventually finished after many hours, but in the meantime other recoveries kicked off on the same cluster when `RecoveryPeriodBlockSeconds` reaches, so we ended up having multiple recoveries occurring on a given cluster.  I agree we can add more logic to clean up a long possibly hanging recovery in future, but this PR is only intended to prevent a new recovery from kicking off if previous recovery wasn't finished, ensuring only one recovery taking place at any given time on a given cluster. Thanks

shlomi-noach:
> but this PR is only intended to prevent a new recovery from kicking off if previous recovery wasn't finished, ensuring only one recovery taking place at any given time on a given cluster. 

Thank you. I understand that, and my concern remains. If, for whatever reason, that recovery hangs or crashes, then it will never be marked as complete. Following that, there will _never_ be a new recovery. I feel uncomfortable with this "never"-style scenario. Perhaps it would make sense for you to just increase `RecoveryPeriodBlockSeconds` to a value safe enough for you (say, 12 hours) that a human will always be able to look into, and acknowledge, either a failed or a successful migration?

mzhou151:
> > but this PR is only intended to prevent a new recovery from kicking off if previous recovery wasn't finished, ensuring only one recovery taking place at any given time on a given cluster.
> 
> Thank you. I understand that, and my concern remains. If, for whatever reason, that recovery hangs or crashes, then it will never be marked as complete. Following that, there will _never_ be a new recovery. I feel uncomfortable with this "never"-style scenario. Perhaps it would make sense for you to just increase `RecoveryPeriodBlockSeconds` to a value safe enough for you (say, 12 hours) that a human will always be able to look into, and acknowledge, either a failed or a successful migration?

I think I can understand your concern, on the other hand, if a recovery hangs a second recovery could step over it, and if the second recovery picks a different candidate replica from what the previous recovery picked we might end up in a split brain. We thought about setting `RecoveryPeriodBlockSeconds` to a large number but also had a concern that a legitimate failure would be left unattended for a long time. Do you think if Orchestrator could clear up previous recovery before kicking off a new one after `RecoveryPeriodBlockSeconds` expires, would that be more desirable? Thanks

shlomi-noach:
> Do you think if Orchestrator could clear up previous recovery before kicking off a new one after RecoveryPeriodBlockSeconds expires,

"clear up" as in abort it? It would be nice to have but difficult to implement, I think, because:

- What if a pre/post failover hook is running?
- Can you ensure a query gets killed? Rather than gets blocked?
- Same for TCP

But the idea is appealing to me. Let me think of it some more. If that's the direction, then the change is not small.

Issue #1253, opened 2020-10-16
Orchestrator cannot recover async/semi-sync replicas of failed replication group members


ejortegau:
Some basic support for group replication was previously added in https://github.com/openark/orchestrator/pull/1180
However, it is only about understanding single primary groups and being able to display them in the UI, but it does not tackle
anything related to failure detection and recovery.

While group replication has built-in primary election and fail-over, this is only for the group itself. If any group member (be it the primary or any secondary) has traditional async/semi-sync replicas, and it fails, neither group replication nor orchestrator currently fix that.

Hence, this Issue is a feature request so that Orchestrator can fix traditional replicas of group members.

PR #1254, opened 2020-10-16
Add support for recovery of async/semisync replicas of failed replication group  members

ejortegau:
Related issue:  https://github.com/openark/orchestrator/issues/1253

### Description

This PR addresses the issue mentioned above. It does so by adding failure detection and recovery for replication group members that have traditional async/semi-sync replicas.

cc @sjmudd, @dveeden, @luisyonaldo.

### DIFF
--- a/go/inst/analysis.go
+++ b/go/inst/analysis.go
@@ -57,6 +57,8 @@ const (
 	AllIntermediateMasterReplicasNotReplicating                          = "AllIntermediateMasterReplicasNotReplicating"
 	FirstTierReplicaFailingToConnectToMaster                             = "FirstTierReplicaFailingToConnectToMaster"
 	BinlogServerFailingToConnectToMaster                                 = "BinlogServerFailingToConnectToMaster"
+	// Group replication problems
+	DeadReplicationGroupMemberWithReplicas = "DeadReplicationGroupMemberWithReplicas"
 )
 
 const (
@@ -110,6 +112,7 @@ const (
 	AnalysisInstanceTypeMaster             AnalysisInstanceType = "master"
 	AnalysisInstanceTypeCoMaster           AnalysisInstanceType = "co-master"
 	AnalysisInstanceTypeIntermediateMaster AnalysisInstanceType = "intermediate-master"
+	AnalysisInstanceTypeGroupMember        AnalysisInstanceType = "group-member"
 )
 
 // ReplicationAnalysis notes analysis on replication chain status, per instance
@@ -122,6 +125,7 @@ type ReplicationAnalysis struct {
 	AnalyzedInstancePhysicalEnvironment       string
 	AnalyzedInstanceBinlogCoordinates         BinlogCoordinates
 	IsMaster                                  bool
+	IsReplicationGroupMember                  bool
 	IsCoMaster                                bool
 	LastCheckValid                            bool
 	LastCheckPartialSuccess                   bool
@@ -212,6 +216,9 @@ func (this *ReplicationAnalysis) AnalysisString() string {
 func (this *ReplicationAnalysis) GetAnalysisInstanceType() AnalysisInstanceType {
 	if this.IsCoMaster {
 		return AnalysisInstanceTypeCoMaster
+	}
+	if this.IsReplicationGroupMember {
+		return AnalysisInstanceTypeGroupMember
 	}
 	if this.IsMaster {
 		return AnalysisInstanceTypeMaster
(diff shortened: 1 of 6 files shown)


go/inst/analysis.go
```diff
@@ -57,6 +57,8 @@ const (
 	AllIntermediateMasterReplicasNotReplicating                          = "AllIntermediateMasterReplicasNotReplicating"
 	FirstTierReplicaFailingToConnectToMaster                             = "FirstTierReplicaFailingToConnectToMaster"
 	BinlogServerFailingToConnectToMaster                                 = "BinlogServerFailingToConnectToMaster"
+	// Group replication problems
+	DeadReplicationGroupMemberWitReplicas = "DeadReplicationGroupMemberWithReplicas"
```
shlomi-noach:
```suggestion
	DeadReplicationGroupMemberWithReplicas = "DeadReplicationGroupMemberWithReplicas"
```

go/inst/instance_dao.go
```diff
@@ -3184,6 +3186,8 @@ func PopulateGroupReplicationInformation(instance *Instance, db *sql.DB) error {
 		@@global.group_replication_single_primary_mode
 	FROM
 		performance_schema.replication_group_members
+	WHERE
+		MEMBER_STATE <> 'OFFLINE'
```
shlomi-noach: just for consistency 
```suggestion
		MEMBER_STATE != 'OFFLINE'
```

go/logic/topology_recovery.go
```diff
@@ -1411,6 +1455,46 @@ func checkAndRecoverGenericProblem(analysisEntry inst.ReplicationAnalysis, candi
...
 
+// checkAndRecoverDeadGroupMemberWithReplicas checks whether action needs to be taken for an analysis involving a dead
+// replication group member, and takes the action if applicable. Notice that under our view of the world, a primary
+// replication group member is akin to a master in traditional async/semisync replication; whereas secondary group
+// members are akin to intermediate masters. Considering also that a failed group member can always be considered as a
+// secondary (even if it was primary, the group should have detected its failure and elected a new primary), then
+// failure of a group member with replicas is akin to failure of an intermediate master.
+func checkAndRecoverDeadGroupMemberWithReplicas(analysisEntry inst.ReplicationAnalysis, candidateInstanceKey *inst.InstanceKey, forceInstanceRecovery bool, skipProcesses bool) (bool, *TopologyRecovery, error) {
+	// Don't proceed with recovery unless it was forced or automatic intermediate source recovery is enabled.
+	// We consider failed group members akin to failed intermediate masters, so we re-use the configuration for
```
shlomi-noach, line 1466: > so we re-use the configuration 

but in `analysis_dao.go` it seems like you've changed that: intermediate master recovery only takes place under
```golang
if !a.IsReplicationGroupMember {
```

go/logic/topology_recovery.go
```diff
@@ -1411,6 +1455,46 @@ func checkAndRecoverGenericProblem(analysisEntry inst.ReplicationAnalysis, candi
...
+	if recoveredToGroupMember != nil {
+		// success
+		recoverDeadReplicationGroupMemberSuccessCounter.Inc(1)
+
+		if !skipProcesses {
+			// Execute post failover processes
+			topologyRecovery.SuccessorKey = &recoveredToGroupMember.Key
+			topologyRecovery.SuccessorAlias = recoveredToGroupMember.InstanceAlias
+			// For the same reasons that were mentioned above, we re-use the post intermediate master fail-over hooks
+			executeProcesses(config.Config.PostIntermediateMasterFailoverProcesses, "PostIntermediateMasterFailoverProcesses", topologyRecovery, false)
```
shlomi-noach, line 1490: I don't run Group Replication myself, but I think it can be debatable whether it is correct to run `PostIntermediateMasterFailoverProcesses`. For now, let's keep it at that, but I predict that someone in the future will argue against this.

shlomi-noach (commented):
please see inline comments

go/logic/topology_recovery.go
ejortegau: What  I mean here is that we re-use  `analysisEntry.ClusterDetails.HasAutomatedIntermediateMasterRecovery` configuration to decide whether we want to fail-over group members as opposed to having a separate configuration. As mentioned in the method's doc comment, we are operating under the assumption that group secondaries with replicas are akin to intermediate masters in the sense that they perform a very similar function in the replication chain; get and apply changes from the primary (except, via GR instead of binlog), and distribute them to replicas (via the binlog). I hope this clarifies my intent.

go/logic/topology_recovery.go
ejortegau: For now, our use case does not seem to require different hooks for these. If the need arises (or someone comes knocking at your door asking for it) I'd be happy to change this to have different GR and intermediate source hooks.

shlomi-noach (approved):

PR #1265, opened 2020-11-16
writeManyInstances: improve error message if we run out of placeholders.

sjmudd:
### Description

If orchestrator uses prepared statements and uses the bulk insert options it can trigger a large bulk insert statement to be triggered which exceeds the 64k placeholders that are the limit in the MySQL server.

Related issue: 
* https://bugs.mysql.com/bug.php?id=101630

I have bumped into this on several occasions and the client side workarounds while not good have to be taken.  This can also adversely affect the performance of keeping a cluster up to date.

To help the orchestrator administrator understand how to resolve this I have added code to recognise the specific error and to provide some more details of why the error happened, suggesting a configuration change.

Something like:
```
2020-11-16T15:11:37.295707+01:00 orchestrator orchestrator[57507]: flushInstanceWriteBuffer last_seen: writeManyInstances(?,true,true): error: Error 1390: Prepared statement contains too many placeholders, len(instances): 999
, len(args): 99999.  Reduce InstanceWriteBufferSize to avoid len(args) being > 64k, a limit in the MySQL source code.
```

Reducing the size of `InstanceWriteBufferSize` or alternatively setting: ` "MySQLInterpolateParams": true,` would work.

### DIFF
--- a/go/inst/instance_dao.go
+++ b/go/inst/instance_dao.go
@@ -2639,6 +2639,14 @@ func mkInsertOdkuForInstances(instances []*Instance, instanceWasActuallyFound bo
 }
 
 // writeManyInstances stores instances in the orchestrator backend
+//
+// This may trigger this error: Error 1390: Prepared statement contains too
+// many placeholders if you are monitoring several thousand servers as it
+// may trigger the total number of place holders to exceed (uint)UINT_MAX16
+// in the MySQL source code.
+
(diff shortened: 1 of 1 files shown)


sjmudd:
I see a system test failing but don't see/understand what's actually broken.

shlomi-noach:
> I see a system test failing but don't see/understand what's actually broken.

Seems like a flaky test which I should fix.

shlomi-noach (approved):

Issue #1273, opened 2020-11-19
Race-condition in Consul KV updates during failover


timvaillancourt:
👋 recently we encountered a key-value-update race condition under [Orchestrator release v3.1.4](https://github.com/openark/orchestrator/releases/tag/v3.1.4) using a setup that stores key-values for the current MySQL master in Consul

Our setup has Orchestrator deployed in Raft mode with KV updates sent to Consul on topology changes. Our MySQL load balancer reads from the same Consul server(s) and updates HaProxy backend configs (via [github.com/hashicorp/consul-template](https://github.com/hashicorp/consul-template)) when KVs are changed in Consul, causing a restart of HaProxy - which is used to direct our MySQL traffic. This setup is outlined by @shlomi-noach in this blog post: https://github.blog/2018-06-20-mysql-high-availability-at-github/

Currently [`go/kv/consul.go`](https://github.com/openark/orchestrator/blob/master/go/kv/consul.go) uses 4 x separate Consul "put" calls to update 4 x different keys in Consul **without** using [an atomic Consul transaction](https://www.consul.io/api-docs/txn)

In the problem scenario:
1. A cluster under Orchestrator was issued a `graceful-master-takeover` to a new Primary
2. Orchestrator updated some **but not all** of the 4 x Consul KVs
3. Consul-template on our load balancer sees a change with the partially-updated KVs, triggering a restart of HaProxy
   1. All HaProxy process saw a Consul KV change and restarted during the KV changes, however not all got a consistent config
4. HaProxy ran a config built from the partially-updated Consul KVs, causing some queries to route to the demoted Primary
5. Orchestrator completed updating all 4 x Consul keys
    1. For whatever reason, consul-template didn't trigger subsequent reloads - a theory is it was "busy" when the final updates came in

The timing of this scenario is under 1-3 seconds. We have many HaProxy instances in each site and this race-condition happens on only a portion of the proxies, somewhat intermittently, perhaps 1 in every 10-20 failovers. This situation is remedied by manually triggering a reload of the Load Balancer once the Consul KVs are consistently updated by Orchestrator

I would like to propose the solution of adding a new, optional [`KVStore` interface](https://github.com/openark/orchestrator/blob/master/go/kv/kv.go#L24-L27) that uses [a Consul transaction to update the KVs atomically in a single operation](https://www.consul.io/api-docs/txn). The reason for this is this removes the race condition in the update of Consul KVs. It also is more efficient to perform a single HTTP call to Consul

[...]

PR #1274, opened 2020-11-23
Add .PutKVPairs() method to KVStore interface

timvaillancourt:
Related issue: https://github.com/openark/orchestrator/issues/1273

### Description

This PR adds the `.PutKVPairs()` function to the `KVStore` interface _(from `go/kv/kv.go`)_ to allow decisions around updating multiple KVs to be pushed down to the various KV store implementations. This new method replaces the single-KVPair method `.PutKVPair()` from `go/kv/kv.go`

The following files were updated to use `.PutKVPairs()` vs single-KVPair updates:
1. `go/logic/command_applier.go`
2. `go/logic/orchestrator.go`
3. `go/logic/topology_recovery.go`

For now the `.PutKVPairs()` function in each `KVStore` is just doing a plain-loop over `.PutKeyValue()`. In the near-future this method will be used to implement atomic multi-KVPair updates for the Consul KV store (https://github.com/openark/orchestrator/issues/1273) and potentially the Internal/ZooKeeper store, too

cc @shlomi-noach

### DIFF
--- a/go/logic/orchestrator.go
+++ b/go/logic/orchestrator.go
@@ -444,7 +444,7 @@ func SubmitMastersToKvStores(clusterName string, force bool) (kvPairs [](*kv.KVP
 	log.Debugf("kv.SubmitMastersToKvStores: submitKvPairs: %+v", len(submitKvPairs))
 	if orcraft.IsRaftEnabled() {
 		for _, kvPair := range submitKvPairs {
-			_, err = orcraft.PublishCommand("put-key-value", kvPair)
+			_, err := orcraft.PublishCommand("put-key-value", kvPair)
 			if err == nil {
 				submittedCount++
 			} else {
--- a/go/logic/orchestrator.go
+++ b/go/logic/orchestrator.go
@@ -442,23 +442,19 @@ func SubmitMastersToKvStores(clusterName string, force bool) (kvPairs [](*kv.KVP
 		submitKvPairs = append(submitKvPairs, kvPair)
 	}
 	log.Debugf("kv.SubmitMastersToKvStores: submitKvPairs: %+v", len(submitKvPairs))
-	var kvPutKVPairs [](*kv.KVPair)
-	for _, kvPair := range submitKvPairs {
-		if orcraft.IsRaftEnabled() {
+	if orcraft.IsRaftEnabled() {
+		for _, kvPair := range submitKvPairs {
 			_, err = orcraft.PublishCommand("put-key-value", kvPair)
 			if err == nil {
 				submittedCount++
 			} else {
 				selectedError = err
 			}
-		} else {
-			kvPutKVPairs = append(kvPutKVPairs, kvPair)
 		}
-	}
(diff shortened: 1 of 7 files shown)


go/logic/orchestrator.go
```diff
@@ -442,14 +442,23 @@ func SubmitMastersToKvStores(clusterName string, force bool) (kvPairs [](*kv.KVP
...
 		} else {
-			err = kv.PutKVPair(kvPair)
+			kvPutKVPairs = append(kvPutKVPairs, kvPair)
 		}
+	}
+	if len(kvPutKVPairs) > 0 {
+		err := kv.PutKVPairs(kvPutKVPairs)
 		if err == nil {
-			submittedCount++
+			submittedCount = submittedCount + len(kvPutKVPairs)
```
shlomi-noach: I'm not sure I follow the logic here: iterate kvpairs, and if we're on in raft-mode, then it's a no-op, we generate and append to a new slice, which is then tested and applied?

How about this alternative pseudo code:
```
if orcraft.IsRaftEnabled() {
  for _, kvPair := range submitKvPairs {
    _, err = orcraft.PublishCommand("put-key-value", kvPair)
    ...
  }
} else {
  err := kv.submitKvPairs(kvPutKVPairs)
  ...
}
```
this way we don't need the extra slice `kvPutKVPairs` and the logic is simplified. What do you think?

shlomi-noach (changes requested):
The addition of `PutKVPairs()` is good and makes sense. See inline a comment on how `PutKVPairs()` is used within `orchestrator.go`

go/logic/orchestrator.go
timvaillancourt: @shlomi-noach thanks for reviewing!

Yes, on second look this makes no sense 😄 - for some reason I thought I needed to filter each KV against `orcraft.IsRaftEnabled()` but in hindsight that setting is all or nothing

I'll update with the simpler pseudo code you proposed 👍

timvaillancourt:
Updated PR based on @shlomi-noach's suggestions

go/logic/orchestrator.go
```diff
@@ -442,14 +442,19 @@ func SubmitMastersToKvStores(clusterName string, force bool) (kvPairs [](*kv.KVP
 		submitKvPairs = append(submitKvPairs, kvPair)
 	}
 	log.Debugf("kv.SubmitMastersToKvStores: submitKvPairs: %+v", len(submitKvPairs))
-	for _, kvPair := range submitKvPairs {
-		if orcraft.IsRaftEnabled() {
+	if orcraft.IsRaftEnabled() {
+		for _, kvPair := range submitKvPairs {
 			_, err = orcraft.PublishCommand("put-key-value", kvPair)
```
shlomi-noach:
```suggestion
			_, err := orcraft.PublishCommand("put-key-value", kvPair)
```

shlomi-noach (approved):
Looks good with minor `suggested change`.

Issue #1275, opened 2020-11-27
Recommendations around fencing / STONITH


binwiederhier:
We are experimenting with Orchestrator and the GitHub-style failover strategies (HAproxy + consul + consul-template) and so far it is working great. Thank you for making this great piece of software, and for describing your failover strategies in such great detail.

I've read various posts and questions around how to fence off a failed master/source host when Orchestrator has begun a failover/recovery, but it's not clear if there is a **recommendation** from you of how to do it, so if time permits, @shlomi-noach, I'd like to hear your thoughts on best practices here.

**What I've implemented currently in my dev setup:** On the host that runs HAproxy, I have a tiny python script that queries Orchestrator's `/api/all-instances` every 5 seconds and caches the results. I added an `external-check` into HAproxy that reads that file and fails if the host is marked `downtimed`. This works, but I believe is not ideally fencing the old source off, because I believe the `IsDowntimed` flag is not set early enough. 

**What I think would be ideal:** If Orchestrator could update the KV store the second it detects the `DeadMaster` event (or others) with an empty value, or something else indicating that this host has been marked dead, we could update the consul-template to produce a dead-end-type `server` entry in HAproxy. Sure this can likely be implemented via `PreFailoverProcesses` but I'd feel much better if this was an Orchestrator feature (potentially opt-in). Heck if you think it's a good idea I'd even be willing to implement it, will likely not be that hard, right?

Again, thanks for your time and this great project!

jrsmiley:
STONITH has many known issues with reliability and is generally not a recommended approach to ensuring that a member leaves a group.  The best practice is to have a lease protocol where a node that cannot communicate with the lease arbiter is to SMITH.

shlomi-noach:
> how to fence off a failed master/source host when Orchestrator has begun a failover/recovery,

anything based on `orchestrator`'s hooks, or on a proxy detecting the change etc. will only _reduce_ the time the failed primary can receive writes. That is, fencing will take place _sooner rather than later_, but is it _soon enough_? What is soon enough? If you're good with letting the old primary receive writes for 1sec or for 100ms, you're still letting it deviate from the newly promoted server.
So first thing is to understand what you're trying to accomplish.

[...]

PR #1278, opened 2020-12-10
Rudimentary version of "KVWipeValuesPreFailover" option

binwiederhier:
Related issue: https://github.com/openark/orchestrator/issues/1275

### Description

This is an entirely untested proposal to introduce an option called `KVWipeValuesPreFailover`. If enabled, the values in the KV store will be zeroed out at the beginning of the failover process, i.e. `mysql/master/$cluster/{hostname,port,ipv4,ipv6}` will all be set to `""` (empty string). The purpose of this is to allow immediate fencing of the mysql host that has failed, e.g. by reacting to it in consul-template ("if empty then remove from writer hostgroup" or "if empty then repoint to dead end").

The same can currently be achieved by running `consul kv put ...` in the `PreFailover...` hooks, but having it in Orchestrator seemed like a pretty good idea to me.

If this proposal is well-received, I'm happy to make any adjustments to the code that are necessary to make this a real thing.

### DIFF
--- a/go/config/config.go
+++ b/go/config/config.go
@@ -265,8 +265,9 @@ type Configuration struct {
 	ConsulScheme                               string            // Scheme (http or https) for Consul
 	ConsulAclToken                             string            // ACL token used to write to Consul KV
 	ConsulCrossDataCenterDistribution          bool              // should orchestrator automatically auto-deduce all consul DCs and write KVs in all DCs
-	ZkAddress                                  string            // UNSUPPERTED YET. Address where (single or multiple) ZooKeeper servers are found, in `srv1[:port1][,srv2[:port2]...]` format. Default port is 2181. Example: srv-a,srv-b:12181,srv-c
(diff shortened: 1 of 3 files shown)


shlomi-noach:
Hi! Thank you for this contribution. Could you explain again about the use case? How will intermediate fencing improve the failover process, or rather what is the specific problem this solves?

binwiederhier:
Thank you for your response. I understand your time is valuable so I'll try to keep it brief. 

As briefly outlined in https://github.com/openark/orchestrator/issues/1275, I'd like to fence off a failing host as early as possible, so that clients won't erroneously continue writing to it when orchestrator has already decided to fail over to another host, e.g. in the event of temporary network blips, or a misconfigured MySQL systemd service that restarts MySQL when it crashes.

A concrete example that happened in my lab environment with a Orchestrator/Consul/Consul Template/HAproxy setup: 

1. killall mysqld on the source host
2. Orchestrator decides to fail over
3. systemd decides to restart MySQL on the source host
4. Clients continue writing to original source
...

You get the idea. And while this particular thing was solved by `Restart=no` in the systemd service, network blips may not be solved by that. 

I just want to make sure that if orchestrator decides a host is bad, that clients immediately are stopped from communicating with that host.

PR #1284, opened 2020-12-29
improve some language in failure-detection.md

seeekr:
Continued to read some docs, found some more things that could be worded better :)

By the way, I assume that orchestrator is still fine with handling MySQL 8.0 standard async replication? There doesn't seem to be any info in the docs and around the web about this, which to me would mean that since 8.0 is acting the same as e.g. 5.7, orchestrator doesn't care and works as expected?

### DIFF
--- a/docs/failure-detection.md
+++ b/docs/failure-detection.md
@@ -2,14 +2,14 @@
 
 `orchestrator` uses a holistic approach to detect master and intermediate master failures.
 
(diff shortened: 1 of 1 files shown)


shlomi-noach:
Orchestrator should be fine with 8.0 async replication.

shlomi-noach (approved):
Thank you

PR #1295, opened 2021-02-02
Drop fixed list of cipher suites.

### LINKED ISSUE #1294: Hard-coded ciphersuite list incompatible with mariadb 10.1 on debian stretch
Currently there's a hard-coded list of accepted ciphersuites: https://github.com/openark/orchestrator/blob/df4187371f48273ee02c5acd439e2501e8be7057/go/ssl/ssl.go#L19-L30

This list hasn't been updated in 5 years. It also happens to be incompatible with Mariadb 10.1 on debian stretch:

`2021-02-02 09:23:30 ERROR ReadTopologyInstance(HOSTNAME:3306) show global status like 'Uptime': remote error: tls: handshake failure back`.

kormat:
The list of cipher suites hasn't been updated in 5 years, and the go
docs since go1.12 say that it defaults to secure ciphers.

Fixes https://github.com/openark/orchestrator/issues/1294

### DIFF
--- a/go/ssl/ssl.go
+++ b/go/ssl/ssl.go
@@ -16,19 +16,6 @@ import (
 	"github.com/openark/orchestrator/go/config"
 )
 
-var cipherSuites = []uint16{
-	tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
-	tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
-	tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
-	tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
-	tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
-	tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,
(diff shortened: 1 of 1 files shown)


marostegui:
This patch fixed the issue in our environment.

go/ssl/ssl.go
```diff
@@ -46,8 +33,8 @@ func NewTLSConfig(caFile string, verifyCert bool) (*tls.Config, error) {
 
 	// Set to TLS 1.2 as a minimum.  This is overridden for mysql communication
 	c.MinVersion = tls.VersionTLS12
-	// Remove insecure ciphers from the list
-	c.CipherSuites = cipherSuites
+	// "If CipherSuites is nil, a default list of secure cipher suites is used"
```
shlomi-noach, line 36: What is the default list of secure cipher suites?

shlomi-noach (commented):
Sorry for the late review. Please see inline question.

go/ssl/ssl.go
kormat: It's roughly https://golang.org/src/crypto/tls/cipher_suites.go?s=1340:1374#L40
(The ordering can vary depending on what hardware features are available)

shlomi-noach: Thank you

shlomi-noach (approved):

PR #1319, opened 2021-03-04
If access to ORCHESTRATOR_API fails do not expose the password(s)

sjmudd:
### Description

Recently seen:

```
orchestrator-client[10450]: Cannot access orchestrator at http://user:password@host.com/api.  Check ORCHESTRATOR_API is configured correctly and orchestrator is running
```

This exposes the password.

Change to:

```
orchestrator-client[10450]: Cannot access orchestrator at http://user:<REMOVED>@host.com/api.  Check ORCHESTRATOR_API is configured correctly and orchestrator is running
```
Also works if `ORCHESTRATOR_API` contains multiple hostnames.


shlomi-noach (approved):
Thank you

Issue #1320, opened 2021-03-04
Docker: run orchestrator as non root user 


garutilorenzo:
Run orchestrator as non root user for security reasons (Docker best practices) I will submit a pull request with the necessary changes.
In short: 

* create a user named "docker" UID 1000
* create a group named "docker" GID 1000
* run chown on /app direcotry and /etc/orchestrator.conf.json file

PR #1321, opened 2021-03-04
Docker: run orchestrator as non root user

garutilorenzo:
Related issue: https://github.com/openark/orchestrator/issues/1320

### Description

Run orchestrator as non root user for security reasons (Docker best practices).
On local environment orchestrator starts up normally and the orchestrator process is run by "docker" user.

Changes apply only for docker/Dockerfile


docker/Dockerfile
```diff
@@ -41,3 +41,22 @@ COPY --from=build /etc/orchestrator.conf.json /etc/orchestrator.conf.json
...
+    --home /usr/local/orchestrator \
+    --ingroup "$USER" \
+    --no-create-home \
+    --uid "$UID" \
+    "$USER"
+
+RUN mkdir -p /app/data && chown -R docker:docker /app \
+    && chown -R docker:docker /etc/orchestrator.conf.json
+
+USER docker
```
till: If you move this up (including adduser) you can avoid the chown later. Before copying anything into the container.

garutilorenzo: If i move this lines before copying files into the container, when the container tries to start i get the following error: FATAL unable to open database file: no such file or directory

PR #1330, opened 2021-03-18
Introducing RecoverReadOnlyMaster option, ReadOnlyMaster failure analysis and recovery

### LINKED ISSUE #1328: Orchestrator to monitor RW status of the primary node.
Hi,

Our config file has super_read_only set by default so whenever a mysql instance is started, it will be in super_read_only mode . 
On the initial configuration of a cluster to be monitored by Orchestrator, the primary node is set to RW and subsequent failovers/takeovers are handled by Orchestrator which will set the newly promoted primary to RW. 
There are cases when mysql instance is restarted so fast that Orchestrator will not detect it fast enough and will not do a failover. One such example is when memory manager picks mysql instance as the biggest memory offender and kills it. mysqld_safe will restart it quickly and Orchestrator does not get to do a failover. In this case, former primary which continues to be primary, will be started with RO and we need to have an outside process to monitor it and switch it to RW. We have tested with  MASTER_CONNECT_RETRY lowered to 1 second to make sure that standbys would detect the change of status fast enough but even with this value, Orchestrator did not trigger the failover in our tests.
Since Orchestrator has the topology info, would it be possible to implement a feature that monitors the status of the primary and switches it back to RW if it becomes RO? 

Thank you,
 Liviu

shlomi-noach:
Fixes https://github.com/openark/orchestrator/issues/1328

### Description

This PR introduces an opt-in behavior: when a replication cluster's primary is found to be `read_only`, `orchestrator` analyzes that as a failure scenario and, based on opt-in configuration, spawn a recovery.

This behavior is controlled by the new variable `RecoverReadOnlyMaster` (defaults `false`)


shlomi-noach:
I wonder if we should just use `NoWriteableMasterStructureWarning` as the failure indicator.

shlomi-noach:
Closed in favor of https://github.com/openark/orchestrator/pull/1332

Issue #1343, opened 2021-04-28
Raft Reverse-Proxy doesn't work when UseSSL = true


akatashev:
This issue is similar to https://github.com/openark/orchestrator/issues/873

### Actual behaviour:
When `orchestrator.conf.json` is configured that way:
```json
{
  "HTTPAdvertise": "http://{{ orchestrator_hostname }}:{{ orchestrator_port }}",
  "UseSSL": false
}
```
Then reverse-proxy works fine and `curl - D - http://{orchestrator_ip}:{orchestrator_port}/api/clusters` request returns correct information about known clusters both for a raft leader and its followers.

But when it is configured that way:
```json
{
  "HTTPAdvertise": "https://{{ orchestrator_hostname }}:{{ orchestrator_port }}",
  "UseSSL": true
}
```
Then `curl -D - -k https://{orchestrator_ip}:{orchestrator_port}/api/clusters` request returns correct information only if it was sent to a raft leader. If it's sent to a follower, it returns `502 Bad Gateway` and Orchestrator logs say: `remote error: tls: bad certificate`.

### Expected behaviour:

`reverse-proxy` should work properly, so any follower should be able to provide requested information from its leader.

### Root cause analysis:

https://github.com/openark/orchestrator/blob/f818d8a135667574a14b4a3cb33dacb54f330dab/go/http/raft_reverse_proxy.go#L46-L47
Creates a new proxy without any HTTP transport and starts serving HTTP . Since no transport is configured, a default HTTP transport is created automatically:
```go
   func (p *ReverseProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
   	transport := p.Transport
    	if transport == nil {
    		transport = http.DefaultTransport
    	}
```
This works when SSL is disabled, but it breaks routing logic in the middle when it is enabled.

### Workaround:
Sending all requests directly to a raft leader works pretty well both with and without SSL.

### Solution proposal:
A suitable HTTP transport is already being created for raft health requests:
https://github.com/openark/orchestrator/blob/f818d8a135667574a14b4a3cb33dacb54f330dab/go/raft/http_client.go#L37-L68
This code can be reused to get a proper HTTP transport for a reverse-proxy as well.

shlomi-noach:
also possibly related to #979?

akatashev:
> also possibly related to #979?

I don't really think so. It is about authentication and this issue is about wrong `TLSClientConfig` for Proxy transport.
[...]

PR #1344, opened 2021-04-28
Use RaftHttpTransport for reverse-proxy

akatashev:
Related issue: https://github.com/openark/orchestrator/issues/1343

### Description

This PR introduces a `GetRaftHttpTransport` function in `orcraft` package that returns an `http.Transport` object that can be used for both raft Health Checks and raft reverse-proxy.
This function is used to set a correct HttpTransport for a raft reverse-proxy to make it SSL compatible.


shlomi-noach (approved):
LGTM, thank you!

shlomi-noach:
Fixes #1343

PR #1346, opened 2021-04-29
Update topology-recovery.md for typo

michaelcoburn:
Fixing it because otherwise it breaks copy-paste and we sure don't want that! :)


shlomi-noach (approved):
thank you!

Issue #979, opened 2019-09-11
security function basic extended  readonly user  automately switch to dba user(write privilege)


402309718:
when I try to test auth fuction I found  basic extended  with AuthenticationMethod=multi,I use readonly user login to non-leader  node,but it switch the writable user ,I login to th leader node ,it shows readonly 

about redirect this part whether exists some problem,and if possiable  it better add logout icon and add some group ,role

shlomi-noach:
Hi,
I'm not sure I understand the exact problem, actually. And what setup are you using? If that orchestrator/raft? (in which case most traffic does indeed get auto-routed to the leader?).

> if possiable it better add logout icon and add some group ,role

If you want groups/roles then you should not use `Basic` or `Multi` authentication method. That's not what they're for.

Right now `orchestrator` does not support `ldap`; the common solution is to put it behind `apache` or `nginx` or whatever reverse proxy you work with, such that they do the authentication. See https://github.com/github/orchestrator/blob/master/docs/security.md

402309718:
Hi guy,it deployed with raft
I use multi just diff dba operator and other maintainer not writtable
it is that I login use auth ,but if i want to switch to readonly user in the same web broswer,I can't manually  logout,maybe broswer saves the cookie，in a period of time always the the user logined before and then I restart all the orchestrator instance it turns out that sometime readonly user ,some time readonly user and I should closed the window and reopen it

shlomi-noach:
> maybe broswer saves the cookie

That's how Basic Auth works. I might be able to look into it in the future.

andrein:
> when I try to test auth fuction I found basic extended with AuthenticationMethod=multi,I use readonly user login to non-leader node,but it switch the writable user ,I login to th leader node ,it shows readonly
> 
> about redirect this part whether exists some problem,and if possiable it better add logout icon and add some group ,role

I think I'm hitting the same issue.  Our setup is as follows: 
* 3 node cluster with raft + sqlite. 
* Authentication configured in multi-mode. 

If we log in with the admin user on any node, we have full privileges and the follower nodes proxy our requests to the leader, as expected.

If we log in with the readonly user on the leader node, we have read only privileges, as expected.
If we log in with the readonly user on any of the follower nodes, we have read-write privileges instead of read-only privileges.

[...]

PR #1349, opened 2021-05-01
Remove resetting Auth credentials in reverse-proxy

akatashev:
Related issue: https://github.com/openark/orchestrator/issues/979

### Description

This PR removes setting Auth credentials in reverse-proxy.
@andrein was right about the fact that lines
https://github.com/openark/orchestrator/blob/d536bc655f1dcff0a196b438709958e75bc292cf/go/http/raft_reverse_proxy.go#L42-L45
reset auth credentials for every request going through `reverse-proxy` and therefore provide this request read/write permissions even for a `readonly` user.

These lines are unnecessary, because there are 3 basic scenarios:
1. Username and password are incorrect, so an attempt to sent a request receives `401 Unauthorized`. In this case we shouldn't care about proxying, because it never reaches the proxy.
2. Leader and follower have different configuration. For example, a follower has `"AuthenticationMethod": "multi"` and a leader has `"AuthenticationMethod": "basic"`. But that isn't a proxy problem.
3. Leader and follower are configured correctly, so if Authorization in a request is good for a follower, it should be good for a leader when request is proxied to it. So it means there is no need to modify authorisation for a request.

### Testing
During the whole testing both servers had the following configuration:
```json
{
  "AuthenticationMethod": "multi",
  "HTTPAuthUser":         "dba_team",
  "HTTPAuthPassword":     "time_for_dinner"
}
```

**Before fix behaviour** (Orchestrator version built from the latest `master` state):
```bash
bash-4.2$ curl https://orch-leader:3000/api/forget-cluster/mysql-cluster -k -u readonly:123
{"Code":"ERROR","Message":"Unauthorized","Details":null}
bash-4.2$ curl https://orch-follower:3000/api/forget-cluster/mysql-cluster -k -u readonly:123
{"Code":"OK","Message":"Cluster forgotten: mysql-server-1:3306","Details":null}
```
Request to the leader is rejected, because it's executed as `readonly`, but the same request sent to the follower is accepted, because its user is implicitly changed to `dba_team` by reverse-proxy.

**After fix behaviour:**
```bash
bash-4.2$ curl -k https://orch-leader:3000/api/forget-cluster/mysql-cluster -u readonly:123
{"Code":"ERROR","Message":"Unauthorized","Details":null}
bash-4.2$ curl -k https://orch-follower:3000/api/forget-cluster/mysql-cluster -u readonly:123
{"Code":"ERROR","Message":"Unauthorized","Details":null}
```
Since credentials are not rewritten by reverse-proxy, both requests produce expected results.

To ensure that we didn't break actual functionality, let's execute a valid request through a follower:
```bash
bash-4.2$ curl -k https://orch-follower:3000/api/forget-cluster/mysql-cluster -u dba_team:time_for_dinner
{"Code":"OK","Message":"Cluster forgotten: mysql-server-1:3306","Details":null}
```
It shows that the request was relayed and executed successfully.

### DIFF
--- a/go/http/raft_reverse_proxy.go
+++ b/go/http/raft_reverse_proxy.go
@@ -4,12 +4,10 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
-	"strings"
 
 	"github.com/openark/golib/log"
 
 	"github.com/go-martini/martini"
-	"github.com/openark/orchestrator/go/config"
 	"github.com/openark/orchestrator/go/raft"
 )
 
@@ -39,10 +37,6 @@ func raftReverseProxy(w http.ResponseWriter, r *http.Request, c martini.Context)
 		return
 	}
 	r.Header.Del("Accept-Encoding")
-	switch strings.ToLower(config.Config.AuthenticationMethod) {
-	case "basic", "multi":
-		r.SetBasicAuth(config.Config.HTTPAuthUser, config.Config.HTTPAuthPassword)
-	}
 	proxy := httputil.NewSingleHostReverseProxy(url)
 	proxy.Transport, err = orcraft.GetRaftHttpTransport()
 	if err != nil {


shlomi-noach (approved):
Thank you for the fix and for the analysis!

PR #1361, opened 2021-05-31
explicitly load db drivers, before removing drivers from vendor package

shlomi-noach:
`openark/golib` is to drop imports for `github.com/mattn/go-sqlite3` and `github.com/go-sql-driver/mysql` ; they are not directly used by `openark/golib` and they create overhead when importing `openark/golib`.

We therefore now explicitly load these two packages in `orchestrator`.


shlomi-noach:
Related: https://github.com/outbrain/golib/compare/master...openark:remove-db-dependencies?expand=1

shlomi-noach:
Related: https://github.com/openark/golib/pull/11

PR #1364, opened 2021-06-03
Fixing a minor typo in the message output if the replication is lagging too much

egezonberisha:
Fixing a minor typo of the output message if the replication is lagging too much and the failover aborts.


shlomi-noach (approved):
Thank you

PR #1366, opened 2021-06-08
Ensure to start SQL thread on recovery emergency operation

shlomi-noach:
### Description

In #1309 (see issue #1308) we removed `stop slave sql_thread` & `start slave sql_thread` , because the replica might be committing a large transaction, and that sequence of command could take long due to rollback and roll forward.

However, it still makes sense to run `start slave sql_thread` (and, regardless, even on MariaDB)


shlomi-noach:
closed in favor of https://github.com/openark/orchestrator/pull/1374

PR #1356, opened 2021-05-18
refactor(go mod): migrate to go modules

cndoit18:
During the migration process, some code is updated directly in the vendor folder, causing inconsistent with the upstream code repository.
I move this part of the code to the migrate
folder.

fixes #1355 

Signed-off-by: cndoit18 <cndoit18@outlook.com>

PR #1370, opened 2021-06-17
Set GO111MODULE=auto to allow building with go1.16+

binwiederhier:
Allow building with go 1.16+

### DIFF
--- a/script/build
+++ b/script/build
@@ -23,6 +23,7 @@ scriptdir="$PWD"/script
 version=$(git rev-parse HEAD)
 describe=$(git describe --tags --always --dirty)
 
+export GO111MODULE=auto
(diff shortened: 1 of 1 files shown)


shlomi-noach:
related/redundant by https://github.com/openark/orchestrator/pull/1356/?

binwiederhier:
Ah yes, this will be obsolete when you switch to go modules. I was almost gonna do the go modules change myself because I couldn't get it to build without a go.mod file. 

I'll close this then and am patiently awaiting the go modules change. 

Semi-related: I cannot get orchestrator to work properly inside intellij without lots of warnings and such. Creating a go.mod file solved almost all of these problems. If you have pointers on how to set up intellij that'd be awesome.

PR #1371, opened 2021-06-17
Allow semi-sync even if a replica's promotion rule is "must_not"

### LINKED ISSUE #1369: Semi-sync is not enforced if promotion rule is must_not
_(This is related to https://github.com/openark/orchestrator/issues/1360, and I'll describe more in that ticket)_

We have MySQL hosts in our topology that we'd like to only ever use as a backup host, i.e. we drive backups of these hosts. On these hosts, we have the promotion rule set to `must_not`, meaning that we won't ever fail over to these hosts. 

```
         ---> 2 (prefer, semi-sync replica)
       /
1 (source) 
       \
         ---> 3 (must_not, async replica)
```

In a failover scenario of the source (1), we'd like to enable semi-sync on that backup host (3) to ensure that we're not losing any transactions. At this time, Orchestrator does not support enforcing semi-sync for non-promotable hosts (note the `sendACK`):

```go
	// If async fallback is disallowed, we'd better make sure to enable replicas to
	// send ACKs before START SLAVE. Replica ACKing is off at mysqld startup because
	// some replicas (those that must never be promoted) should never ACK.
	// Note: We assume that replicas use 'skip-slave-start' so they won't
	//       START SLAVE on their own upon restart.
	if instance.SemiSyncEnforced {
		// Send ACK only from promotable instances.
		sendACK := instance.PromotionRule != MustNotPromoteRule
		// Always disable master setting, in case we're converting a former master.
		if err := EnableSemiSync(instanceKey, false, sendACK); err != nil {
			return instance, log.Errore(err)
		}
	}
```

What is the history behind not allowing semi-sync for non-promotable hosts? If I were to implement a config option `AllowSemiSyncForUnpromotableHosts` that would effectively be used like this: 

```go
sendACK := config.AllowSemiSyncForUnpromotableHosts || instance.PromotionRule != MustNotPromoteRule
```

binwiederhier:
This PR resolves https://github.com/openark/orchestrator/issues/1369 by introducing a flag that allows enabling semi-sync on an unpromotable host.

### DIFF
--- a/go/config/config.go
+++ b/go/config/config.go
@@ -192,6 +192,7 @@ type Configuration struct {
 	DetectRegionQuery                          string            // Optional query (executed on topology instance) that returns the region of an instance. If provided, must return one row, one column. Overrides RegionPattern and useful for installments where Region cannot be inferred by hostname
 	DetectPhysicalEnvironmentQuery             string            // Optional query (executed on topology instance) that returns the physical environment of an instance. If provided, must return one row, one column. Overrides PhysicalEnvironmentPattern and useful for installments where env cannot be inferred by hostname
 	DetectSemiSyncEnforcedQuery                string            // Optional query (executed on topology instance) to determine whether semi-sync is fully enforced for master writes (async fallback is not allowed under any circumstance). If provided, must return one row, one column, value 0 or 1.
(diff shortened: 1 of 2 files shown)


binwiederhier:
Superseded by https://github.com/openark/orchestrator/pull/1373

Issue #1360, opened 2021-05-27
When I turn on SemiSyncEnforced, should it be automatically configured when the instance is discovered?


cndoit18:
When I turn on semi-sync enforced replication and import the database nodes via discovery, it doesn't bother to configure the semi-sync option for me.
This makes me need additional operations when I initialize.

binwiederhier:
This is actually a good point. Orchestrator also does not react on when SemiSyncEnforced changes through the DetectSemiSync... query. We're just thinking about implementing something that does the toggling, but it'd be nice to have it happen within Orchestrator.

binwiederhier:
I think this is a dup of https://github.com/openark/orchestrator/issues/1270 possibly.

binwiederhier:
@shlomi-noach if you think this is a good idea, I'd be more than happy to implement it.

shlomi-noach:
Sorry for the delay. 

History: `SemiSyncEnforced` was developed a few years back by the Vitess authors to handle a failover scenario, such that after failover a new replica is enabled as semi-sync. It was about enforcing an _operation_ after failover, never about enforcing a _state_.

Later on I began work on enforcing a state. I went as far as analyzing the topology and generating a structure analysis: `NotEnoughValidSemiSyncReplicasStructureWarning`. See `analysis.go` and `analysis_dao.go`.

Let's first discuss how we might want to go about it. Will `SemiSyncEnforced` imply exactly `SemiSyncMasterWaitForReplicaCount` number of replicas have to be semi-sync? Is _more_ OK? Which replicas do we pick if there's less?

binwiederhier:
I just wanted to let you know that I am very much thinking about a solution for this, and that my team has collectively spent hours and hours on a solution outside of orchestrator, so I am sure we can come up with something elegant within it. I didn't have a ton of time today but I should be able to respond tomorrow. I want to make sure that everything I post is concise and doesn't waste your time.

shlomi-noach:
Awesome. Please take your time.

binwiederhier:
Ok so I don't have a solution yet, but I feel like we can start the conversation (I'm happy to chat on Slack/whatever or even to Zoom if that makes it easier for you). My focus is obviously on our own topology, but we can generalize it later. It's just much easier to think and talk about a concrete example.

### Example topology
We have dozens (soon hundreds) of 3-host clusters that look like this:
```
         ---> 2 (prefer, semi-sync replica)
       /
1 (source) 
       \
[...]

PR #1373, opened 2021-06-19
EnforceSemiSyncReplicas & RecoverLockedSemiSyncMaster - actively enable/disable semi-sync replicas to match master's wait count

binwiederhier:
This is a WIP PR that attempts to address https://github.com/openark/orchestrator/issues/1360.

There are tons of open questions and things missing, but this is the idea.

Open questions (all answered in https://github.com/openark/orchestrator/pull/1373#pullrequestreview-693389059):
1. ~`EnableSemiSync` also manages the master flag. do we really want that? should we not have an `EnableSemiSyncReplica`?~

2. ~Should there be two modes: `EnforceSemiSyncReplicas: exact|enough` (`exact` would handle `MasterWithTooManySemiSyncReplicas` and `LockedSemiSyncMaster`, and `enough` would only handle `LockedSemiSyncMaster`)?~

3. ~`LockedSemiSyncMasterHypothesis` waits `ReasonableReplicationLagSeconds`. I'd like there to be another variable to control the wait time. This seems like it's overloaded.~

TODO:
- [x] properly succeed failover; currently it kinda retries even though it succeeded, not sure why
- [x] discuss downtime behavior with shlomi
- [x] possibly implement `MasterWithIncorrectSemiSyncReplicas`, see PoC: https://github.com/binwiederhier/orchestrator/pull/1
- [x] when a replica is downtimed but replication is enabled, `MasterWithTooManySemiSyncReplicas` does not behave correctly
- [x] `MaybeEnableSemiSyncReplica` does not manage the `master` flag though it previously did (in the new logic only)
- [x] `excludeNotReplicatingReplicas` should be a specific instance, not _all_ non-replicating instances!
- [x] re-test old logic
- [x] handle master failover semi-sync enable/disable
- [x] semi-sync replica priority (or come up with better concept)
- [x] enabled `RecoverLockedSemiSyncMaster` without exact mode
- [x] perform sanity checks in `checkAndRecover*` functions BEFORE enabling/disabling replicas
- [x] add `ReasonableLockedSemiSyncSeconds` with fallback to `ReasonableReplicationLagSeconds`


go/inst/analysis_dao.go
```diff
@@ -531,6 +531,10 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
 					a.Description = "Semi sync master seems to be locked, more samplings needed to validate"
 				}
 				//
+			} else if config.Config.EnforceSemiSyncReplicaCount && a.IsMaster && a.SemiSyncMasterEnabled && a.SemiSyncMasterStatus && a.SemiSyncMasterWaitForReplicaCount > 0 && a.SemiSyncMasterClients > a.SemiSyncMasterWaitForReplicaCount {
+				a.Analysis = MasterWithTooManySemiSyncReplicas
+				a.Description = "Semi sync master has more semi sync replicas than configured"
+				//
```
shlomi-noach, line 537: good!

go/logic/topology_recovery.go
```diff
@@ -1659,6 +1769,8 @@ func getCheckAndRecoverFunction(analysisCode inst.AnalysisCode, analyzedInstance
 		} else {
 			return checkAndRecoverLockedSemiSyncMaster, true
 		}
+	case inst.MasterWithTooManySemiSyncReplicas:
+		return checkAndRecoverMasterWithTooManySemiSyncReplicas, true
```
shlomi-noach, line 1724: :+1:

go/logic/topology_recovery.go
```diff
@@ -1479,14 +1479,124 @@ func checkAndRecoverNonWriteableMaster(analysisEntry inst.ReplicationAnalysis, c
...
+	replicas, err := inst.ReadReplicaInstances(&analysisEntry.AnalyzedInstanceKey)
+	if err != nil {
+		return false, topologyRecovery, err
+	}
+	if len(replicas) == 0 {
+		AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("no replicas found for %+v.", analysisEntry.AnalyzedInstanceKey))
+		return false, topologyRecovery, nil
+	}
+
+	reversePriorityReplicas := replicas // TODO sort replicas by priority
```
shlomi-noach: In my opinion, `promotion_rule`s strongly correlate with semi-sync priority. I'd like to suggest dropping the `priority` query column and work  down the promotion rule `must`>`prefer`>`neutal`>`prefer_not`>`must_not`.

As I'm writing this, I also note that region/datacenter play a critical role. The thing is that some people want semi-sync _only within same region/dc_ and some _only outside region/dc_. Which makes it impossible for `orchestrator` to do prioritization.

Still, I feel like the `priority` query puts a burden on the user, and is not trivial to compute. It's e.g. not trivial to generate the query in the config file in a puppet module.

Perhaps two-level sorting: 1st order is user's `priority`, 2nd is promotion-rule?

More ideas are welcome.

go/logic/topology_recovery.go
```diff
@@ -1479,14 +1479,124 @@ func checkAndRecoverNonWriteableMaster(analysisEntry inst.ReplicationAnalysis, c
...
+	}
+
+	// No way back. We're doing this.
+
+	// TODO before actually doing it, do sanity check (downtimed, etc.)
+
+	enabled := 0
+	for _, replica := range reversePriorityReplicas {
+		if replica.IsDowntimed || replica.SemiSyncReplicaEnabled {
+			continue
```
shlomi-noach: We might actually want to _disable_ semi-sync when we have more than `enough` semi sync replicas enables.

shlomi-noach: To clarify the previous comment: we might have the exact number of desired semi-sync replicas, but wish to reassign to different replicas. As we assign semi-sync to a replica, we may wish to deassign the role from a lower-priority replica.

shlomi-noach (commented):
> `EnableSemiSync` also manages the master flag. do we really want that? should we not have an ...

Right. I suggest moving away from `EnableSemiSync()`. Recall that this function was originally authored by the Vitess authors to manage post-failover semi-sync behavior, specifically for Vitess. What I'm thinking is that that implementation is not inline with `orchetsrator`'s design; your proposed solution is. Let's just not use `EnableSemiSync` _at all_.

For your convenience, the functions `SetSemiSyncMaster` and `SetSemiSyncReplica` already exist and are used by the API/CLI.

> Should there be two modes: EnforceSemiSyncReplicas: exact|enough (exact would handle MasterWithTooManySemiSyncReplicas and LockedSemiSyncMaster, and enough would only handle LockedSemiSyncMaster)?

Yes. Many production deployments will have `1` as `rpl_semi_sync_master_wait_for_slave_count` (BTW MariaDB does not support this variable and it is implicitly always `1`) and with _multiple_ semi-sync replicas. I managed such a deployment it it worked well for us.

> LockedSemiSyncMasterHypothesis waits ReasonableReplicationLagSeconds. I'd like there to be another variable to control the wait time. This seems like it's overloaded.

That makes sense. Let's fallback to `ReasonableReplicationLagSeconds` is the new variable is not configured. It must be non-zero in reality, so the value of `0` can indicate that the value is "not configured".

go/logic/topology_recovery.go
binwiederhier: I like the idea of a two-level priority. I've implemented this idea in the latest PR (though it's not fully tested). It now sorts the replicas in this order:

- priority (I have redefined `SemiSyncEnforced` as an int (lower numbers are higher priorities), which is nice because that actually makes it backwards compatible.
- promotion rule
- instance key (a terrible fallback, but the only way to break a tie)

go/logic/topology_recovery.go
binwiederhier: I have implemented this in the latest PR (still WIP). It now orders the replicas (see https://github.com/openark/orchestrator/pull/1373#discussion_r660566882) and turns semi-sync on/off based on the desired state. 

Aside from `LockedSemiSyncMaster` (too few) and `MasterWithTooManySemiSyncReplicas` (too many), there is also the situation of "the wrong ones" now that may need its own analysis code, maybe `MasterWithIncorrectSemiSyncReplicas`. I'll add that to the TODO list. The remediation for all of them is the same, i.e. the logic in `recoverExactSemiSyncReplicas()`.

binwiederhier:
_You can choose to ignore the comment below if you want to. I think I've convinced myself that this is a good idea and I'll implement it. I can always revert if it doesn't pan out._

-----

Aside from some battles I'm fighting in the code, there's one more open design question regarding `LockedSemiSync`:

In the current iteration of the PR, I made a config option `EnforceSemiSyncReplicas` with three possible values:

- empty (current behavior; do not handle `LockedSemiSyncMaster` or `MasterWithTooManySemiSyncReplicas`)
- `exact` to enable/disable semi sync on replicas to match priority order (for both  `LockedSemiSyncMaster` and `MasterWithTooManySemiSyncReplicas`)
- `enough` to handle `LockedSemiSyncMaster` by _enabling_ replicas in priority order, but never disabling it. 

This requires basically an opt-in to handle `LockedSemiSyncMaster`.

Alternatively we could make a `RecoverLockedSemiSyncMaster bool` flag (like you did in https://github.com/openark/orchestrator/pull/1332 for `RecoverNonWriteableMaster`) and make `EnforceSemiSyncReplicas` a bool too. So we'd have:

```
RecoverLockedSemiSyncMaster bool
EnforceSemiSyncReplicas bool
```

Which would behave like this:

- For `LockedSemiSyncMaster`: `if EnforceSemiSyncReplicas { recoverExactSemiSyncReplicas() } else { recoverEnoughSemiSyncReplicas() }`
- For `MasterWithTooManySemiSyncReplicas` (and later for `MasterWithIncorrectSemiSyncReplicas`): `if EnforceSemiSyncReplicas { recoverExactSemiSyncReplicas() } else { nothing() }`

What do you think?

shlomi-noach:
Sounds good and either choice. If you pick using `EnforceSemiSyncReplicas`, then I'd rename it as `EnforceExactSemiSyncReplicas` to be more precise.

binwiederhier:
I think I am making progress: 

I have now implemented both the "exact" mode (`EnforceExactSemiSyncReplicas=true`, see `allowDisable = true` in `recoverSemiSyncReplicas`), as well as the "enough" mode (`RecoverLockedSemiSyncMaster=true`, but `EnforceExactSemiSyncReplicas=false`, see `allowDisable = false`). Other than possibly making this more resilient and breaking things into smaller functions I am quite happy with the logic. It works nicely.

I've also implemented `ReasonableStaleBinlogCoordinatesSeconds` with a fallback to `ReasonableReplicationLagSeconds` as discussed.

### Questions

1. A scenario this will not detect is if the counts match, but the priority order does not match. I wanted to implement `MasterWithIncorrectSemiSyncReplicas`, but for that we'd need to read more in `GetReplicationAnalysis` and I'm not sure if I can get all that I need in that giant query. What are your thoughts on this?
2. Could you check the `TODO`s in the PR and give me some feedback as to whether I'm doing the right thing. Specifically, I pretty consistently get repeated failure detection happening after a successful recovery until another round of detection happens. Not sure why that is.
3. I have chosen to treat `priority=0` (previously: `semi_sync_enforced=0`) as "this is an async replica". In the "exact" mode (allowDisable=true), that means we'll always disable semi sync on these replicas (see `actions` array); in the "enough" mode (allowDisable=false), I don't touch them at all. Does that make sense to you? Obviously this only applies if `EnforceExactSemiSyncReplicas` or `RecoverLockedSemiSyncMaster` are set.

Here are some logs of what it currently looks like:

### MasterWithTooManySemiSyncReplicas

```
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO executeCheckAndRecoverFunction: proceeding with MasterWithTooManySemiSyncReplicas recovery on pheckel-devm-db-g0-1:3306; isRecoverable?: true; skipProcesses: false
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO topology_recovery: master semi-sync wait count is 1, currently we have 2 semi-sync replica(s)
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO topology_recovery: possible semi-sync replicas (in priority order):
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO topology_recovery: - pheckel-devm-db-g0-2:3306: semi-sync enabled = true, priority = 1, promotion rule = neutral, downtimed = false, last check = true, replicating = true
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO topology_recovery: - pheckel-devm-db-g0-3:3306: semi-sync enabled = true, priority = 1, promotion rule = must_not, downtimed = false, last check = true, replicating = true
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO topology_recovery: always-async replicas: (none)
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO topology_recovery: excluded replicas (downtimed/defunct): (none)
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO topology_recovery: taking actions:
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO topology_recovery: - pheckel-devm-db-g0-3:3306: setting rpl_semi_sync_slave_enabled=false, restarting slave_io thread
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO auditType:register-candidate instance:pheckel-devm-db-g0-3:3306 cluster:pheckel-devm-db-g0-1:3306 message:must_not
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO auditType:register-candidate instance:pheckel-devm-db-g0-3:3306 cluster:pheckel-devm-db-g0-1:3306 message:must_not
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO auditType:register-candidate instance:pheckel-devm-db-g0-1:3306 cluster:pheckel-devm-db-g0-1:3306 message:neutral
Jun 30 01:28:48 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:28:48 INFO topology_recovery: recovery complete; success = true
```

### LockedSemiSyncMaster
Both "exact" and "enough" mode look the same in my setup currently until I get another replica going (I think).

```
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO executeCheckAndRecoverFunction: proceeding with LockedSemiSyncMaster recovery on pheckel-devm-db-g0-1:3306; isRecoverable?: true; skipProcesses: false
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO topology_recovery: master semi-sync wait count is 1, currently we have 0 semi-sync replica(s)
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO topology_recovery: possible semi-sync replicas (in priority order):
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO topology_recovery: - pheckel-devm-db-g0-2:3306: semi-sync enabled = false, priority = 1, promotion rule = neutral, downtimed = false, last check = true, replicating = true
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO topology_recovery: - pheckel-devm-db-g0-3:3306: semi-sync enabled = false, priority = 1, promotion rule = must_not, downtimed = false, last check = true, replicating = true
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO topology_recovery: always-async replicas: (none)
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO topology_recovery: excluded replicas (downtimed/defunct): (none)
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO topology_recovery: taking actions:
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO topology_recovery: - pheckel-devm-db-g0-2:3306: setting rpl_semi_sync_slave_enabled=true, restarting slave_io thread
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO auditType:register-candidate instance:pheckel-devm-db-g0-2:3306 cluster:pheckel-devm-db-g0-1:3306 message:neutral
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO auditType:register-candidate instance:pheckel-devm-db-g0-2:3306 cluster:pheckel-devm-db-g0-1:3306 message:neutral
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO auditType:register-candidate instance:pheckel-devm-db-g0-1:3306 cluster:pheckel-devm-db-g0-1:3306 message:neutral
Jun 30 01:26:12 pheckel-devm-orch-1 orchestrator[591830]: 2021-06-30 01:26:12 INFO topology_recovery: recovery complete; success = true
```

shlomi-noach:
> A scenario this will not detect is if the counts match, but the priority order does not match. I wanted to implement MasterWithIncorrectSemiSyncReplicas, but for that we'd need to read more in GetReplicationAnalysis and I'm not sure if I can get all that I need in that giant query. What are your thoughts on this?

You're right, this will be hard to detect. Expanding the giant query (BTW it's literally commonly referred to as "orchestrator's giant query") to include the extra information is impractical at this time. Let's not deal with this right now. Detection/recovery will only take place if the _number_ of semi-sync replicas is different/smaller than expected, but not when the number is identical, however distribution may be.

Perhaps, in the future, way may hack around this by tricking `orchetrator` into a recovery.

I'll review further later today, and please be advised that I'll then be out till mid next week.

`allowDisable` is a confusing name. Can we make it more descriptive?

binwiederhier:
> MasterWithIncorrectSemiSyncReplicas

I will hold off on this for now.

> I'll review further later today, and please be advised that I'll then be out till mid next week.

No rush.

> allowDisable is a confusing name. Can we make it more descriptive?

I named it `exactReplicaTopology bool` now; it's still a little wonky, but okay.

-----

Today I worked on fixing the `EnableSemiSyncMaster` stuff and now we're correctly enabling the semi-sync flags during a failover. I only tested in a 3-node topology, but I'm hoping that I don't have to change anything to make it work for more replicas.

I left more `TODO` markers for you to look at. I think I managed to keep everything 100% backwards compatible if none of the new flags are enabled, so I think that's good :-)

binwiederhier:
I dabbled today with `MasterWithIncorrectSemiSyncReplicas` -- This works, but it requires changing the giant query and some weird `Instance` parsing; I don't know if I like it. https://github.com/binwiederhier/orchestrator/pull/1

binwiederhier:
Even though this is fun, I think I will **hold off now and wait for your comments**. I think you may want to try it out too, because the behavior is much easier to see when you see it in action.

A few notes/questions (some of which are also in the `TODO`s):

1. `SemiSyncPriority` is a little overloaded, because it currently also controls `rpl_semi_sync_slave_enabled` (now in `MaybeEnableSemiSyncReplica`) as well as `rpl_semi_sync_master_enabled` (now in `MaybeDisableSemiSyncMaster` and `SetReadOnly`). During a `DeadMaster` failover, the master flag is set as a side-effect of `SetReadOnly`, but in a `LockedSemiSyncMaster` or `MasterWithTooManyReplicas` case, it does not toggle the master flag. This is inconsistent. Should we also toggle the master flag when we fix the semi-sync replicas? 
2. `ClassifyAndPrioritizeReplicas` does this "is excluded check" (~ `replica.IsDowntimed || !replica.IsLastCheckValid || !replica.ReplicaRunning()`) and classifies these replicas as "do not change anything for these". This leads to odd behavior for "downtimed" replicas that still have semi-sync enabled. Should we maybe just classify downtimed replicas as "always async" and simply disable semi-sync in that case?
3. Async fallback: During the entire development of this I have not thought about the case of "async fallback" (caused by `rpl_semi_sync_master_timeout`). Is there anything to consider here?
4. I was not able to resolve the topology recovery properly at the end of `recoverSemiSyncReplicas`. For some reason the `MasterWithTooManySemiSyncReplicas` is retriggered a couple of times and I'm not entirely sure why. Can you help with this?
5. Please note that the `MasterWithIncorrectSemiSyncReplicas` PR (https://github.com/binwiederhier/orchestrator/pull/1) is a proof-of-concept that works quite well. It can be hardened to make it less brittle (use a stripped down version of `Instance`, e.g. `ReplicaInstance` so it's safer), and it would make `MasterWithTooManySemiSyncReplicas` obsolete. It is a little awkward though.
6. If it helps, I can record some videos of failovers or we can do a screenshare. Let me know.

No rush with this. I appreciate all your work. This is a great piece of software.

shlomi-noach:
Looking into

binwiederhier:
Awesome. Let me know if you have any questions. I have demoed this inside the company twice already and I like this branch more (https://github.com/binwiederhier/orchestrator/pull/1) :-D

cndoit18 (commented):
Great, I'm looking forward to it. what is the progress now?

go/config/config.go
```diff
@@ -275,6 +275,9 @@ type Configuration struct {
 	KVClusterMasterPrefix                      string            // Prefix to use for clusters' masters entries in KV stores (internal, consul, ZK), default: "mysql/master"
 	WebMessage                                 string            // If provided, will be shown on all web pages below the title bar
 	MaxConcurrentReplicaOperations             int               // Maximum number of concurrent operations on replicas
+	EnforceExactSemiSyncReplicas               bool              // If true, semi-sync replicas will be enabled/disabled to match the wait count in the desired priority order; this applies to LockedSemiSyncMaster and MasterWithTooManySemiSyncReplicas
+	RecoverLockedSemiSyncMaster                bool              // If true, orchestrator will recover from a LockedSemiSync state by enabling semi-sync on replicas to match the wait count; this behavior can be overridden by EnforceExactSemiSyncReplicas
+	ReasonableLockedSemiSyncMasterSeconds      uint              // Time to evaluate the LockedSemiSyncHypothesis before triggering the LockedSemiSync analysis; falls back to ReasonableReplicationLagSeconds if not set
```
shlomi-noach, line 280: Good names

go/inst/analysis.go
```diff
@@ -230,3 +231,12 @@ func (this *ReplicationAnalysis) GetAnalysisInstanceType() AnalysisInstanceType
...
 }
+
+// ReasonableLockedSemiSyncMasterSeconds returns the number of seconds before a LockedSemiSyncMaster is triggered.
+// If the ReasonableLockedSemiSyncMasterSeconds setting is not set, ReasonableReplicationLagSeconds is used.
+func ReasonableLockedSemiSyncMasterSeconds() uint {
+	if config.Config.ReasonableLockedSemiSyncMasterSeconds == 0 {
+		return uint(config.Config.ReasonableReplicationLagSeconds)
+	}
+	return config.Config.ReasonableLockedSemiSyncMasterSeconds
+}
```
shlomi-noach: I'd instead just overwrite `ReasonableLockedSemiSyncMasterSeconds` in `postReadAdjustments()` but this is fine.

go/inst/analysis_dao.go
```diff
@@ -531,6 +531,10 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
 					a.Description = "Semi sync master seems to be locked, more samplings needed to validate"
 				}
 				//
+			} else if config.Config.EnforceExactSemiSyncReplicas && a.IsMaster && a.SemiSyncMasterEnabled && a.SemiSyncMasterStatus && a.SemiSyncMasterWaitForReplicaCount > 0 && a.SemiSyncMasterClients > a.SemiSyncMasterWaitForReplicaCount {
+				a.Analysis = MasterWithTooManySemiSyncReplicas
+				a.Description = "Semi sync master has more semi sync replicas than configured"
+				//
```
shlomi-noach, line 537: :+1:

go/inst/instance_dao.go
```diff
@@ -776,7 +776,7 @@ func ReadTopologyInstanceBufferable(instanceKey *InstanceKey, bufferWrites bool,
 		waitGroup.Add(1)
 		go func() {
 			defer waitGroup.Done()
-			err := db.QueryRow(config.Config.DetectSemiSyncEnforcedQuery).Scan(&instance.SemiSyncEnforced)
+			err := db.QueryRow(config.Config.DetectSemiSyncEnforcedQuery).Scan(&instance.SemiSyncPriority)
```
shlomi-noach, line 793: :+1:

go/inst/instance_topology_dao.go
```diff
@@ -545,13 +547,13 @@ func StartReplicationUntilMasterCoordinates(instanceKey *InstanceKey, masterCoor
...
-		sendACK := instance.PromotionRule != MustNotPromoteRule
-		// Always disable master setting, in case we're converting a former master.
-		if err := EnableSemiSync(instanceKey, false, sendACK); err != nil {
-			return instance, log.Errore(err)
-		}
+	instance, err = MaybeDisableSemiSyncMaster(instance)
+	if err != nil {
+		return instance, log.Errore(err)
+	}
+	instance, err = MaybeEnableSemiSyncReplica(instance)
```
shlomi-noach, line 554: Can you please explain the logic behind `MaybeDisableSemiSyncMaster` immediately followed by `MaybeEnableSemiSyncReplica`?

go/inst/instance_topology_dao.go
```diff
@@ -1108,10 +1268,8 @@ func SetReadOnly(instanceKey *InstanceKey, readOnly bool) (*Instance, error) {
 
 	// If async fallback is disallowed, we're responsible for flipping the master
 	// semi-sync switch ON before accepting writes. The setting is off by default.
-	if instance.SemiSyncEnforced && !readOnly {
-		// Send ACK only from promotable instances.
-		sendACK := instance.PromotionRule != MustNotPromoteRule
-		if err := EnableSemiSync(instanceKey, true, sendACK); err != nil {
+	if instance.SemiSyncPriority > 0 && !readOnly { // TODO I feel like this flag is a little overloaded: this is used here to determine if the MASTER's semi-sync is enabled...
```
shlomi-noach: I agree it's overloaded. On the other hand, creating a `MasterSemiSyncPriority` flag is also a bit messy. Let's keep it as it is in the meantime.

go/logic/topology_recovery.go
```diff
@@ -1482,14 +1482,87 @@ func checkAndRecoverNonWriteableMaster(analysisEntry inst.ReplicationAnalysis, c
...
 		return false, nil, err
 	}
+	if config.Config.EnforceExactSemiSyncReplicas {
+		return recoverSemiSyncReplicas(topologyRecovery, analysisEntry, true)
+	}
+	if !config.Config.RecoverLockedSemiSyncMaster {
+		AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("no action taken to recover locked semi sync master on %+v. Enable RecoverLockedSemiSyncMaster or EnforceExactSemiSyncReplicas change this behavior.", analysisEntry.AnalyzedInstanceKey))
+		return false, nil, err
+	}
+	return recoverSemiSyncReplicas(topologyRecovery, analysisEntry, false)
```
shlomi-noach: A matter of preference. The first `if` is `if config.Config.EnforceExactSemiSyncReplicas then recover...`, and so the 2nd should take the same form. I'd rewrite as
```suggestion
	if config.Config.EnforceExactSemiSyncReplicas {
		return recoverSemiSyncReplicas(topologyRecovery, analysisEntry, true)
	}
	if config.Config.RecoverLockedSemiSyncMaster {
	return recoverSemiSyncReplicas(topologyRecovery, analysisEntry, false)
	}
	AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("no action taken to recover locked semi sync master on %+v. Enable RecoverLockedSemiSyncMaster or EnforceExactSemiSyncReplicas change this behavior.", analysisEntry.AnalyzedInstanceKey))
	return false, nil, err
```

go/logic/topology_recovery.go
```diff
@@ -1482,14 +1482,87 @@ func checkAndRecoverNonWriteableMaster(analysisEntry inst.ReplicationAnalysis, c
...
+
+	// Log analysis
+	AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("semi-sync: analysis results for recovery of %+v:", masterInstance.Key))
+	AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("semi-sync: semi-sync wait count = %d, semi-sync replica count = %d", analysisEntry.SemiSyncMasterWaitForReplicaCount, analysisEntry.SemiSyncMasterClients))
+	inst.LogSemiSyncReplicaAnalysis(possibleSemiSyncReplicas, asyncReplicas, excludedReplicas, actions, func(s string, a ...interface{}) { AuditTopologyRecovery(topologyRecovery, fmt.Sprintf(s, a...)) })
+
+	// Bail out if we cannot succeed
+	if uint(len(possibleSemiSyncReplicas)) < analysisEntry.SemiSyncMasterWaitForReplicaCount {
+		AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("semi-sync: not enough valid live replicas found to recover from %s on %+v.", analysisEntry.AnalysisString(), analysisEntry.AnalyzedInstanceKey))
+		return true, topologyRecovery, nil // TODO recoveryAttempted = true; is this correct? what are the implications of this?
```
shlomi-noach: - If `recoveryAttempted = true` then we run `PostFailoverProcesses` or `PostUnsuccessfulFailoverProcesses`
- If `recoveryAttempted = false` then we skip the processes.
I think we should return `true`, because we're _trying_ to recover, and failing to.

But also we need to return an error. not `nil`. We have failed to recover.

go/logic/topology_recovery.go
```diff
@@ -1482,14 +1482,87 @@ func checkAndRecoverNonWriteableMaster(analysisEntry inst.ReplicationAnalysis, c
...
+	}
+	if len(actions) == 0 {
+		AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("semi-sync: cannot determine actions based on possible semi-sync replicas; cannot recover from %s on %+v.", analysisEntry.AnalysisString(), analysisEntry.AnalyzedInstanceKey))
+		return true, topologyRecovery, nil // TODO recoveryAttempted = true; is this correct? what are the implications of this?
```
shlomi-noach: Here, too, we should return an error, because we have failed to recover.

go/logic/topology_recovery.go
```diff
@@ -1482,14 +1482,87 @@ func checkAndRecoverNonWriteableMaster(analysisEntry inst.ReplicationAnalysis, c
...
+	AuditTopologyRecovery(topologyRecovery, "semi-sync: taking actions:")
+	for replica, enable := range actions {
+		// TODO should we also set master_enabled = false here?
+		AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("semi-sync: - %s: setting rpl_semi_sync_slave_enabled=%t, restarting slave_io thread", replica.Key.String(), enable))
+		_, err := inst.SetSemiSyncReplica(&replica.Key, enable)
+		if err != nil {
+			AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("semi-sync: cannot change semi sync on replica %+v.", replica.Key))
+			return true, topologyRecovery, nil
+		}
+	}
```
shlomi-noach, line 1546: I wonder: say primary expects 5 semi sync replicas and there's 6, and we need to disable one. However, as it turns out in the above loop we first disable 2 then enable 1. At the end, we're logically correct. But along the way, we created a scenario where the primary writes are locked.

I would therefore suggest to loop twice: 1st iteration, only handle replicas where `enable=true`, 2md iteration only handle replicas where `enable=false`.

go/inst/instance_topology_dao.go
```diff
@@ -579,14 +581,172 @@ func StartReplicationUntilMasterCoordinates(instanceKey *InstanceKey, masterCoor
...
+			actions[replica] = false
+		}
+	}
+	for _, replica := range asyncReplicas {
+		if replica.SemiSyncReplicaEnabled {
+			actions[replica] = false
+		}
+	}
+	return actions
+}
```
shlomi-noach, line 728: :+1:

go/inst/instance_topology_dao.go
```diff
@@ -579,14 +581,172 @@ func StartReplicationUntilMasterCoordinates(instanceKey *InstanceKey, masterCoor
...
+		if !replica.SemiSyncReplicaEnabled {
+			actions[replica] = true
+			enabled++
+		}
+		if enabled == waitCount-currentSemiSyncReplicas {
+			break
+		}
+	}
+	return actions
+}
```
shlomi-noach, line 745: :+1:

go/inst/instance_topology_dao.go
```diff
@@ -579,14 +581,172 @@ func StartReplicationUntilMasterCoordinates(instanceKey *InstanceKey, masterCoor
...
+
+	// TODO This is a little odd. We sometimes get actions for replicas that are not us. If we do not take this action
+	// TODO then we'll absolutely have a MasterWithTooManySemiSyncReplicas event following this recovery. We could prevent this
+	// TODO by just executing all actions, but we're in the scope of `replicaInstance` here, so..... idk
+	for replica, enable := range actions {
+		if replica.Key.Equals(&replicaInstance.Key) {
+			log.Infof("semi-sync: %s: setting rpl_semi_sync_slave_enabled: %t", &replicaInstance.Key, enable)
+			return SetSemiSyncReplica(&replicaInstance.Key, enable)
+		}
+	}
```
shlomi-noach, line 658: Feels like this code and `recoverSemiSyncReplicas` have a lot in common. I don't feel strongly, but I think there's a significant amount of logic we can extract to an external function.

go/inst/instance_topology_dao.go
```diff
@@ -579,14 +581,172 @@ func StartReplicationUntilMasterCoordinates(instanceKey *InstanceKey, masterCoor
...
+// ClassifyAndPrioritizeReplicas takes a list of replica instances and classifies them based on their semi-sync priority, excluding
+// replicas that are downtimed or down. It furthermore prioritizes the possible semi-sync replicas based on SemiSyncPriority, PromotionRule
+// and hostname (fallback).
+func ClassifyAndPrioritizeReplicas(replicas []*Instance, includeNonReplicatingInstance *InstanceKey) (possibleSemiSyncReplicas []*Instance, asyncReplicas []*Instance, excludedReplicas []*Instance) {
+	// Filter out downtimed and down replicas
+	possibleSemiSyncReplicas = make([]*Instance, 0)
+	asyncReplicas = make([]*Instance, 0)
+	excludedReplicas = make([]*Instance, 0)
+	for _, replica := range replicas {
+		if replica.IsDowntimed || !replica.IsLastCheckValid || (!replica.Key.Equals(includeNonReplicatingInstance) && !replica.ReplicaRunning()) {
```
shlomi-noach: This logic is confusing to read: too many negatives or negatives (and negatives or negatives)...
Let's convert this `if-else` into a `switch` statement, and break down the specific `if` clause into multiple `case`s.

shlomi-noach (commented):
Please see inline comments

shlomi-noach:
> SemiSyncPriority is a little overloaded

I agree, and it's confusing. Separate into two distinct variables? Not sure.

> This leads to odd behavior for "downtimed" replicas that still have semi-sync enabled. Should we maybe just classify downtimed replicas as "always async" and simply disable semi-sync in that case?

Different users perceive Downtime to mean different things. The official meaning of Downtime is: `orchestrator` will not failover _this_ instance if it's downtimed. But is a primary crashes, `orchestrator` may promote a replica that is downtimed (exception is graceful takeover).
So let's treat the downtimed replicas just like any other replica.

> Async fallback: During the entire development of this I have not thought about the case of "async fallback" (caused by rpl_semi_sync_master_timeout). Is there anything to consider here?

I think not. That's for MySQL to handle. We will still attempt to set the right number of semi-sync replicas.

> I was not able to resolve the topology recovery properly at the end of recoverSemiSyncReplicas. For some reason the MasterWithTooManySemiSyncReplicas is retriggered a couple of times and I'm not entirely sure why. Can you help with this?

I think that may be because you did not return an error (see inline comments). Please let me know if this isn't the solution.

binwiederhier:
Thank you very much for the comments. I appreciate it!

Before I can start addressing all your comments, since I have two different branches, do you have a preference on whether you'd like `MasterWithTooManySemiSyncReplicas` (this branch) or `MasterWithIncorrectSemiSyncReplicas` (https://github.com/binwiederhier/orchestrator/pull/1). I can probably address the comments in both branches, but it'd make it easier if we can make a decision sooner rather than later. My preference is `MasterWithIncorrectSemiSyncReplicas` because it handles _all_ situations, but I understand if you'd want to go the other way.

go/inst/instance_topology_dao.go
binwiederhier: This was an attempt to decouple the master enabled status from the replica enabled status (related to the overloaded `SemiSyncEnforced` (now: `SemiSyncPriority`) flag, while also keeping the logic that existed before.

Before (pseudo-code):
```
if SemiSyncEnforced {
   SetSemiSyncMaster(false)
   SetSemiSyncReplica(sendACK)
}
```

Now:
``` 
if SemiSyncPriority > 0 {
  SetSemiSyncMaster(false)
}
if <we should enable semi-sync replica according to the priority and such> {
  SetSemiSyncReplica(true)
} else {
  SetSemiSyncReplica(false)
}
```

The assumption here is that if you're starting replication, you'll want to disable the semi-sync master, which I agree with (at least in our topology). We have observed that not disabling the semi-sync master on the replicas leads to our dashboards being filled with "failed semi-sync transactions" (no_tx I believe), so I think this is the right course of action. 

However, I cannot speak for other topologies such as this A -> B -> C in which B is also a semi-sync master. That would not have been supported currently though either.

Does that make it clearer?

go/inst/instance_topology_dao.go
binwiederhier: You are correct. I'll make another attempt to harmonize them. Though the "view of the world" is different: 
- `recoverSemiSyncReplicas` looks wholistically and is "allowed" to take actions for all replicas
- `MaybeEnableSemiSyncReplica` is obviously focused only on _one_ replica, which explains the `TODO` comment.

Do you have any opinions on that? The way it is implemented may lead to another recovery event, but I don't think that's too terrible, though a little disappointing.

shlomi-noach:
I was only reviewing this PR, this branch, so let's go with that.

go/logic/topology_recovery.go
binwiederhier: Great observation; I'll adjust the logic to loop twice. 

Do you know if `stop slave io_thread; start slave io_thread` is 100% synchronous, i.e. once we return from SetSemiSyncReplica, do is the replica up and running again? If not, we may have multiple replicas restarting replication at the same time creating a locked state temporarily.

binwiederhier:
~I've worked on it more but I'm still not done with all your comments; I'll work more on it tomorrow. In particular I'll try to make the "two loop action" thing a little more streamlined and i may drop the `logFn` stuff and just go with `log.Debugf` and `log.Infof`.~

I think I am done with all of your comments. Let me know if I overlooked something. I removed the code duplication too, which means there are a lot of changes that you may want to re-review. I left the "taking actions" stuff in the individual functions (`MaybeEnableSemiSyncReplica` and `recoverSemiSyncReplicas`) because it was too awkward to combine them. It's much clearer like this.

----

> I think that may be because you did not return an error (see inline comments). Please let me know if this isn't the solution.

This did not seem to help. I think I'll make a video or two tomorrow so you can see what I mean.

> So let's treat the downtimed replicas just like any other replica.

Done, though this is a little odd to me. If we bring in a new replica from a failover and it'll take an hour or two to catch up to replication, this replica being the "semi-sync enabled" one is definitely odd.

> I agree, and it's confusing. Separate into two distinct variables? Not sure.

I've left it like it was for now; though I think to make the behavior consistent at least, we need to call MaybeDisableSemiSyncMaster during the recovery of MasterWithTooManySemiSyncReplicas and LockedSemiSyncMaster.

go/logic/topology_recovery.go
shlomi-noach: it's not 100% synchronous and the situation you describe is possible. I'd take the risk on that, though.

go/inst/instance_topology_dao.go
shlomi-noach: Your call. I don't feel strongly.

shlomi-noach:
> Done, though this is a little odd to me. If we bring in a new replica from a failover and it'll take an hour or two to catch up to replication, this replica being the "semi-sync enabled" one is definitely odd.

I'm open to suggestions, I feel like any suggestion would either be a bit odd or be a bit complex.

go/inst/instance_topology_dao.go
binwiederhier: I updated this to make it a little more readable, though I left the if/else. I don't quite know how to make this a switch case that is more readable. Maybe you can help:

```golang
for _, replica := range replicas {
	isReplicating := replica.Key.Equals(includeNonReplicatingInstance) || replica.ReplicaRunning()
	if !replica.IsLastCheckValid || !isReplicating {
		excludedReplicas = append(excludedReplicas, replica)
	} else if replica.SemiSyncPriority == 0 || replica.IsDowntimed {
		asyncReplicas = append(asyncReplicas, replica)
	} else {
		possibleSemiSyncReplicas = append(possibleSemiSyncReplicas, replica)
	}
}
```

binwiederhier:
**I think I'm "done"** and reasonably happy with the result. Please let me know what else you'd want, especially tests and such.

I've uploaded a video showing three cases (with `EnforceExactSemiSyncReplicas: true`):
- Video: https://phil.nopaste.net/jHcg1bB5rN?d=1&f=orch-demo.mp4&a=Pf7hFfWC89
- Log output from the demo: https://phil.nopaste.net/BfTtK0RggZ?a=ywx2nLkl8C

Cases covered in video/logs:
1. _Starting at 18:09:39:_ Failure of our only semi-sync replica (...-db-g0-2), leading to `LockedSemiSyncMaster`
2. _Starting at 18:12:39:_ Re-adding of that replica and manually clicking "Start replication", leading to `MaybeEnableSemiSyncReplica` being called (which does NOT correct the situation) and a subsequent `MasterWithTooManySemiSyncReplicas` recovery (*starting at 18:12:43*).
3. _Starting at 18:14:00:_ Correct semi-sync handling during a `DeadMaster`

**Notable decisions I've made (debatable):**
- I've removed the `IsDowntimed` logic as requested; I think that is the right move unless we decide to go with `MasterWithIncorrectSemiSyncReplicas` eventually
- I am now disabling the semi-sync master flag in the recovery to match the behavior in `SetReadOnly`

**Unresolved issues:**
- We are still re-triggering recoveries right after they are resolved 4-5 times

binwiederhier:
Friendly reminder. :grin:

shlomi-noach:
> Friendly reminder. grin

Yup! I hope to conclude tomorrow

go/inst/instance_dao.go
```diff
@@ -204,6 +204,20 @@ func ReadTopologyInstance(instanceKey *InstanceKey) (*Instance, error) {
...
+// ReadTopologyInstances is a convenience method that calls ReadTopologyInstance
+// for all the instance keys and returns a slice of Instance.
+func ReadTopologyInstances(instanceKeys []InstanceKey) ([]*Instance, error) {
+	instances := make([]*Instance, 0)
+	for _, replicaKey := range instanceKeys {
+		replica, err := ReadTopologyInstance(&replicaKey)
+		if err != nil {
+			return nil, err
+		}
+		instances = append(instances, replica)
```
shlomi-noach: we are reading instances, not replicas. Change suggested
```suggestion
	for _, instanceKey := range instanceKeys {
		instance, err := ReadTopologyInstance(&instanceKey)
		if err != nil {
			return nil, err
		}
		instances = append(instances, instance)
```

go/inst/instance_topology_dao.go
shlomi-noach: Thank you, yes!

go/logic/topology_recovery.go
```diff
@@ -1482,14 +1482,72 @@ func checkAndRecoverNonWriteableMaster(analysisEntry inst.ReplicationAnalysis, c
...
+	for replica, enable := range actions {
+		if !enable {
+			AuditTopologyRecovery(topologyRecovery, fmt.Sprintf("semi-sync: - %s: setting rpl_semi_sync_slave_enabled=%t, restarting slave_io thread", replica.Key.String(), enable))
+			if _, err := inst.SetSemiSyncReplica(&replica.Key, enable); err != nil {
+				return true, topologyRecovery, fmt.Errorf("cannot disable semi sync on replica %+v", replica.Key)
+			}
+		}
+	}
```
shlomi-noach, line 1546: :+1:

shlomi-noach (changes requested):
Looks good! One cosmetic requested change, see inline

shlomi-noach:
> We are still re-triggering recoveries right after they are resolved 4-5 times

Do you mean: immediately following a recovery there's another recovery running to fix the exact same problem, even though it's already fixed, and this repeats 4-5 times?

If this is the case, then something is missing: we have probably not marked the recovery as successful. Alternatively, anti-flapping doesn't work for this scenario. Also, is there a detection that still claims a semi-sync issue? I can understand if it takes 5-10 seconds for `orchestrator` to diagnose the resolved scenario. We should ensure that detections during that time are ignored.

binwiederhier:
> Do you mean: immediately following a recovery there's another recovery running to fix the exact same problem, even though it's already fixed, and this repeats 4-5 times?

It's detecting the same situation again a few times after it is resolved. The recovery registration is guarding the repeated execution, so it's fine. It's just ugly:

```
2021-07-14 18:11:25 ERROR AttemptRecoveryRegistration: instance pheckel-devm-db-g0-1:3306 has recently been promoted (by failover of pheckel-devm-db-g0-1:3306) and is in active period. It will not be failed over. You may acknowledge the failure on 
```

You can see the full output here: https://phil.nopaste.net/BfTtK0RggZ?a=ywx2nLkl8C -- Here's a longer excerpt:

```
2021-07-14 18:11:24 INFO executeCheckAndRecoverFunction: proceeding with LockedSemiSyncMaster recovery on pheckel-devm-db-g0-1:3306; isRecoverable?: true; skipProcesses: false
....
2021-07-14 18:11:24 INFO topology_recovery: semi-sync: recovery complete; success = true
...
2021-07-14 18:11:25 INFO topology_recovery: done running PostFailoverProcesses hooks
2021-07-14 18:11:25 INFO topology_recovery: Waiting for 0 postponed functions
2021-07-14 18:11:25 DEBUG PostponedFunctionsContainer: waiting on 0 postponed functions
2021-07-14 18:11:25 DEBUG PostponedFunctionsContainer: done waiting
2021-07-14 18:11:25 INFO topology_recovery: Executed 0 postponed functions

2021-07-14 18:11:25 INFO executeCheckAndRecoverFunction: proceeding with LockedSemiSyncMaster detection on pheckel-devm-db-g0-1:3306; isActionable?: true; skipProcesses: false
2021-07-14 18:11:25 INFO checkAndExecuteFailureDetectionProcesses: could not register LockedSemiSyncMaster detection on pheckel-devm-db-g0-1:3306
2021-07-14 18:11:25 INFO executeCheckAndRecoverFunction: proceeding with LockedSemiSyncMaster recovery on pheckel-devm-db-g0-1:3306; isRecoverable?: true; skipProcesses: false
2021-07-14 18:11:25 ERROR AttemptRecoveryRegistration: instance pheckel-devm-db-g0-1:3306 has recently been promoted (by failover of pheckel-devm-db-g0-1:3306) and is in active period. It will not be failed over. You may acknowledge the failure on pheckel-devm-db-g0-1:3306 (-c ack-instance-recoveries) to remove this blockage
2021-07-14 18:11:25 INFO topology_recovery: found an active or recent recovery on pheckel-devm-db-g0-1:3306. Will not issue another RecoverLockedSemiSyncMaster.
2021-07-14 18:11:25 ERROR 2021-07-14 18:11:25 ERROR AttemptRecoveryRegistration: instance pheckel-devm-db-g0-1:3306 has recently been promoted (by failover of pheckel-devm-db-g0-1:3306) and is in active period. It will not be failed over. You may acknowledge the failure on pheckel-devm-db-g0-1:3306 (-c ack-instance-recoveries) to remove this blockage

2021-07-14 18:11:26 INFO executeCheckAndRecoverFunction: proceeding with LockedSemiSyncMaster detection on pheckel-devm-db-g0-1:3306; isActionable?: true; skipProcesses: false
2021-07-14 18:11:26 INFO executeCheckAndRecoverFunction: proceeding with LockedSemiSyncMaster recovery on pheckel-devm-db-g0-1:3306; isRecoverable?: true; skipProcesses: false
2021-07-14 18:11:26 ERROR AttemptRecoveryRegistration: instance pheckel-devm-db-g0-1:3306 has recently been promoted (by failover of pheckel-devm-db-g0-1:3306) and is in active period. It will not be failed over. You may acknowledge the failure on pheckel-devm-db-g0-1:3306 (-c ack-instance-recoveries) to remove this blockage
2021-07-14 18:11:26 INFO topology_recovery: found an active or recent recovery on pheckel-devm-db-g0-1:3306. Will not issue another RecoverLockedSemiSyncMaster.
2021-07-14 18:11:26 ERROR 2021-07-14 18:11:26 ERROR AttemptRecoveryRegistration: instance pheckel-devm-db-g0-1:3306 has recently been promoted (by failover of pheckel-devm-db-g0-1:3306) and is in active period. It will not be failed over. You may acknowledge the failure on pheckel-devm-db-g0-1:3306 (-c ack-instance-recoveries) to remove this blockage

2021-07-14 18:11:27 INFO executeCheckAndRecoverFunction: proceeding with LockedSemiSyncMaster detection on pheckel-devm-db-g0-1:3306; isActionable?: true; skipProcesses: false
2021-07-14 18:11:27 INFO executeCheckAndRecoverFunction: proceeding with LockedSemiSyncMaster recovery on pheckel-devm-db-g0-1:3306; isRecoverable?: true; skipProcesses: false
2021-07-14 18:11:27 ERROR AttemptRecoveryRegistration: instance pheckel-devm-db-g0-1:3306 has recently been promoted (by failover of pheckel-devm-db-g0-1:3306) and is in active period. It will not be failed over. You may acknowledge the failure on pheckel-devm-db-g0-1:3306 (-c ack-instance-recoveries) to remove this blockage
2021-07-14 18:11:27 INFO topology_recovery: found an active or recent recovery on pheckel-devm-db-g0-1:3306. Will not issue another RecoverLockedSemiSyncMaster.
2021-07-14 18:11:27 ERROR 2021-07-14 18:11:27 ERROR AttemptRecoveryRegistration: instance pheckel-devm-db-g0-1:3306 has recently been promoted (by failover of pheckel-devm-db-g0-1:3306) and is in active period. It will not be failed over. You may acknowledge the failure on pheckel-devm-db-g0-1:3306 (-c ack-instance-recoveries) to remove this blockage
2021-07-14 18:11:33 DEBUG analysis: ClusterName: pheckel-devm-db-g0-1:3306, IsMaster: false, LastCheckValid: false, LastCheckPartialSuccess: false, CountReplicas: 0, CountValidReplicas: 0, CountValidReplicatingReplicas: 0, CountLaggingReplicas: 0, CountDelayedReplicas: 0, CountReplicasFailingToConnectToMaster: 0
2021-07-14 18:11:34 ERROR ReadTopologyInstance(pheckel-devm-db-g0-2:3306) show global status like 'Uptime': dial tcp 10.40.182.188:3306: connect: connection refused
2021-07-14 18:11:34 WARNING DiscoverInstance(pheckel-devm-db-g0-2:3306) instance is nil in 0.007s (Backend: 0.006s, Instance: 0.001s), error=dial tcp 10.40.182.188:3306: connect: connection refused
2021-07-14 18:11:39 INFO auditType:forget-unseen-differently-resolved instance::0 cluster: message:Forgotten instances: 0
2021-07-14 18:11:39 INFO auditType:forget-clustr-aliases instance::0 cluster: message:Forgotten aliases: 0
```

Note how executeCheckAndRecoverFunction is executed multiple times and (luckily) fails with `ERROR AttemptRecoveryRegistration`. It may just be that it takes this long for everything to fully recover though.

> If this is the case, then something is missing: we have probably not marked the recovery as successful. Alternatively, anti-flapping doesn't work for this scenario. Also, is there a detection that still claims a semi-sync issue? I can understand if it takes 5-10 seconds for orchestrator to diagnose the resolved scenario. We should ensure that detections during that time are ignored.

I will investigate this a little more today and tomorrow. If my understanding is correct, the analysis if completely independent of the recovery, so during the recovery, we will trigger and attempt recovery multiple times anyway. So this may in fact be an un-avoidable situation, since it takes a while to restart the replica threads and all that.

binwiederhier:
I've added two commits to remedy the situation. It is working nicely now. We're now re-reading the master after taking actions until the desired state (based on the count only).

Looks like this in the logs:

```17:56:52 DEBUG semi-sync: analysis results for recovery of cluster pheckel-devm-db-g0-1:3306:
17:56:52 DEBUG semi-sync: master = pheckel-devm-db-g0-1:3306, master semi-sync wait count = 1, master semi-sync replica count = 2
17:56:52 DEBUG semi-sync: possible semi-sync replicas (in priority order):
17:56:52 DEBUG semi-sync: - pheckel-devm-db-g0-2:3306: semi-sync enabled = true, priority = 1, promotion rule = neutral, last check = true, replicating = true
17:56:52 DEBUG semi-sync: - pheckel-devm-db-g0-3:3306: semi-sync enabled = true, priority = 1, promotion rule = must_not, last check = true, replicating = true
17:56:52 DEBUG semi-sync: always-async replicas: (none)
17:56:52 DEBUG semi-sync: excluded replicas (defunct): (none)
17:56:52 DEBUG semi-sync: suggested actions:
17:56:52 DEBUG semi-sync: - pheckel-devm-db-g0-3:3306: should set semi-sync enabled = false
17:56:52 INFO topology_recovery: semi-sync: taking actions:
17:56:52 INFO topology_recovery: semi-sync: - pheckel-devm-db-g0-3:3306: setting rpl_semi_sync_slave_enabled=false, restarting slave_io thread
17:56:52 INFO topology_recovery: semi-sync: waiting for desired state:
17:56:52 INFO topology_recovery: semi-sync: - current semi-sync replica count = 2, desired semi-sync replica count = 1
17:56:53 INFO topology_recovery: semi-sync: - current semi-sync replica count = 1, desired semi-sync replica count = 1
17:56:53 INFO topology_recovery: semi-sync: recovery complete; success = true
```

Note the section `waiting for desired state`. This waits up to `WaitForSemiSyncRecoverySeconds`, which defaults to 3 * `InstancePollSeconds`.

Please find the commits for this here: 
- Add "wait for desired state": https://github.com/openark/orchestrator/pull/1373/commits/664d3c116582c6335020175e5bd103f04fb28302
- Add config option: https://github.com/openark/orchestrator/pull/1373/commits/5382128e7e5f9aa9f5c8d74e8556a3d1b6f537b5

binwiederhier:
I think I'm happy with this. Let me know if there is anything else you'd like changed.

shlomi-noach:
I'm gonna look at the superfluous detection cycle. I'd like to rollback https://github.com/openark/orchestrator/commit/664d3c116582c6335020175e5bd103f04fb28302 as well as https://github.com/openark/orchestrator/commit/5382128e7e5f9aa9f5c8d74e8556a3d1b6f537b5. Blocking the recovery process while waiting for results does not align with how orchestrator runs things.

binwiederhier:
It works fine without the two commits, so I'm fine with removing them. Let me know if you'd like to revert them.

The reason why the detection re-triggers is that when we enable/disable replication on the replicas, the Rpl_master_semi_sync_clients variable on the master is only updated after the clients properly connect. So the whole thing typically takes 1-2 seconds before getting into a proper state. If your poll interval is 5 seconds or even 20 seconds, then you'll re-trigger the detection for that long. 

In my tests, I _never_ had to wait longer than 1-2 seconds in the detection loop.

shlomi-noach:
Gotcha, thanks.

One last thing: this has to be documented?

binwiederhier:
> Gotcha, thanks.

Does that mean you'd like me to remove the two commits?

> One last thing: this has to be documented?

You mean in the Wiki? I can do that if you like.

shlomi-noach:
> Does that mean you'd like me to remove the two commits?

Yes please.

> You mean in the Wiki? 

I'm thinking under https://github.com/openark/orchestrator/blob/master/docs/configuration-discovery-classifying.md and under https://github.com/openark/orchestrator/blob/master/docs/configuration-recovery.md. WDYT?

binwiederhier:
Sounds good. I'll do it tomorrow.

binwiederhier:
I added docs. Let me know what you think. They are best viewed on Github:

- [Semi-sync topology](https://github.com/binwiederhier/orchestrator/blob/enforce-semi-sync-replica-count/docs/configuration-discovery-classifying.md#semi-sync-topology)
- [LockedSemiSyncMaster](https://github.com/binwiederhier/orchestrator/blob/enforce-semi-sync-replica-count/docs/failure-detection.md#lockedsemisyncmaster)
- [MasterWithTooManyReplicas](https://github.com/binwiederhier/orchestrator/blob/enforce-semi-sync-replica-count/docs/failure-detection.md#masterwithtoomanysemisyncreplicas)

Two side notes:
- I read your [blog post](http://code.openark.org/blog/mysql/reducing-my-oss-involvement-and-how-it-affects-orchestrator-gh-ost) regarding your hiatus. Will you be cutting _one last release_ that includes these changes? (_Also: It's super sad that you're leaving Orchestrator behind, but totally understandable!_)
- I will be out of office for 3 weeks starting Wednesday; I'll probably still check in, but I'll mostly be away.

shlomi-noach:
Awesome job on the documentation, much appreciated.

Yes, I'll create one more release with these changes! Won't leave them stranded.

As I mentioned in my post I may be back after some break. I do love and enjoy this product, I'm just unable to keep up.

shlomi-noach (approved):
Thank you!

cndoit18:
👍

PR #1383, opened 2021-07-16
Compatible with single primary replication group slave node query

qihengshan:
### Description

When the bank-end architecture is in Replication Group mode, the values of the count_replicas and count_valid_replicas variables obtained by the function GetReplicationAnalysis are always 0. 
After modification, it is compatible with Replication Group mode, and normal members of the replication group are treated as counter_replicas and count_valid_replicas.


qihengshan:
modify necessary

PR #1387, opened 2021-07-19
Compatible with single primary Replication Group slave node query

qihengshan:
### Description

When the bank-end architecture is single Primary Replication Group mode, the values of the count_replicas and count_valid_replicas and count_valid_replicating_replicas and slave_hosts variables obtained by the function GetReplicationAnalysis are always 0 or null.

After modification, it is compatible with single primary Replication Group mode, and normal members of the replication group are treated as counter_replicas and count_valid_replicas and count_valid_replicating_replicas and slave_hosts.

Please check again, thanks!

### DIFF
--- a/go/inst/analysis_dao.go
+++ b/go/inst/analysis_dao.go
@@ -191,12 +191,13 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
 			) = master_instance.cluster_name
 		) AS is_cluster_master,
 		MIN(master_instance.gtid_mode) AS gtid_mode,
-		COUNT(replica_instance.server_id) AS count_replicas,
+		-- Consider GR group members as replicas
+		GREATEST(COUNT( replica_instance.server_id ), COUNT( member_instance.server_id )-1 ) AS count_replicas,
+		-- Consider GR group members as valid replicas
 		IFNULL(
-			SUM(
-				replica_instance.last_checked <= replica_instance.last_seen
-			),
-			0
+			SUM( replica_instance.last_checked <= replica_instance.last_seen ), 
+			SUM( member_instance.last_checked <= member_instance.last_seen 
+				AND member_instance.replication_group_member_state = 'ONLINE')
 		) AS count_valid_replicas,
 		IFNULL(
 			SUM(
@@ -370,6 +371,12 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
 				master_instance.hostname
 			) = replica_instance.master_host
 			AND master_instance.port = replica_instance.master_port
+			AND replica_instance.replication_group_name = ''
+		)
+		LEFT JOIN database_instance member_instance ON (
+			master_instance.replication_group_name = member_instance.replication_group_name 
+			AND master_instance.PORT = member_instance.PORT
+			AND member_instance.replication_group_name != ''
 		)
 		LEFT JOIN database_instance_maintenance ON (
 			master_instance.hostname = database_instance_maintenance.hostname
--- a/go/inst/analysis_dao.go
+++ b/go/inst/analysis_dao.go
@@ -192,7 +192,12 @@ func GetReplicationAnalysis(clusterName string, hints *ReplicationAnalysisHints)
 		) AS is_cluster_master,
 		MIN(master_instance.gtid_mode) AS gtid_mode,
 		-- Consider GR group members as replicas
-		GREATEST(COUNT( replica_instance.server_id ), COUNT( member_instance.server_id ) ) AS count_replicas,
+		CASE 
+			WHEN COUNT( replica_instance.server_id ) > COUNT( member_instance.server_id ) THEN
+				COUNT( replica_instance.server_id ) 
+			ELSE 
+				COUNT( member_instance.server_id )
+			END AS count_replicas,
 		-- Consider GR group members as valid replicas
 		IFNULL(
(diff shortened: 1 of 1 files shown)


shlomi-noach:
> After modification, it is compatible with single primary Replication Group mode,

Hi! I'm not sure. _Should it_ be compatible with single primary replication group node? Do we want GR replicas to be treated like regular replicas? I'm not sure at all. What will be the implications? 

I'm asking all those questions in honest. I'm not sure we want this and I'm not sure what will be the implications. I need more input and more thinking, and since I don't run GR myself, I need it from other people.

qihengshan:
> > After modification, it is compatible with single primary Replication Group mode,
> 
> Hi! I'm not sure. _Should it_ be compatible with single primary replication group node? Do we want GR replicas to be treated like regular replicas? I'm not sure at all. What will be the implications?
> 
> I'm asking all those questions in honest. I'm not sure we want this and I'm not sure what will be the implications. I need more input and more thinking, and since I don't run GR myself, I need it from other people.


Thank you very much for your reply, the following are my thoughts:

Many Internet applications use database proxy middleware. In the traditional MASTER-SLAVE replication cluster mode, after the master node goes down, the orchestrator will expose the new master node through the hook function. The database proxy middleware can automatically switch to the new master database. perfect.

In single primary replication group mode, when the primary node is down (mysql instance is down or host is down), the replication group will re-elect the primary node.
Compatible with query replication group members as regular replicas, people can easily expose the new primary node to the database middleware through the existing framework, so as to satisfy the self-healing of service failure.

For example: Relying on the existing ReplicationAnalysis structure, the judgment logic can be easily realized, thereby realizing the exposure of the new primary node to database agent.

qihengshan:
Another method is to expand the ReplicationAnalysis structure and add GR-related query fields in the GetReplicationAnalysis query to obtain the required GR information. Then, based on the information, realize the logic of exposing the new primary node.

But I'm not sure which method is better, but based on the existing ReplicationAnalysis structure, the changes are minimal.

qihengshan:
> > > After modification, it is compatible with single primary Replication Group mode,
> > 
> > 
> > Hi! I'm not sure. _Should it_ be compatible with single primary replication group node? Do we want GR replicas to be treated like regular replicas? I'm not sure at all. What will be the implications?
> > I'm asking all those questions in honest. I'm not sure we want this and I'm not sure what will be the implications. I need more input and more thinking, and since I don't run GR myself, I need it from other people.
> 
> Thank you very much for your reply, the following are my thoughts:
> 
> Many Internet applications use database proxy middleware. In the traditional MASTER-SLAVE replication cluster mode, after the master node goes down, the orchestrator will expose the new master node through the hook function. The database proxy middleware can automatically switch to the new master database. perfect.
> 
> In single primary replication group mode, when the primary node is down (mysql instance is down or host is down), the replication group will re-elect the primary node.
> Compatible with query replication group members as regular replicas, people can easily expose the new primary node to the database middleware through the existing framework, so as to satisfy the self-healing of service failure.
> 
> For example: Relying on the existing ReplicationAnalysis structure, the judgment logic can be easily realized, thereby realizing the exposure of the new primary node to database agent.

the judgment logic example:
```// Group member is not reachable, has replicas, and none of its reachable replicas can replicate from it
if !a.LastCheckValid && a.CountReplicas > 0 && a.CountValidReplicatingReplicas == 0 {

	a.Analysis = DeadReplicationGroupMemberWithReplicas
	a.Description = "Group member is unreachable and all its reachable replicas are not replicating"

} else if !a.LastCheckValid && a.IsMaster && a.IsReplicationGroupMember && a.CountReplicas > 0 &&
	a.CountReplicas == a.CountValidReplicas {

	a.Analysis = DeadReplicationGroupPrimaryWithoutReplicas
	a.Description = "Group primary is unreachable but the other members is replicating"

}
```

qihengshan:
Thanks for your check, I noticed where the system tests failed
```sh
Testing: analysis-locked-semi-sync-master/01-analysis-locked-hypothesis
twenty one
ERROR analysis-locked-semi-sync-master/01-analysis-locked-hypothesis output does not match expect_output
```
Also saw the PR: https://github.com/openark/orchestrator/pull/1175
It seems that this problem has been solved.

When LockedSemiSyncMasterHypothesis is sent, the orchestrator has done an urgent write operation, update Instance
 and database_instance_stale_binlog_coordinates information.

I also read carefully when LockedSemiSyncMaster occurs, there will only be two results
1. do nothing
2. found an active or recent recovery on sync master, then do nothing

I need some advice, continue to the next step

shlomi-noach:
Thank you so much for explaining. I'm afraid I'm still unsure how to proceed here. I confess I don't have enough understanding of GR and what the outcomes may be. If you consider GR members as valid replicas, how does that affect a failover scenario? Is this going to create different situations than today?

I'm very sorry indeed, I appreciate your time in this, but I'm not certain about the change.

qihengshan:
> how does that affect a failover scenario?

GR can re-elect a new primary node by itself, we or orchestrator do nothing, when the orchestrator detects that the primary node is down, we only find the new primary node from the database_instance table and expose it to hook function.  
> PostMasterFailoverProcesses

shlomi-noach:
I'm very sorry indeed, and thank you for the contribution. I appreciate your time in explaining this. I'm still not sure if this aligns with existing orchestrator logic, and I'm down to this: http://code.openark.org/blog/mysql/reducing-my-oss-involvement-and-how-it-affects-orchestrator-gh-ost

PR #1389, opened 2021-07-21
Expanded ReplicationAnalysis struct to support GR members info

qihengshan:
### Description

Obtain the Replication Group member information from the entry GetReplicationAnalysis, so that it is convenient to use this information to implement some logical judgments.

like this:
```go
// Group member is not reachable, has replicas, and none of its reachable replicas can replicate from it
if !a.LastCheckValid && a.CountReplicas > 0 && a.CountValidReplicatingReplicas == 0 {
	a.Analysis = DeadReplicationGroupMemberWithReplicas
	a.Description = "Group member is unreachable and all its reachable replicas are not replicating"
}
```

### DIFF
--- a/go/inst/analysis.go
+++ b/go/inst/analysis.go
@@ -136,6 +136,9 @@ type ReplicationAnalysis struct {
 	ReplicationDepth                          uint
 	Replicas                                  InstanceKeyMap
 	SlaveHosts                                InstanceKeyMap // for backwards compatibility. Equals `Replicas`
+	CountReplicationGroupMembers              uint           // GR
+	CountValidReplicationGroupMembers         uint
+	ReplicationGroupMemberHosts               InstanceKeyMap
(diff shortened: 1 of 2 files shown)


shlomi-noach:
I'm very sorry indeed, and thank you for the contribution: http://code.openark.org/blog/mysql/reducing-my-oss-involvement-and-how-it-affects-orchestrator-gh-ost

qihengshan:
> I'm very sorry indeed, and thank you for the contribution: http://code.openark.org/blog/mysql/reducing-my-oss-involvement-and-how-it-affects-orchestrator-gh-ost

Thanks, blessings happy every day.

jiangnanora:
@qihengshan  This PR already supports high-availability switching through orch in a similar GR mode. Trigger the effect of the corresponding hook?

PR #1392, opened 2021-07-27
Minor, but relevant docs updates

binwiederhier:
I wouldn't bother updating the docs for the other stuff, but the conclusion of "Example 1" is incorrect: it should be _"... and disable semi-sync on replica1 ..."_ (not: replica2). That may confuse a lot of people if we don't correct this


docs/configuration-discovery-classifying.md
```diff
@@ -113,17 +113,17 @@ Assuming this topology:
 ```
 
 `orchestrator` would detect a [`MasterWithTooManySemiSyncReplicas`](failure-detection.md#masterwithtoomanysemisyncreplicas) scenario
-and disable semi-sync on replica2.
+and disable semi-sync on replica1 (lower priority).
```
binwiederhier, line 116: This is the only major thing. The rest are cosmetic changes.

shlomi-noach (approved):

Issue #1412, opened 2021-10-13
orchestrator-client fails to auth with basic auth and headers


tomkrouper:
I ran into an issue while testing orchestrator-client where the headers_auth variable's existence caused previous request_auth settings to be ignored. This is an issue to match with the changes I'll be propose in a PR to be filed right after this.

PR #1413, opened 2021-10-13
Update orchestrator-client to append headers_auth

tomkrouper:
Related issue: https://github.com/openark/orchestrator/issues/1412

### Description

Currently if `$headers_auth` is set, it will replace any previous settings in `$requires_auth`. This change updates the script to append `$headers_auth` to `$requires_auth` instead of replace.


shlomi-noach (approved):
Thank you!

PR #1424, opened 2021-12-20
MySQL Orchestrator does not support new semi-sync variables introduced in MySQL 8.0.26

kamil-holubicki:
Problem:
MySql 8.0.26 introduced new semisync replication plugins
semisync_source.so and semisync_replica.so
in place of
semisync_master.so and semisync_slave.so
All semisync replication plugins' variables were renamed in the way
that 'master' was replaced with the term 'source' and 'slave' was
replaced by 'replica'.

As 8.0.26 contains also old plugins and when they are installed
everything works fine it is not enough to detect server version.
Moreover, it is possible to mix old source plugin with new replica
plugin and the opposite.

Solution:
Detect which plugins are installed and use proper variable names.

Related issue: https://github.com/openark/orchestrator/issues/0123456789

### Description

### DIFF
--- a/go/http/api.go
+++ b/go/http/api.go
@@ -75,6 +75,8 @@ var apiSynonyms = map[string]string{
 	"stop-slave-nice":            "stop-replica-nice",
 	"reset-slave":                "reset-replica",
 	"restart-slave-statements":   "restart-replica-statements",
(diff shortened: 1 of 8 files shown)


jiangnanora:
up

Issue #1467, opened 2023-01-17
Acknowledge from the orchestrator doesn't work


jwangace:
Ack from the web page always error out as: `returns "strconv.ParseInt: parsing \"undefined\"\: invalid syntax"`
This is due to a typo in the js file, corrected from audit.**Id** to audit.**ID**, which worked tested on my local web browser.

PR #1468, opened 2023-01-17
fix_Acknowledged_audit.ID_typo in audit-recovery.js file

jwangace:
Related issue: https://github.com/openark/orchestrator/issues/1467

### Description

Ack from the web page always error out as: returns "strconv.ParseInt: parsing \"undefined\"\: invalid syntax"
This is due to a typo in the js file, corrected from audit.Id to audit.ID, which worked tested on my local web browser.


jwangace:
**Before Fix:**
Screen Shot 2023-01-17 at 9 22 32 AM

**After Fix:**
Screen Shot 2023-01-17 at 9 23 54 AM

